PR #18, opened 2021-03-04
M17 test

sp5wwp:
Temporary workaround for the +1kHz TX frequency offset.


silseva:
Hi,
thanks for the modifications!
I only had to remove the offset compensation because, as I told you in the past days on discord, each radio seems to have its own offset. Mine, for example, transmits around 2kHz lower than the expected frequency (without your modification).
I'll fix this bug soon!

PR #22, opened 2021-04-02
Add support for encoder knob on MD-UV380

CBJamo:
## changes
Added: qdec, a header only quadrature encoder library

Added: EXTI15_10_IRQHandler to handle the encoder knob

Changed: platform_init() for the MD-UV380 now configures the EXTI15_10
interrupt.

Changed: platform_getChSelector now returns signed int8

Changed: size of settings_t.valid went from 6 to 7
Without this change, pressing the macro button crashes the radio, I
haven't investigated and I don't remember how I found that solution.

Changed: meson.build includes the qdec library

## Notes
Long term, I don't think the knob should emulate key presses, but rather send a knob-specific event. I don't know how hard that would be to handle in the UI code, but it would be much cleaner on this end. We're currently emulating an absolute encoder to match the MD-380, then turning that back into a left or right event. It does work, and it handles rollover, but is quite messy.

I'm looking into the settings_t.valid issue, but I suspect it doesn't have anything to do with my changes here.


silseva:
Merged, thanks!

PR #31, opened 2021-06-16
WIP: Artificial satellites, SDL simulator shell replacement

tarxvftech:
1st: This has a replacement for the SDL simulator shell. 
---------------
I like it for my purposes, but I didn't get back to it.
The main value for me was that it allowed for scripting interactions. I would run an inotify script so the code would recompile and this command line would run with every change, which made graphical stuff very nice to develop. This also allows for screenshots, which is a killer feature for me.
```
cat menu_test.txt | build_linux/openrtx_linux
```
where menu_test.txt is:
```
sleep 4000
nop SAT menu
key enter DOWN down down down down enter
ready
screenshot sat.bmp
nop GPS menu
key esc DOWN down down down enter
ready
screenshot gps.bmp
```
(you can `convert in.bmp -resize 600% out.jpg` to convert and upscale the screenshot to something reasonable to post to discord)

2nd: Artificial satellite tracking added. It's got two big problems. 
----------
The first is that there's a time offset or some other error relative to other tracking systems. It could very well be just the difference between GPS time and true UTC, or it could be a real bear of a problem or a bug in my conversions. Bill Gray's Lunar library has a few more functions that might help in this regard by replacing my implementations of poor understanding with his tested code. 

The second is that there's no way to update TLEs, so they're hardcoded at compile until and unless we can store them in nvmem. This is the big blocker for further development at the moment. 

Finally, it also doesn't do full pass prediction yet, like I wanted, or a number of other features like setting the transponder frequencies and doing doppler correction. I'd like to come back to this at some point but without the first and second problems both solved, it seemed like there was little immediate value.

Finally, disclaimers: this isn't ready. It needs to be criticized. I'm not as good with C as I allegedly should be. I wasn't there, I didn't do it, nobody saw me do it, you can't prove a thing.


tarxvftech:
Oops. I need to cherry pick a few chunks from another branch that'll actually let this compile, include the submodules, and bring it back up to the newer gfx_print api

tarxvftech:
Hey, I've spent some time on this and discovered `git checkout -p`.
I'm going to put the merged result through its paces and make sure it all still works and fix the inevitable problems and update this.

tarxvftech:
Possible issues for me to come back to:
* [ ] Ptt doesn't seem to have any effect in sim? Or is there just no visual feedback?

tarxvftech (commented):
First draft review looks alright, found a mistaken hunk, noted a couple changes in the pull request comment.

tarxvftech:
screenshot of emulator

I was getting crashes when drawing off the screen, such as when zoomed in and panning around. It also seemed like things that disappeared off the left edge would appear on the right edge, like in this screenshot. 

Nizzo pointed me towards asan for the crashes, which showed the culprit for both problems was setPixel allowing out-of-bounds draws when the pixel coords were negative. That being basically impossible with your previous setup, the only bounds checking was positive - adding some negative checks solved it right away.

tarxvftech:
(closed because of a history rewrite, will rebase these changes on new master)

PR #46, opened 2021-11-03
Add integer/lookup-table M17 Modulator

carpikes:
This PR adds two classes that can replace `M17Modulator`:
`M17IntegerModulator` which uses only integer arithmetic
`M17LookupModulator` which does not use arithmetic but only lookup tables

There is a small unit test available (`tests/unit/test_m17mod.cpp`) which does: 
* Benchmarks and print results
* Checks that given the same input, they all provide the same output

At the moment I can't build the test using meson, so I use this command:
```
g++ -Ofast -march=native tests/unit/test_m17mod.cpp -Iopenrtx/include openrtx/src/protocols/M17/M17*Modulator.cpp -Iopenrtx/include/protocols -o my_test
```
and these are the results on my laptop:
```
$ ./my_test
M17Modulator standard: 0.216000 msec
M17Modulator integer: 0.094000 msec
M17Modulator lookup: 0.197000 msec
```

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -38,6 +38,8 @@ openrtx_src = ['openrtx/src/state.c',
                'openrtx/src/memory_profiling.cpp',
                'openrtx/src/protocols/M17/M17Callsign.cpp',
                'openrtx/src/protocols/M17/M17Modulator.cpp',
+               'openrtx/src/protocols/M17/M17IntegerModulator.cpp',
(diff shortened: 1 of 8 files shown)


n1zzo:
Please rebase on latest `master`.

silseva:
Manually merged in `M17-integer-modulator` branch

PR #47, opened 2021-11-03
Add clang-format 

carpikes:
This PR adds a .clang-format style using a style as similar as possible to the current one and adds a script
`./scripts/clang_format.sh` which if run as it is, it reindents the whole codebase, if run with `--check` argument, it checks and prints which files are not indented correctly


silseva:
Additions:
- `OperandAlignmentStyle : AlignAfterOperator`
- `AlignConsecutiveStyle : Consecutive`
- `ArrayInitializerAlignmentStyle : Right`
- `BinaryOperatorStyle : None`
- `BreakConstructorInitializersStyle : AfterColon`
- `BreakInheritanceListStyle : AfterComma`

silseva:
Is it possible to create a script which runs clang-format only on the latest modified/added files?
Just to avoid clang-formatting all the codebase at every commit,

nolith:
> Is it possible to create a script which runs clang-format only on the latest modified/added files?
> Just to avoid clang-formatting all the codebase at every commit,


You can replace the `FILE_LIST` variable with the following one:

```bash 
FILE_LIST=$(cat <(git ls-files -m) <(git diff --name-only --cached) | sort | uniq | egrep '\.(c|cpp|h)$' | egrep -v 'lib/|subprojects/|platform/mcu')
```

This will only give you modified and staged files.

Issue #50, opened 2021-11-12
Display brightness setting


sp5wwp:
Since settings are not volatile anymore, you should prevent users from turning off the display's backlight (PWM=0).

nolith:
On the MD380, if it happens you can always do M+7 to bump up the brightness. (but you have to remember)

What where you thinking about? Preventing backlight to go to 0, or always bumping it to the min level at boot time if below that?

silseva:
For now let's implement the second one, that is bumping the brightness to a minimum level if a value of zero was saved. In this way we leave the possibility to set it to a very low level during operation, useful for those with delicate eyes.

silseva:
Feature/fix implemented by @nolith, we can close the issue.
Thanks! :D

PR #53, opened 2021-11-19
Never boot with zero brightness

nolith:
Enforce a minimum brightness level at boot time

Related to https://github.com/OpenRTX/OpenRTX/issues/50

### DIFF
--- a/openrtx/src/state.c
+++ b/openrtx/src/state.c
@@ -88,6 +88,14 @@ void state_init()
 
 void state_terminate()
 {
+    /*
+     * Never store a brightness of 0 to avoid booting with a black screen
+     */
+    if(state.settings.brightness == 0)
+    {
+        state.settings.brightness = 25;
+    }
+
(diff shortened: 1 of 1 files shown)


nolith:
I was mostly interested in testing my macOS toolchain + radio_tools.

It seems to be working as I was able to load this on the radio 🎊

silseva:
Thanks for the fix!
I just had and idea: can you move the check here? https://github.com/OpenRTX/OpenRTX/blob/1767470c50e5952b27dd37311e558e91c5a63a5e/openrtx/src/state.c#L98
In this way we are always sure that a value greater than zero is saved in the nonvolatile memory, providing a favorable backward compatibility with the previous versions.

PS: happy to know that everything works on MacOS! :D

nolith:
Thanks @silseva for your feedback. I moved the check in the terminate state as you suggested. Do you mind a second look at my patch? 🙏

PR #54, opened 2021-11-20
Show TX frequency during transmission

nolith:
When the channel has a shift, we should show the TX frequency when PTT is hold.

### DIFF
--- a/openrtx/src/ui/ui_main.c
+++ b/openrtx/src/ui/ui_main.c
@@ -88,11 +88,14 @@ void _ui_drawZoneChannel()
 
 void _ui_drawFrequency()
 {
+  unsigned long frequency = platform_getPttStatus() ?
+       frequency = last_state.channel.tx_frequency : last_state.channel.rx_frequency;
+
     // Print big numbers frequency
     gfx_print(layout.line3_pos, layout.line3_font, TEXT_ALIGN_CENTER,
               color_white, "%03lu.%05lu",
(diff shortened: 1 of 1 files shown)


nolith:
👋 @Nimayer please let me what do you think about this PR 🙏

Nimayer:
> 👋 @Nimayer please let me what do you think about this PR 🙏

I will look at it during the weekend.
Sorry for the delay 😔

nolith:
> I will look at it during the weekend.
> 
> Sorry for the delay 😔

No need to be sorry about this. 

Take your time, I just wanted to check in to make sure it was on your radar. 

Thanks for the time you put in this project 🙏

Nimayer:
Hi @nolith, the Pull Request looks good.

I have tested it on my radio and I confirm it's working.

Unfortunately the frequency change is a bit delayed because the UI is refreshed based on received events, 
and the PTT keypress does not generate an event, so the information is updated at 1Hz from the clock events.

A solution to this is to make the PTT keypress/release generate UI events, but it is outside of the scope of this PR.

nolith:
> Unfortunately the frequency change is a bit delayed because the UI is refreshed based on received events,
> and the PTT keypress does not generate an event, so the information is updated at 1Hz from the clock events.

Yes, I thought about generating the key press event for ptt but This was an easier first iteration. 

Overall I think we should make the tx explicit in the UI, like many radios shows a little "tx" somewhere on screen. So I agree with you, a key press event will make the UI change faster.

PR #55, opened 2021-11-20
Screen backlight timer

nolith:
This is a WIP, right now it send the screen in standby after 30s idle.

TODO:
- [x] get out from standby when the squelch open
- [x] get out from standby when in TX (PPT is not a keyboard event)
- [x] add a menu to configure the desired timeout (also no timeout)
- [x] store the timeout value in the settings

### DIFF
--- a/openrtx/include/settings.h
+++ b/openrtx/include/settings.h
@@ -23,15 +23,37 @@
 
 #include <hwconfig.h>
 
+typedef enum
+{
+    TIMER_OFF =  0,
+    TIMER_5S  =  1,
+    TIMER_10S =  2,
+    TIMER_15S =  3,
+    TIMER_20S =  4,
+    TIMER_25S =  5,
+    TIMER_30S =  6,
+    TIMER_1M  =  7,
+    TIMER_2M  =  8,
+    TIMER_3M  =  9,
+    TIMER_4M  = 10,
+    TIMER_5M  = 11,
+    TIMER_15M = 12,
+    TIMER_30M = 13,
+    TIMER_45M = 14,
+    TIMER_1H  = 15
+}
+display_timer_t;
+
 typedef struct
 {
-    uint8_t brightness;     // Display brightness
-    uint8_t contrast;       // Display contrast
-    uint8_t sqlLevel;       // Squelch level
-    uint8_t voxLevel;       // Vox level
-    int8_t utc_timezone;    // Timezone
-    bool gps_enabled;       // GPS active
-    char callsign[10];      // Plaintext callsign, for future use
+    uint8_t brightness;                  // Display brightness
+    uint8_t contrast;                    // Display contrast
+    uint8_t sqlLevel;                    // Squelch level
+    uint8_t voxLevel;                    // Vox level
+    int8_t utc_timezone;                 // Timezone
+    bool gps_enabled;                    // GPS active
+    char callsign[10];                   // Plaintext callsign, for future use
+    display_timer_t brightness_timer;    // Standby timer
 }
 __attribute__((packed)) settings_t;
 
@@ -48,7 +70,8 @@ static const settings_t default_settings =
     0,                // Vox level
     0,                // UTC Timezone
     false,            // GPS enabled
-    ""                // Empty callsign
+    "",               // Empty callsign
+    TIMER_30S         // 30 seconds
 };
 
 #endif /* SETTINGS_H */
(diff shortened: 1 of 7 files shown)


silseva:
In the next days I'll add an `rtx_squelchStatus()` function returning true when the squelch is open.
Meanwhile I'm thinking where is the best place to put the backlight logic: probably we can put it temporarily in the keyboard task because of its 40Hz update rate.
As a general implementation I'd put a downcounter in the global state, decremented by one at each round by the state task, which ticks at 1Hz. PTT press and squelch opening turn on the backlight and set the counter to the timeout value then, when it reaches zero, the backlight is turned off.

nolith:
Something that I realized using this code on a real QSO is that, like in all the other radio, you don't want the key press that get you out of the standby to be actionable. It's only purpose is to light on the screen

silseva:
Another point in favor of putting the backlight logic inside the keyboard task: in this way you can "trap" the keypress event and do not transmit it to the UI thread if the backlight is off. That is, the first press on any button with backlight off only turns on the backlight.

nolith:
@silseva I'll keep working on my other items from the list, but I think that even in the current shape is a good improvement.

silseva:
I added the `rtx_rxSquelchOpen()` API function, allowing to query the current status of the RX audio squelch

openrtx/src/ui/ui.c
```diff
@@ -381,6 +385,7 @@ layout_t _ui_calculateLayout()
 
 void ui_init()
 {
+    last_event_tick = getTick();
```
nolith, line 388: This is needed otherwise the linux emulator will start in standby

openrtx/src/ui/ui.c
```diff
@@ -640,6 +645,27 @@ void _ui_changeContrast(int variation)
     display_setContrast(state.settings.contrast);
 }
 
+void _ui_enterStandby()
+{
+    if(standby)
+        return;
```
nolith, line 701: Is this allowed under Allman bracing or should I always use {} even with just a return?

nolith:
> I added the `rtx_rxSquelchOpen()` API function, allowing to query the current status of the RX audio squelch

Thank you 🙏 

I'll try to find some time to work on the settings, both UI and memory.

openrtx/src/ui/ui.c
silseva: Keep it as it is now, we're not super-strict about those things!

silseva:
> > I added the `rtx_rxSquelchOpen()` API function, allowing to query the current status of the RX audio squelch
> 
> Thank you pray
> 
> I'll try to find some time to work on the settings, both UI and memory.

Sorry for the long processing time! Even if it was a 5-minute thing, it slipped out from my mind for a while...

openrtx/include/settings.h
```diff
@@ -23,15 +23,37 @@
...
-    bool gps_enabled;       // GPS active
-    char callsign[10];      // Plaintext callsign, for future use
+    uint8_t brightness;                  // Display brightness
+    uint8_t contrast;                    // Display contrast
+    uint8_t sqlLevel;                    // Squelch level
+    uint8_t voxLevel;                    // Vox level
+    int8_t utc_timezone;                 // Timezone
+    bool gps_enabled;                    // GPS active
+    char callsign[10];                   // Plaintext callsign, for future use
+    display_timer_t brightness_timer;    // Standby timer
```
nolith: will `__attribute__((packed))` be enough for this to 4bits only?

i guess no, I think this needs the union trick (like in event.h)

openrtx/include/settings.h
```diff
@@ -23,15 +23,37 @@
...
+    TIMER_30S =  6,
+    TIMER_1M  =  7,
+    TIMER_2M  =  8,
+    TIMER_3M  =  9,
+    TIMER_4M  = 10,
+    TIMER_5M  = 11,
+    TIMER_15M = 12,
+    TIMER_30M = 13,
+    TIMER_45M = 14,
+    TIMER_1H  = 15
```
nolith, line 43: these are the values I have on my anytone, and they nicely fits into 4 bits.

nolith (commented):
I'll do a final review once I have the radio.

It works on the emulator.

I'll probably do a round of refactoring to harmonise names, but the overall feature is ready.

nolith:
@silseva I think this is ready for review now

PR #56, opened 2021-11-22
MD380: adjust squelch with left and right keys

nolith:
I think that adjusting the squelch with an absolute position knob is really painful.

This little patch enables keypad squelch tuning on every radios.

### DIFF
--- a/openrtx/src/ui/ui.c
+++ b/openrtx/src/ui/ui.c
@@ -707,9 +707,12 @@ void _ui_fsm_menuMacro(kbd_msg_t msg, bool *sync_rtx) {
         state.settings.sqlLevel = platform_getChSelector() - 1;
         *sync_rtx = true;
     }
+
+    if(msg.keys & KEY_LEFT || msg.keys & KEY_DOWN)
 #else // Use left and right buttons or relative position knob
     // NOTE: Use up and down for UV380 which has not yet a functional knob
     if(msg.keys & KEY_LEFT || msg.keys & KEY_DOWN || msg.keys & KNOB_LEFT)
+#endif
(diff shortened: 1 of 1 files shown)


openrtx/src/ui/ui.c
```diff
@@ -707,7 +707,7 @@ void _ui_fsm_menuMacro(kbd_msg_t msg, bool *sync_rtx) {
         state.settings.sqlLevel = platform_getChSelector() - 1;
         *sync_rtx = true;
     }
-#else // Use left and right buttons or relative position knob
+#endif// Use left and right buttons or relative position knob
     // NOTE: Use up and down for UV380 which has not yet a functional knob
     if(msg.keys & KEY_LEFT || msg.keys & KEY_DOWN || msg.keys & KNOB_LEFT)
```
nolith, line 714: 🤦 this is changing the value twice in case of an absolute knob

nolith:
I'm not used to `#ifdef` (not available on my daily coding languages) so they always seems weird to me.

Please let me know if this makes sense or not 😄

silseva:
Is perfectly fine!
Probably we can even remove lines 706 - 709, but I'll leave the last word to @Nimayer.

Thanks for the PR, in the meantime! :D

Nimayer:
Thank you @nolith for the Pull Request, contributions are always welcome!

> Probably we can even remove lines 706 - 709, but I'll leave the last word to @Nimayer.

I think it is better to leave lines 706-709 there, this way every radio could use knob and arrows to change the squest.
This is the most consistent solution across radios.

The Pull Request is good, I'm merging it.

PR #59, opened 2021-11-28
macOS support

nolith:
This will allow to compile the linux target on macOS.

Why is this needed?

From https://wiki.libsdl.org/SDL_PumpEvents

> WARNING: This should only be run in the thread that initialized the video subsystem, and for extra safety, you should consider only doing those things on the main thread in any case.

With this PR we add an SDL main loop that replaces the `ui_task` on the main thread, moving the `ui_task` into a new thread.

We also introduce an unbuffered channel to lock both writer and reader to make sure `display_renderRows` terminates only when the rows are rendered by the SDL main loop.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -215,7 +215,7 @@ mk22fn512_def = {'_POSIX_PRIORITY_SCHEDULING':''}
 ## Linux
 ##
 linux_platform_src = ['platform/targets/linux/emulator/emulator.c',
-		      'platform/targets/linux/emulator/sdl_engine.c',
+                      'platform/targets/linux/emulator/sdl_engine.c',
                       'platform/drivers/display/display_libSDL.c',
                       'platform/drivers/keyboard/keyboard_linux.c',
                       'platform/drivers/NVM/nvmem_linux.c',
--- a/meson.build
+++ b/meson.build
@@ -30,6 +30,7 @@ openrtx_src = ['openrtx/src/state.c',
                'openrtx/src/input.c',
                'openrtx/src/calibUtils.c',
                'openrtx/src/queue.c',
+               'openrtx/src/chan.c',
                'openrtx/src/rtx/rtx.cpp',
                'openrtx/src/rtx/OpMode_FM.cpp',
                'openrtx/src/rtx/OpMode_M17.cpp',
@@ -214,6 +215,7 @@ mk22fn512_def = {'_POSIX_PRIORITY_SCHEDULING':''}
 ## Linux
 ##
 linux_platform_src = ['platform/targets/linux/emulator/emulator.c',
+		      'platform/targets/linux/emulator/sdl_engine.c',
                       'platform/drivers/display/display_libSDL.c',
                       'platform/drivers/keyboard/keyboard_linux.c',
                       'platform/drivers/NVM/nvmem_linux.c',
--- a/openrtx/src/main.c
+++ b/openrtx/src/main.c
@@ -27,6 +27,9 @@
 #include <interfaces/graphics.h>
 #include <interfaces/delays.h>
 #include <hwconfig.h>
+#ifdef PLATFORM_LINUX
+#include <emulator/sdl_engine.h>
+#endif
 
 extern void *ui_task(void *arg);
 
@@ -74,6 +77,16 @@ int main(void)
     // Create OpenRTX threads
     create_threads();
 
+#ifndef PLATFORM_LINUX
     // Jump to the UI task
     ui_task(NULL);
+#else
+    // macOS requires SDL main loop to run on the main thread.
+    // Here we create a new thread for ui_task and utilize the masin thread for
+    // the SDL main loop.
+    pthread_t      ui_thread;
+    pthread_create(&ui_thread, NULL, ui_task, NULL);
+
+    sdl_task();
+#endif
 }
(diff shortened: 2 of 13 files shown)


silseva:
Super, thank you!
Can you tell briefly what was wrong with SDL events?

nolith:
From https://wiki.libsdl.org/SDL_PumpEvents

WARNING: This should only be run in the thread that initialized the video subsystem, and for extra safety, you should consider only doing those things on the main thread in any case.

nolith:
The thing is that Linux seems to be fine with that, but macOS crashes immediately. 

Now this PR, while showing how to have it working on macOS, is not in mergiable state because the UI thread runs one e each second and it gives a poor performance in handling keystrokes. 

I think a better way will be to spawn a new thread for the event loop inside the emulator stuff and use it as sync point for sdl management. Both UI rendering and input parsing. 

This means that the display driver should ve re written to enqueue rendering operations inside this new thread

silseva:
Good to know, thanks!
Indeed, on Linux we never had such an issue and, consequently, this problem went ahead silently. I think that there is no problem in spawning a dedicated thread which handles both SDL render and SDL keyboard events: one can allocate an intermediate framebuffer and use a condition variable to signal that a new render operation is needed. On the other hand, the thread can push keypresses to a circular buffer being read by the `keyboard_getKeys` function.
At least it can work as a bridge solution until we perform an extensive refactoring of the codebase (already planned) which will merge into a single thread the UI, state and keyboard management.

meson.build
```diff
@@ -30,6 +30,7 @@ openrtx_src = ['openrtx/src/state.c',
                'openrtx/src/input.c',
                'openrtx/src/calibUtils.c',
                'openrtx/src/queue.c',
+               'openrtx/src/chan.c',
```
nolith, line 33: This is an unbuffered channel. It provides a synchronous way to handle data between threads.

I've put it in here as it's a generic structure, but we can move it under the emulator folder if we want to conditionally include only for the linux platform.

openrtx/src/main.c
```diff
@@ -74,6 +77,17 @@ int main(void)
...
     // Jump to the UI task
     ui_task(NULL);
+#else
+    // macOS requires SDL main loop to run on the main thread.
+    // Here we create a new thread for ui_task and utilize the masin thread for
+    // the SDL main loop.
+    pthread_t      ui_thread;
+    pthread_create(&ui_thread, NULL, ui_task, NULL);
+
+    sdl_task();
```
nolith, line 90: When I moved all the DSL logic into a custom thread I ended up with crashes at boot time. On macOS you can run the SDL main loop only on the Main Thread :(

So I had to rewire a bit of logic here in main, if on linux the `ui_task` runs on a thread and the `sdl_task` runs on the main thread.

osx.txt
```diff
@@ -0,0 +1,18 @@
+# deps: brew install gcc@9 codec2 binutils
+[binaries]
+c       = 'gcc-9'
```
nolith, line 3: As mentioned on discord, openrtx does not compile with macOS clang.

We coul try to make it compile, but in the end you still need to install gcc9 to compile miosix toolchain, so 🤷

platform/drivers/display/display_libSDL.c
```diff
@@ -25,139 +25,20 @@
...
-#endif
-
-SDL_Renderer *renderer;      /* SDL renderer           */
-SDL_Window *window;          /* SDL window             */
-SDL_Texture *displayTexture; /* SDL rendering surface  */
 void *frameBuffer;           /* Pointer to framebuffer */
 bool inProgress;             /* Flag to signal when rendering is in progress */
 
-
-int screenshot_display(const char *filename)
```
nolith, line 61: moved as is into `sdl_enigine.c`

platform/drivers/display/display_libSDL.c
```diff
@@ -259,36 +116,39 @@ void display_terminate()
...
     (void) startRow;
     (void) endRow;
-    PIXEL_SIZE *pixels;
-    int pitch = 0;
-    if (SDL_LockTexture(displayTexture, NULL, (void **) &pixels, &pitch) < 0)
+    inProgress = true;
+    if(!sdl_ready)
     {
-        printf("SDL_lock failed: %s\n", SDL_GetError());
+        sdl_ready = sdl_main_loop_ready();
```
nolith, line 131: This whole `sdl_ready` local variable may be an over optimisation. It's just a way to avoid an extra mutex check on each loop.

nolith (commented):
@silseva I think this is ready for review now.

silseva:
I just have to ask you three little things:
- replace tabs with four spaces
- to provide a bit of documentation for the `chan` functions following the doxygen style
- to uniform the code style to the one of the rest of the codebase, with Allman bracing.

Some guidelines can be found here: https://github.com/OpenRTX/openrtx.github.io/blob/master/coding_style.md
Don't get too much crazy at uniforming all the code style, update only your parts.

Thanks! :)

nolith:
> replace tabs with four spaces

🤦 I guess too many years of automatic linters and automatic reformatting left a marks on me.
Running emacs with whitespace mode to highlight tabs and spaces shows a very inconsistent auto-indentation. I have to fix my editor configuration for C/C++ code.

In any case, I should have removed all the tabs (in my files) now.

PR #63, opened 2022-01-10
Add NVM for settings and VFO on linux platform

nolith:
This is a very simple NVM implementation for settings and VFO under the linux platform.

This is preparatory work for implementing a factory reset procedure (it will be in another PR).

### DIFF
--- a/platform/drivers/NVM/nvmem_linux.c
+++ b/platform/drivers/NVM/nvmem_linux.c
@@ -19,21 +19,208 @@
  ***************************************************************************/
 
 #include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
 #include <interfaces/nvmem.h>
+#include <sys/stat.h>
+#include <sys/errno.h>
+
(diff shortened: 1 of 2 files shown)


silseva:
I just want to test it under linux to see if `strlcpy`, which seems not being part of the ISO C library, may cause compilation problems.

silseva:
Is it possible to substitute the `strlcpy` with the `strncpy`?
The `strlcpy` generates linking errors on linux: even if those can be easily fixed by adding a linker directive in the meson.build, it would be preferable to use functions from the standard C library.

nolith:
Thanks @silseva, I didn't realize that I was looking at a BSD man page instead of the standard C documentation.

I decided to opt for `strcpy`/`strcat` and get rid of the pointer arithmetics. The resulting code is easier to read and a desktop machine should not be afraid of the performance hit due to scanning a couple of strings to find the NULL terminators.

PR #68, opened 2022-03-15
inputStream for Linux

carpikes:
Hi, with this PR I added support for inputStream emulation on linux. It reads the inputStream from `AudioSource.raw`, e.g. mic source -> `MIC.raw`and emulates both with functionality and timings what the radio hardware should do.

Let me know what do you think


silseva:
I finally did a proper code review and I have only one request: instead of throwing an exception when the file is missing or has an invalid size, make `inputStream_start` return -1 and `inputStream_getData` return an empty data block.

PR #77, opened 2022-06-02
Module17-oriented PR

sp5wwp:
1. Changing the TX pot value in Module17 isn't needed, as the default value can be left as is. The intention was to reduce the baseband output amplitude, so the Yaesu FTM-6000 is happy with the signal. Of course this isn't the only radio that is M17 capable, so the modification is pointless. This is just a workaround anyway, as there's a menu entry needed for setting both TX and RX baseband levels.
2. The udev rule allows the use of `dfu-util` to program GD32-based Module17 without having root privileges.

### DIFF
--- a/99-openrtx.rules
+++ b/99-openrtx.rules
@@ -5,3 +5,5 @@
 SUBSYSTEM=="usb", ATTRS{idVendor}=="15a2", ATTRS{idProduct}=="0073", MODE="0666"
 # MDx family (example: Tytera MD380)
 SUBSYSTEM=="usb", ATTRS{idVendor}=="0483", ATTRS{idProduct}=="df11", MODE="0666"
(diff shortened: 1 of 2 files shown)


silseva:
Can you please provide more informative commit messages explicitly referring to Module 17 (first commit) and GD32 MCU (second commit)?

sp5wwp:
Sure.
1. This change isn't needed, as the default value can be left as is. The intention was to reduce the baseband output amplitude, so the Yaesu FTM-6000 is happy with the signal. Of course this isn't the only radio that is M17 capable, so the modification is pointless. This is just a workaround anyway, as there's a menu entry needed for setting both TX and RX baseband levels.
2. 2. The udev rule allows the use of `dfu-util` to program GD32-based Module17 without having root privileges.

PR #92, opened 2022-07-13
[do not merge] MD380 Two-point modulation

USA-RedDragon:
**Please do not merge this, as this is a draft PR for keeping track of information related to modulation issues in OpenRTX on the MD380. It adds bad code and extra verbose comments**

the change from `calData->freqAdjustMid` to `calData->freqAdjustHigh` is a red herring based on a mental off-by-one error. However, it did seem to actively change the output waveform and make it much closer to on center. 

Useful and related links:
- C6000 English translated datasheet - https://github.com/open-ham/OpenGD77/blob/main/docs/data_sheets/C6000/HR_C6000_updated_chinese_translated_english.pdf
- WIP Mapping of C5000 registers - https://gist.github.com/USA-RedDragon/cab6de88bc13dbb525f59cb47a62ff31
- D13.020 symbols - https://github.com/USA-RedDragon/md380tools/blob/master/applet/src/symbols_d13.020
- Known C5000 registers: https://www.qsl.net/dl4yhf/RT3/HR_C5000_english.html#registers
- Known C6000 registers: https://github.com/open-ham/OpenGD77/blob/main/docs/data_sheets/C6000/HR-C6000%20Registers_G4EML.docx

Here's the RE work that I've got so far: 

1_spi_flash_security_registers

I started by checking the references of `md380_spiflash_security_registers_read` to find where the calibration data is loaded into memory on the stock MD380 firmware. 

2_spi_flash_read

256 bytes are read from the flash starting at the 0x1000 address, ending up at 0x2001c380-0x2001c480 similar to https://github.com/OpenRTX/OpenRTX/blob/master/platform/drivers/NVM/nvmem_MD3x0.c#L46

3_c5000_register_set

Here, in the function that most seems like the C5000 initialization function (at address 0803f8f2), the C5000 0x47 register is set to 128 minus the value in 0x2001c389 that seems tied to the two-point modulation as the "Two Point Modulation Offset Low 8 Bits". Note that the source is 9 bytes after the MD380 calibration data. 

If we take this struct as a source of truth (I need to make sure on this), https://github.com/USA-RedDragon/OpenRTX/blob/two-point-modulation/openrtx/include/calibration/calibInfo_MDx.h#L35-L75, `freqAdjustMid` is the correct value to use. 

So at this point, we've got a bit more knowledge but don't seem to be closer to stock yet.

### DIFF
--- a/openrtx/include/calibration/calibInfo_MDx.h
+++ b/openrtx/include/calibration/calibInfo_MDx.h
@@ -34,17 +34,17 @@
  */
 typedef struct
 {
-    uint8_t vox1;
-    uint8_t vox10;
-    uint8_t rxLowVoltage;
-    uint8_t rxFullVoltage;
-    uint8_t rssi1;
-    uint8_t rssi4;
-    uint8_t analogMic;
-    uint8_t digitalMic;
-    uint8_t freqAdjustHigh;
-    uint8_t freqAdjustMid;
-    uint8_t freqAdjustLow;
+    uint8_t vox1;  // 0x2001c380
+    uint8_t vox10; // 0x2001c381
+    uint8_t rxLowVoltage; // 82
+    uint8_t rxFullVoltage; // 83
+    uint8_t rssi1; // 84
+    uint8_t rssi4; // 85
+    uint8_t analogMic; // 86
+    uint8_t digitalMic; // 87
+    uint8_t freqAdjustHigh; // 88
+    uint8_t freqAdjustMid; // 89 <<-!
+    uint8_t freqAdjustLow; // 8A
     freq_t  rxFreq[9];
     freq_t  txFreq[9];
     uint8_t txHighPower[9];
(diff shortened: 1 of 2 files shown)


sp5wwp:
The frequency error went down from 650 to 50Hz? What about the other cal values that are stored in the NV mem? Are they being used properly by OpenRTX?

USA-RedDragon:
> The frequency error went down from 650 to 50Hz

Yeah, I'm reflashing both variants of this so I can show a waterfall. My original waterfall recordings had a different scale so not very useful.

> The frequency error went down from 650 to 50Hz? What about the other cal values that are stored in the NV mem? Are they being used properly by OpenRTX?

There is one stark difference I'm seeing, but I'm still trying to find out if it's relevant. See screenshot 2 above. It reads in 256 bytes starting from 0x1000 on the SPI Flash chip. To my understanding, this would make the offsets 0x0001 for byte data and 0x0004 for uint32_t. So the expected value for C5000's 0x47 might actually be from SPI Flash 0x1009 which doesn't seem to be obtained: https://github.com/OpenRTX/OpenRTX/blob/7e8a96000107c1b087ff46c49466e16f349fadef/platform/drivers/NVM/nvmem_MD3x0.c#L39

Edit as I'm typing this reply: `freqAdjustMid` never gets set anywhere in OpenRTX for the MD380 that I can tell. This might pan out to be interesting

PR #93, opened 2022-07-15
MD3x0: RF fixes for frequency "wiggle" and offset

USA-RedDragon:
Two separate fixes here. One is to explicitly read the calibration data to fix the two-point modulation. The other sets the mic input register of the C5000 in the function to set FM mode to ensure it gets set early. When this first gets set right before TX the "frequency wiggle" is present.

Supersedes #92

I would greatly appreciate if someone could test these out (especially if you've got a spectrum analyzer), as I've only verified with an RTL-SDR


silseva:
Hi!
Thanks for the fixes, I'll test and integrate them as soon as I can! :)

PR #102, opened 2022-08-08
[Very very very WIP] DMR OpMode

USA-RedDragon:
This is a WIP build towards adding DMR to OpenRTX. Right now this is likely to break other platforms than the MD3x0. Some organization and cleanup still needs to be done, and there's the codec issue. It would be cool to see Codec2 DMR mode (in a technical context, in real life this would just send bad data to AMBE vocoders). 

This currently does not quite work, however the hardest parts are probably done. I'm mostly having issues with the SPI2 bus right now, awaiting an oscilloscope in the mail tomorrow to check it out.


silseva (changes requested):
1) I'd rename the `startDigitalTx` and `stopDigitalTx` functions to, respectively,  `startDmrTx` and `stopDmrTx`. Since the firmware can handle more than a digital mode, explicitly mentioning DMR in the function names makes things a little bit clear about what the chip is doing there.

2) Can you put commit 48d70c9c470c903cc83ecbe45cf86508064915cf in a separate PR, please? Being it a change related to M17, we can merge it immediately.

platform/drivers/baseband/radio_MD3x0.cpp
```diff
@@ -231,6 +232,10 @@ bool radio_checkRxDigitalSquelch()
 
 void radio_enableRx()
 {
+    if(config->opMode == OPMODE_DMR) {
```
silseva: Please don't assign the DMR ID inside the radio driver. Instad, do it in the OpMode handler.
The rationale is that this driver is meant to handle only the low-level management of the radio part itself, like PLL and signal path, while all the protocol-specific stuff is handled by each OpMode handler.

openrtx/include/interfaces/radio.h
```diff
@@ -55,6 +55,14 @@ void radio_init(const rtxStatus_t *rtxState);
  */
 void radio_terminate();
 
+void radio_initVocoder();
```
silseva, line 58: I think is better to move these API functions to a dedicated module and/or inside the OpMode DMR handler.
As I said above, the "radio" interface is meant primarily to handle only the management of the baseband section.

platform/drivers/baseband/HR_C5000_MDx.cpp
```diff
@@ -251,11 +275,39 @@ uint8_t HR_Cx000< M >::uSpi_sendRecv(const uint8_t value)
...
         incoming = (incoming << 1) | gpio_readPin(DMR_MISO);
+
+        gpio_clearPin(DMR_CLK);
+
         delayUs(1);
     }
 
     return incoming;
 }
+
```
silseva, line 293: Sorry, I didn't understand why you added a receive-only function. Isn't the `sendRecv` good enough?

silseva:
You did a great work, thank you! :D
Probably is a good idea to arrange a meeting so that I can tell you better how the codebase is organized and plan how to lay out the sources. Also, put your name and callsign in the copyright headers!

PR #105, opened 2022-08-27
Allow newly-added settings to take their default values

USA-RedDragon:
This allows devs to ensure new values added to the NVM will come into new releases with a sane value. Without this patch set, a new setting in the NVM would default to the value of data that previously occupied this space, regardless of whether that data was a part of the settings structure.

The process for adding a setting doesn't change, just ensure that the settings changes are append-only. The NVM driver has been updated to save the current structure size into the settings, to ensure we only load the data we want from the NVM.

### DIFF
--- a/openrtx/include/core/settings.h
+++ b/openrtx/include/core/settings.h
@@ -59,7 +59,8 @@ typedef struct
     uint8_t vpLevel         : 3,  // Voice prompt level
             vpPhoneticSpell : 1,  // Phonetic spell enabled
             _reserved       : 4;
-
+    unsigned int structSize;
+    uint16_t   crc;
 }
 __attribute__((packed)) settings_t;
 
@@ -81,7 +82,9 @@ static const settings_t default_settings =
     0,                // not used
     0,                // Voice prompts off
     0,                // Phonetic spell off
-    0                 // not used
+    0,                // not used
+    0,                // struct size
+    0,                // crc
 };
 
 #endif /* SETTINGS_H */
--- a/openrtx/src/core/state.c
+++ b/openrtx/src/core/state.c
@@ -38,12 +38,12 @@ void state_init()
     pthread_mutex_init(&state_mutex, NULL);
 
     /*
-     * Try loading settings from nonvolatile memory and default to sane values
-     * in case of failure.
+     * Default to sane values, then try overwriting default
+     * settings from nonvolatile memory.
      */
+    state.settings = default_settings;
     if(nvm_readSettings(&state.settings) < 0)
     {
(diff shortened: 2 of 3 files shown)


silseva:
Unfortunately, given the current implementation of how settings and VFO state are saved inside the NVM memory, this mechanism won't work :(
Currently the NVM driver computes the CRC of the whole data block composed by the `settings_t` and `channel_t` data structures: this means that, if the layout of the `settings_t` changes then also its dimension changes, causing a bad CRC and, consequently, the load of all the defaults. This can be easily fixed by bringing the CRC field inside the `settings_t` data structure, though.

Also, probably the settings migration can be simply done using a `memcpy` and an array with the sizes of the previous versions. The idea is to do something like this:
```
static const uint8_t settingSize[] = {12, 13, 16, 18};

settings_t settings = default_settings;
settings_t temp;
nvm_readSettings(&temp);
uint8_t size = settingSize[temp.version];
memcpy(&settings, &temp, size);
```
Basically you initialise all the values in the settings to their default values and then you overwrite only the relevant pieces.

USA-RedDragon:
That makes sense. I wrote this prior to the CRC check on NVM. Will keep that in mind for a new revision

USA-RedDragon:
This is ready for re-review. A couple of notes/questions:

1. For https://github.com/OpenRTX/OpenRTX/pull/105/files#diff-d31345e7d1ad5d9e8fe58079f885bf10597763ee8df26473a0fda51daa3ec5b3R47, why don't we just set the default OPNRTX here: https://github.com/OpenRTX/OpenRTX/pull/105/files#diff-a49baffec8eddc844ce91d696311062b41b8ad43edd56366929ce5ac07704864R80 
2. The previous CRCs were based on both the settings and the channel, but that seemed a bit out of place, especially with the CRC now being a part of settings. Should I restore this behavior, leave it as it is, or implement a CRC for the channel data as well? https://github.com/OpenRTX/OpenRTX/pull/105/files#diff-8f9fcfa94391c8b9d3875019954492a22208f271c2de11fc40a62c86a746ba73L92

silseva:
Hi!
Thanks for the re-bump, I'm going to review the PR soon.
About the crash: if you are using the TYT firmware update tool for flashing, see #125

USA-RedDragon:
Thanks @silseva. I don't think it's the TYT tool crash, since I'm using https://dmr.tools to flash. It seems like something about the `memcpy` in `findActiveBlock`.

silseva:
Keep in mind that a freshly erased flash memory sector has all its cells containing 0xFF, that's can be the issue.

PR #106, opened 2022-08-27
Standby led

USA-RedDragon:
Builds on settings migration in #105 to add a standby LED driver, allowing the LED to blink at an interval while the screen is on standby

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -69,6 +69,7 @@ openrtx_inc = ['openrtx/include',
                'platform/drivers/ADC',
                'platform/drivers/NVM',
                'platform/drivers/GPS',
+               'platform/drivers/LED',
                'platform/drivers/USB',
                'platform/drivers/tones',
                'platform/drivers/baseband',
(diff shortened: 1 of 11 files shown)


USA-RedDragon:
Video example of the feature: https://discord.com/channels/802861646816083978/994636505102430258/1000116355388936202

USA-RedDragon:
MP4 direct upload of same asset: https://user-images.githubusercontent.com/13051767/187044952-041128f2-eed7-4195-9d53-edb2d717d164.mp4

silseva:
Hi!
The idea is good, however I cannot merge this PR as it is now for two reasons:
- LEDs are already managed via the `platform_ledOn` and `platform_ledOff` functions in "platform.h".
- at the moment the LEDs are managed by the RF stage, which means that blink is not guaranteed in every condition.

My suggestion is to fully implement all the mechanism inside the `ui_updateFSM` function, which is called every 40ms and to blink the display backlight instead of the LEDs.

PR #116, opened 2022-10-02
Fix README hardware mod link

turnrye:
# What

The README had a dead link. I replaced it with the current version of the referenced page (I think).

# Why

Because having dead links is distracting.

# How

Updated the link

# Anything else

I'd like someone more familiar with the project to confirm that I put the right value for the page. I think the one I used is logical, but I may be wrong.


silseva:
My fault, thanks for the fix!

PR #135, opened 2023-04-18
Add menu wrapping for channel, bank and contact menu

### LINKED ISSUE #64: feature: Wrap selection when scrolling up from top of list
Latest nightly, v0.3.2-82-g2bb05a2f. TYT MD-UV390.

Opening the `Zone` menu, I can't scroll up to get to items at the bottom, not with the up/down buttons or the clicker knob. Maybe this is just what I'm used to from the stock firmware but it's pretty handy when you have a lot of zones or channels and want one near the bottom.

marcoSchr:
This will fix #64 

We check the number of channels, banks and contacts once on startup and then use those numbers to allow wrapping the menus.

### DIFF
--- a/openrtx/include/core/cps.h
+++ b/openrtx/include/core/cps.h
@@ -275,4 +275,10 @@ __attribute__((packed)) cps_header_t; // 88B
  */
 channel_t cps_getDefaultChannel();
 
+/**
+ * This function initializes the number of channels, banks
+ * and contacts
+ */
+void cps_init();
+
 #endif // CPS_H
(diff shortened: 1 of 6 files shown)


silseva:
Thanks a lot!
I'm going to merge your PR in the next days. :)

silseva:
Before merging, I have to ask you to do a little change to the code: please move the `menu_init()` function in `openrtx/include/core/cps.h` and call it `cps_init()`, then call the `cps_init()` here: https://github.com/OpenRTX/OpenRTX/blob/5f7729821857e7279fecd196e23df0dac4e04b4f/openrtx/src/core/openrtx.c#L67

Thank you!

PR #139, opened 2023-06-02
Check if in M17 mode before entering edit mode

marcoSchr:
# Observed behavior

- When in FM mode and clicking the HASH key the keypad would no longer respond
- The only way to leave edit mode would then be to use the macro key to switch to m17 to exit edit mode and then switch back to FM

# Expected Behavior

- When in FM mode and clicking the HASH key nothing happens and the keypad still responds

# Fix

To prevent this behavior we check if we are in M17 mode before entering edit mode


silseva:
Thank you!

73,
Silvano IU2KWO.

PR #141, opened 2023-07-01
Add 1750 Hz tone

### LINKED ISSUE #101: 1750 Hz Tone
Is there a possibility to generate a 1750 Hz tone yet? Otherwise I will look if I could contribute that :-)

For the background: German FM repeaters use this tone to open a repeater, instead of CTCSS.

marcoSchr:
When pressing the Hash key and the PTT key in FM mode a 1750 Hz tone is transmitted

This is currently only implemented for the MD-UV380, once the code is reviewed and OK, I can add the relevant logic to other radios.

This will resolve #101

### DIFF
--- a/openrtx/include/core/state.h
+++ b/openrtx/include/core/state.h
@@ -51,6 +51,7 @@ typedef struct
     bool       bank_enabled;
     uint16_t   bank;
     uint8_t    rtxStatus;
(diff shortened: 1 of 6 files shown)


silseva:
Thanks and sorry for the delay!

PR #148, opened 2023-07-20
Enable 1750Hz tone for GDx

marcoSchr:
This should allow the GDx to send a 1750Hz tone.
I did not test this, as I do not own the device.


silseva:
Thanks!
I'll test it on a DM-1801 (the code is the same), but I'm quite sure it works

PR #150, opened 2023-08-02
assets: Remove (not used)

antoniovazquezblanco:
Seems that there are a couple of files not being used. Should we remove them? Thanks!


silseva:
Thanks!

PR #152, opened 2023-08-07
Change English string of DST from Broadcast to ALL and updated brightness settings

k5jae:
This aligns with M17 Specification and other tools.


k5jae:
Accidentally updated the PR with the new brightness updates. Sorry. I'll leave intact unless there is an issue.

silseva:
Thank you!

PR #154, opened 2023-08-08
Fix unit tests, disable broken unit tests, and add GitHub action to run unit tests

turnrye:
# What

Give unit tests a little love. Get them (most of them) back executing and integrate running them into CI/CD.

# Why

I fell down a rabbit hole trying to generate a draft `.rtxc` file, which depended on the unit tests running. Along the way fixing them I figured integrating some rudimentary CI would help avoid this recurring in the future.

# How

- Create new GitHub Action for running unit tests, specifically not running "Sine Test" or "Linux InputStream Test" because they appear to be flaky when run via CI (See https://github.com/turnrye/OpenRTX/actions/runs/5800301499/job/15722219281)
- Update meson build config to include the refactored UI default files in the `unit_test_src` var
- Update meson build to skip the definition of M17 Demodulator Test because it needs refactoring to match recent M17 module changes, and that is beyond my comfort level today; it breaks building currently, so it needs to be skipped at the build
- Update meson build to skip the definition of Voice Prompts Test because the tests timeout at 30s consistently both on my local workstation and when run via CI; the test needs fixing but this is beyond my comfort level today
- Update unit tests M17 Golay Unit Test, M17 Viterbi Unit Test, M17 Demodulator Test, and M17 RRC Test to conform with recent M17 module changes; these were trivial changes and so I went ahead and fixed these
- Update unit tests Codeplug Test to reflect that M17 mode is index 2 and skip execution of `test_contactIndexFix()` because it was not passing but I do not know what the proper fix is

# Next Steps

- [ ] Log follow-up tasks for fixing disabled M17 test suites
- [ ] Log follow-up task for fixing disabled codeplug test case
- [ ] Log follow-up tasks for fixing skipped CI flaky test suites

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -1,0 +1,45 @@
+name: "Run unit tests"
+on:
+  workflow_dispatch:
+  push:
+  pull_request:
+
+jobs:
+  unit-test:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
+      - name: install-deps
+        run: |
+          sudo apt update
+          sudo apt install git pkg-config build-essential meson libsdl2-dev libreadline-dev dfu-util cmake libusb-1.0-0 libusb-1.0-0-dev
+      - name: build install codec2
+        run: |
+          cd ${{github.workspace}}
+          meson subprojects download
+          cd subprojects/codec2
+          mkdir build_linux
+          cd build_linux
+          cmake ..
+          make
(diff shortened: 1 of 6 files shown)


turnrye:
This is related to https://github.com/OpenRTX/OpenRTX/issues/158, as getting unit tests working was required to evaluate the current implementation of OBCF.

silseva:
Thank you!
Excluding some tests from being run via CI is perfectly fine: they are called "unit tests" but the ones currently present were written more to be used as entrypoints to test some parts of the codebase on x86 while they were being developed, up to the point that some does not even return a pass/fail value.

PR #156, opened 2023-08-08
Macro latching

### LINKED ISSUE #142: About latching of the macro menu
According to the source code and my observations the latching of the macro menu should behave as follows:
- When pressing the macro button for a long time, the menu latches
- Settings in the macro menu can be changed, buttons that do not change the macro menu are ignored
- 3 seconds after the last button press, the menu can be unlatched by any button press, and the button press will be interpreted as if not in the macro menu

I think this is not very intuitive. As a user I would expect a few things:
- When the macro menu is latched, at any time I can close it again with the macro button
- When the macro button is latched, the button press to unlatch the macro menu should not be processed

I propose the following changes to make the macro menu more intuitive:
- Pressing the macro button and quickly selecting a setting should behave as before, changing the setting and closing the macro menu again
- Pressing and holding the macro button should lock the macro menu
- Any settings can now be changed, any button not related to the macro button should be ignored
- Pressing the macro button again should unlock and hide the macro menu again

Additionally I noticed, that when I turn the radio on and latch/unlatch the macro menu multiple times, after a few times it will no longer latch, but it will still beep.

marcoSchr:
This merge request will change the behavior of the macro latching

Now when the macro button is held it will latch the macro menu.
The letter L will be displayed in the upper left corner to show the latched status.
When the macro menu is latched, the macro button will unlatch the menu.

It would be good if some one who is vision impaired can check, if the menu is still usable with this change.

Fix #142

### DIFF
--- a/openrtx/include/core/state.h
+++ b/openrtx/include/core/state.h
@@ -43,6 +43,7 @@ typedef struct
 
     uint8_t    ui_screen;
     uint8_t    tuner_mode;
+    bool       macro_latched;
 
     uint16_t   channel_index;
     channel_t  channel;
--- a/openrtx/src/ui/default/ui.c
+++ b/openrtx/src/ui/default/ui.c
@@ -1262,17 +1262,17 @@ void ui_updateFSM(bool *sync_rtx)
 
         // If MONI is pressed, activate MACRO functions
         bool moniPressed = (msg.keys & KEY_MONI) ? true : false;
-        if(moniPressed || FunctionKeyIsLatched())
+        if(moniPressed || state.macro_latched)
         {
             macro_menu = true;
             // long press moni on its own latches function.
-            if (moniPressed && msg.long_press && !input_getPressedNumber(msg))
+            if (moniPressed && msg.long_press && !state.macro_latched)
             {
-                SetFunctionLatchTimer(); // 3000 ms.
+                state.macro_latched = true;
             }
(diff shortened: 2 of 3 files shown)


antoniovazquezblanco:
Nice!

I was looking at this and thought of a possible alternative that may be even simpler but I don't know if desirable. Would it make sense to add a settings option to enable/disable macro latching? That would remove any "hidden" behaviour that may be complex for a newcomer to pickup. One just selects how they want their menu to behave and that's it...

Just a late idea :)

turnrye:
I've been using this with my `mduv3x0` and have found this macro menu to be much more intuitive and usable. Consider this another vote of support for this change.

silseva:
Thanks for the implementation!
Just one thing: I think is better to keep the latch timeout define, so that the the minimum time the macro button has to be kept pressed before latching the menu can be made independent from the long button press threshold.
The long press threshold is set to 700ms, which may be a bit short if you just want to activate the menu, have a quick look and then change something. Probably something like the original three seconds is better, or anyways something more than 700ms.

marcoSchr:
I have not changed how long it takes to latch the menu. It still latches after 700ms as before.
The 3s timer was only there to check, that no settings were changed in the last 3 seconds.
After the 3s it was possible to unlatch the menu.

silseva:
Thanks!

PR #164, opened 2023-08-15
M17 info on screen

### LINKED ISSUE #162: Show details when receiving M17
When receiving M17 some details should be displayed.
This includes:
- Source call sign
- Destination call sign
- Metadata?

For implementation, we would need to do at least the following:
- store data when receiving M17
- display data to the user (maybe instead of the destination currently shown when using M17)

marcoSchr:
This will show information on the screen when receiving M17.

When receiving a stream without the extended callsign data, the source and destination will be shown instead of the configured destination address.

M17_info

When receiving a stream with extended callsign data, the destination and original source will be displayed instead of the configured destination address. The original source and optional the reflector source will then be displayed instead of the frequency.
M17_info_extended

This will fix #162


marcoSchr:
This does currently not display info out of the meta field of the LSF.
When a repeater/hotspot transmits voice, the original source is found inside the meta field.
We would need to add the functionality to parse the meta field.

openrtx/src/ui/default/ui_main.c
```diff
@@ -127,14 +127,24 @@ void _ui_drawModeInfo(ui_state_t* ui_state)
...
-            if(ui_state->edit_mode)
-                dst = ui_state->new_callsign;
-            else
-                dst = (!strnlen(cfg.destination_address, 10)) ?
-                    currentLanguage->broadcast : cfg.destination_address;
+            rtxStatus_t rtxStatus = rtx_getCurrentStatus();
+            if (rtxStatus.M17_rx && rtxStatus.lsfOk) {
             gfx_print(layout.line2_pos, layout.line2_font, TEXT_ALIGN_CENTER,
-                  color_white, "#%s", dst);
+                      color_white, "dst:%s", rtxStatus.lsf_dst);
```
turnrye: Does it make sense to treat dst and src as localizable strings? If so, please add them to `stringsTable_t` and `englishStrings`.

marcoSchr: Probably yes.
At the moment I try to add support to the M17 code base to get the extended call sign
With it we would need to display up to 3 "call signs" for the source in addition to the destination.
Because we do not have enough space on the screen, I think using icons will be better in order to not make the screen to crammed.

sp5wwp:
What about the `Module17` target?

marcoSchr:
> What about the `Module17` target?

I do not own a `Module17`. So I wanted to implement this first for a target, which I can test.
I can later add the UI for `Module17` but someone else would need to test it.

marcoSchr:
The implementation is now finished. I now need to cleanup the code to display the text

openrtx/src/ui/default/ui_main.c
marcoSchr: We now use icons instead  of text

PR #165, opened 2023-08-16
Add GPS status icon to the macro menu and general symbol support

### LINKED ISSUE #161: Add support for symbols
# Motivation

To more efficiently communicate concepts in the UI, OpenRTX needs to be able to display symbols in addition to normal text.

## Requirements

- Support a variety of symbols, and make it easy to add new symbols in the future

## Choices

- Support SVGs

## References

The GFX package being used has a [good tutorial](https://learn.adafruit.com/creating-custom-symbol-font-for-adafruit-gfx-library/overview) on how to make new fonts.

turnrye:
# What

Now within the macro menu, when you have GPS enabled, it displays an icon in the top right corner indicating that it is on. The icon shows a different variant when there is a fix as well. This is somewhat hidden right now behind macro, but the intent over time is to experiment with locating more "secondary" UI elements there. In implementing this, I also put in place tooling for generating fonts from svgs.

# Why

While there aren't a lot of symbols needed today, the general capability will be needed. Start with a simple use case like GPS status: off/on(location unknown)/on(location known) and macro latch.

# How

A python script iterates over a list of SVGs which get interpreted into glyph bitmaps using their alpha channel. These SVGs are then stored into bitmaps using the same Adafruit_GFX font structures as the existing fonts in the project and saved as auto-generated `.h` files. Currently three sizes are being generated based on the sizes referenced in the various device configurations.

A new `gfx_printSymbol()` method was added to handle switching between the normal text font and the symbol font. It uses a new type and enum to reference the symbols. The enum values are auto generated by the python script as well, so the order of the list of the char values to reference symbols are all arbitrary (aside from space being a space by convention).

In the `ui_menu.c` implementation, add logic to handle drawing a gps status indicator depending on if the setting is turned on and if a fix is known. Also handle displaying a macro latch icon on the left-hand side.

By @marcoSchr 's recommendation, I also added some automation to help keep things in sync since this introduced generated code. There is a github action now that runs if there are changes in `openrtx/include/fonts/symbols` to ensure that the results of the script are what are currently checked in.

Screenshot from 2023-08-22 18-26-32

# Next Steps

There will probably need to be follow-up work to adjust the font generation to make it smarter. There are some opportunities for things like removing unnecessary white space and reusing symbols to produce complex shapes via layered elements. Additionally, much of the positioning associated with this in the font/glyph definition is based on trial and error with narrow use cases tested. I expect both of these areas would be revisited over time as the use case for symbols is expanded.

I based the python on a lot of stackoverflow posts and some notes from past projects. It's spaghetti at this point, but we can invest more there if and when needed.

Closes #161.


openrtx/include/fonts/symbols/scripts/generate_symbols.py
```diff
@@ -0,0 +1,158 @@
...
+            + "\n"
+        )
+        return out
+
+
+###################################################################
+
+for size in sizes:
+    font = FontDefinition("Symbols", size)
+    for symbol in symbols:
```
marcoSchr: Do we want to hardcode all svgs or use something like `for file in glob.glob(r'../sources/*.svg'):`?
This would make it easier to add svgs, by not requiring to touch the code when adding new icons

marcoSchr (commented):
We might want to have a test, that checks, that the user ran generate_symbols.py, when adding new SVGs.

openrtx/src/core/graphics.c
```diff
@@ -569,6 +580,21 @@ point_t gfx_print(point_t start, fontSize_t size, textAlign_t alignment,
     return gfx_printBuffer(start, size, alignment, color, text);
 }
 
+point_t gfx_printSymbol(point_t start, fontSize_t size, textAlign_t alignment,
```
marcoSchr, line 579: The function definition needs to be added to graphics.h:
```C
point_t gfx_printSymbol(point_t start, fontSize_t size, textAlign_t alignment,
                  color_t color, const char *fmt, ... );
```

openrtx/include/fonts/symbols/scripts/generate_symbols.py
turnrye: The downside here is that the order is what determines the index. So unless we're strict about the file names, it risks a new file being added in a location that would shift the chars used to access existing symbols.

turnrye:
> We might want to have a test, that checks, that the user ran generate_symbols.py, when adding new SVGs.

I'll look at doing that with GitHub actions.

openrtx/include/fonts/symbols/scripts/generate_symbols.py
marcoSchr: Ok, so it will be a tradeoff between maintaining the list or maintaining a coherent naming scheme for the svgs.
I think the list will then be easier to maintain.

openrtx/src/core/graphics.c
turnrye: Doh!

openrtx/include/fonts/symbols/scripts/generate_symbols.py
silseva, line 1: Can you move this inside the "scripts" folder in the repository root, please? Just to keep all the scripts together for consistency.

openrtx/src/core/graphics.c
```diff
@@ -77,12 +82,17 @@ static const GFXfont fonts[] = { TomThumb,            // 5pt
...
+                                 UbuntuRegular24pt7b,  // 24pt
     #else
     #error Unsupported font family!
     #endif
+                                 Symbols5pt7b,      // 5pt
+                                 Symbols6pt7b,      // 6pt
+                                 Symbols8pt7b       // 8pt
                                };
 
+static const uint8_t symbol_offset = SYMBOLS_SIZE_6PT - FONT_SIZE_6PT;
```
silseva: Instead of creating the `symbol_offset` constant, I'd create a dedicated enum for the symbols. `symbolSize_t` or something like that.

openrtx/include/core/graphics.h
```diff
@@ -238,6 +241,18 @@ point_t gfx_printBuffer(point_t start, fontSize_t size, textAlign_t alignment,
...
+/**
+ * Prints symbols the screen at the specified coordinates.
+ * @param start: text line start point, in pixel coordinates.
+ * @param size: text font size, defined as enum.
+ * @param alignment: text alignment type, defined as enum.
+ * @param color: text color, in color_t format.
+ * @param fmt: printf style format string
+ * @return text width and height as point_t coordinates
+ */
+point_t gfx_printSymbol(point_t start, fontSize_t size, textAlign_t alignment,
```
silseva, line 254: From what I understand from the code, symbols are bound to text characters. What if we use an enum to identify the symbol, instead?

openrtx/include/fonts/symbols/scripts/generate_symbols.py
turnrye: Good news, by adding an enum for the symbol name and having that autogenerated too, now this is indeed auto-discovered rather than hand curated.

marcoSchr:
There is already a `requirements.txt` file in the root of the repository. They should probably be merged to one file.

turnrye:
> There is already a `requirements.txt` file in the root of the repository. They should probably be merged to one file.

Thanks, fixed!

openrtx/src/ui/ui_menu.c
```diff
@@ -0,0 +1,1026 @@
+/***************************************************************************
```
turnrye: Mistaken commit

PR #166, opened 2023-08-16
Add a lint job

turnrye:
# What

When I was contributing, I realized that [code styles](http://openrtx.org/#/coding_style) could be replaced with just enforcing clang-format. Then styles are no longer a thought to be concerned with. 

# Why

To make contributing easier, both for the contributor and the maintainer.

# How

This PR leverages where meson automatically creates the `clang-format` and `clang-format-check` target if the host has clang-format installed.

The timing of landing this PR is critical -- just before it lands, catch this branch up with wherever it's being merged, and then `mason setup build; ninja -C build clang-format` needs to run and have the resulted checked in. Otherwise builds will be blocked.

# Open questions

- [ ] I think we should evaluate changing the clang-format configuration. The configuration used was already present, but it didn't appear to be getting followed. 
- [ ] There's a wrapper script present in the repo too, I think we should remove that to avoid any confusion about entry (do it via `meson setup build; ninja -C build clang-format`)


silseva:
Hi!
Thanks for the PR! About the current clang-format configuration: I started working on a version more adherent to the code stile of the OpenRTX codebase but, due to lack of time, I never manage to complete it (also because fine-tuning everything is a long process). I have to search where I left the modified file and attach it here.

Btw, something I tried to configure is to avoid the linter aggressively reorder the whitespaces in the struct definitions. For example [this file](https://github.com/OpenRTX/OpenRTX/blob/master/openrtx/include/calibration/calibInfo_GDx.h) became [this](https://github.com/OpenRTX/OpenRTX/blob/75b2ae8845d87ede6ebc9a0a27244cc1925a09a2/openrtx/include/calibration/calibInfo_GDx.h) after a lint pass, something I consider a regression in the code style. If you know a wayt to tell clang-format to not bother with these alignment, please let me know.

turnrye:
> Hi! Thanks for the PR! About the current clang-format configuration: I started working on a version more adherent to the code stile of the OpenRTX codebase but, due to lack of time, I never manage to complete it (also because fine-tuning everything is a long process). I have to search where I left the modified file and attach it here.
> 
> Btw, something I tried to configure is to avoid the linter aggressively reorder the whitespaces in the struct definitions. For example [this file](https://github.com/OpenRTX/OpenRTX/blob/master/openrtx/include/calibration/calibInfo_GDx.h) became [this](https://github.com/OpenRTX/OpenRTX/blob/75b2ae8845d87ede6ebc9a0a27244cc1925a09a2/openrtx/include/calibration/calibInfo_GDx.h) after a lint pass, something I consider a regression in the code style. If you know a wayt to tell clang-format to not bother with these alignment, please let me know.

Thanks @silseva ! Looking forward to your guidance on where these changes are undesired and clang-format needs more configuration.

Regarding the whitespace concern you mention, I can fix this. I'll push up a change in a few minutes,

openrtx/include/core/datetime.h
```diff
@@ -35,15 +35,14 @@ extern "C" {
...
-}
-datetime_t;
+    int8_t hour;    // Hours (0-23)
+    int8_t minute;  // Minutes (0-59)
+    int8_t second;  // Seconds (0-59)
+    int8_t day;     // Day of the week (1-7)
+    int8_t date;    // Day of the month (1-31)
+    int8_t month;   // Month (1-12)
+    uint8_t year;   // Year (0-99)
+} datetime_t;
```
turnrye: Todo: make the struct name appear in next line

.github/workflows/lint.yml
```diff
@@ -0,0 +1,44 @@
...
+          cmake ..
+          make
+          sudo make install
+      - name: setup meson
+        run: |
+          cd ${{github.workspace}}
+          meson setup build
+      - name: Dump some debugs
+        run: |
+          which clang-format
```
turnrye: Why does clang-format exist here yet at the previous job it said it didn't?

openrtx/include/calibration/calibInfo_GDx.h
```diff
@@ -29,63 +29,63 @@
...
-    uint16_t _DigitalRxGainNarrowband; // 0x000 IF Gain, RX Fine
-    uint16_t _DigitalTxGainNarrowband; // 0x002 IF Gain, TX Fine
-    uint16_t _DigitalRxGainWideband;   // 0x004 IF Gain, RX Coarse
-    uint16_t _DigitalTxGainWideband;   // 0x006 IF Gain, TX Coarse
-    uint16_t modBias;                  // 0x008
-    uint8_t mod2Offset;                // 0x00A
-    uint8_t txLowPower[16];            // 0x00B - 0x02A
-    uint8_t txHighPower[16];           // 0x00B - 0x02A
-    uint8_t analogSqlThresh[8];        // 0x03F
+    uint16_t _DigitalRxGainNarrowband;  // 0x000 IF Gain, RX Fine
```
turnrye: Make minimum spacing before comments be 1 space, not 2

openrtx/include/core/gps.h
```diff
@@ -28,33 +28,31 @@
...
-    uint32_t   active_sats;          // Bitmap representing which sats are part of the fix
-    float      latitude;             // Latitude coordinates
-    float      longitude;            // Longitude coordinates
-    float      altitude;             // Antenna altitude above mean sea level (geoid) in m
-    float      speed;                // Ground speed in km/h
-    float      tmg_mag;              // Course over ground, degrees, magnetic
-    float      tmg_true;             // Course over ground, degrees, true
-}
-gps_t;
+    datetime_t timestamp;        // Timestamp of the latest GPS update
```
turnrye: Make variable names align

openrtx/include/fonts/adafruit/FreeSans10pt7b.h
```diff
@@ -1,216 +1,215 @@
 static const uint8_t FreeSans10pt7bBitmaps[] PROGMEM = {
```
turnrye: This file is supposed to be ignored

silseva:
Here you are the clang-format file I began to tweak. As you'll see, I set the `PenaltyBreakComment` to an insanely high value to prevent the lint to break a line when the comment overflows the 80 line limit by some characters. Rationale for this is:
1) comments are... comments, I don't consider it crucial to have them fall within the 80 column limit;
2) there are places where the lint pass makes the code worse than before because it breaks the line in an insane place, like [here](https://github.com/OpenRTX/OpenRTX/blob/75b2ae8845d87ede6ebc9a0a27244cc1925a09a2/openrtx/include/calibration/calibInfo_GDx.h#L84) (before it was [this](https://github.com/OpenRTX/OpenRTX/blob/e689e2f7ae53f15c2828bc5edeb8ed801ddd8050/openrtx/include/calibration/calibInfo_GDx.h#L84)).
 
Thanks for the work done so far!
[clang-format.txt](https://github.com/OpenRTX/OpenRTX/files/12361711/clang-format.txt)

openrtx/include/core/datetime.h
turnrye: Didn't find a way to do this.

turnrye:
@silseva when you get a moment, please re-review and point me to some examples of the patterns of issues that you see. 


I do notice right now where the following rule from [the website](http://openrtx.org/#/coding_style) is being contradicted:

> A space must be put between angle brackets and template argument in C++ code

Unfortunately it does not appear that this is compatible with clang-format. The formatter will always remove the spaces.

There are some changes this introduces to typedef names collapsing onto the end of the struct's definition block, but I'm also not seeing a way to change this with clang-format.

Other than these, I'm not seeing other coding styles that are obstructed by this change. But C/C++ is not my experience, and so I'm sure that I'm missing things.

openrtx/include/calibration/calibInfo_GDx.h
```diff
@@ -76,16 +76,16 @@ typedef struct
...
-}
-bandCalData_t;
+} bandCalData_t;
 
 typedef struct
 {
-    bandCalData_t data[2];      // Calibration data for VHF (index 0) and UHF (index 1) bands
-    freq_t vhfCalPoints[8];     // VHF calibration points for both TX power and mod1Amplitude
+    bandCalData_t
+        data[2]; // Calibration data for VHF (index 0) and UHF (index 1) bands
```
silseva: Is there a way to avoid having the lint doing this? Ok that the line goes above the 80 column limit, but breaking it like this make it unreadable. Probably the only way to fix this behavior is to split the comment (but geez).

openrtx/include/calibration/calibInfo_Mod17.h
```diff
@@ -32,11 +32,8 @@ typedef struct
...
     uint16_t tx_wiper;
     uint16_t rx_wiper;
-    uint8_t  mic_gain;
-    uint8_t  tx_invert : 1,
-             rx_invert : 1,
-             _padding  : 6;
-}
-mod17Calib_t;
+    uint8_t mic_gain;
+    uint8_t tx_invert : 1, rx_invert : 1, _padding : 6;
```
silseva: We should prevent the lint from doing this

openrtx/include/core/cps.h
```diff
@@ -45,27 +44,23 @@
...
-    uint8_t rxToneEn : 1,   //< RX CTC/DCS tone enable
-            rxTone   : 7;   //< RX CTC/DCS tone index
-    uint8_t txToneEn : 1,   //< TX CTC/DCS tone enable
-            txTone   : 7;   //< TX CTC/DCS tone index
-}
-__attribute__((packed)) fmInfo_t; // 2B
-
-
+    uint8_t rxToneEn : 1,           //< RX CTC/DCS tone enable
+        rxTone       : 7;           //< RX CTC/DCS tone index
```
silseva: All the fields should stay aligned, like it was before. This makes them more readable

openrtx/include/core/graphics.h
```diff
@@ -223,8 +221,11 @@ uint8_t gfx_getFontHeight(fontSize_t size);
  * @param buf: char buffer
  * @return text width and height as point_t coordinates
  */
-point_t gfx_printBuffer(point_t start, fontSize_t size, textAlign_t alignment,
-                        color_t color, const char *buf);
+point_t gfx_printBuffer(point_t start,
```
silseva: Probably is better to avoid having the pile of arguments, when they are more than three the linted version becomes a bit more messy than the original one

openrtx/include/core/ringbuf.hpp
```diff
@@ -26,17 +26,17 @@
...
 
 /**
  * Class implementing a statically allocated circular buffer with blocking and
  * non-blocking push and pop functions.
  */
-template < typename T, size_t N >
+template <typename T, size_t N>
 class RingBuffer
 {
-public:
```
silseva: Is it possible to leave the `private`, `public` and `protected` fields with indentation zero (as they where before)?

openrtx/include/core/gps.h
```diff
@@ -28,33 +28,31 @@
...
-    uint32_t   active_sats;          // Bitmap representing which sats are part of the fix
-    float      latitude;             // Latitude coordinates
-    float      longitude;            // Longitude coordinates
-    float      altitude;             // Antenna altitude above mean sea level (geoid) in m
-    float      speed;                // Ground speed in km/h
-    float      tmg_mag;              // Course over ground, degrees, magnetic
-    float      tmg_true;             // Course over ground, degrees, true
-}
-gps_t;
+    datetime_t timestamp;       // Timestamp of the latest GPS update
```
silseva: Is it possible to tell the lint to do the opposite of what is doing here? Instead of chewing up the extra spaces, I think is better, for readability, to have all the field names aligned.

openrtx/include/core/state.h
```diff
@@ -35,32 +35,31 @@
  */
 typedef struct
 {
```
silseva: Just another example: removing the extra spaces broke the alignment of the fields, making the code less clear

openrtx/include/core/voicePrompts.h
```diff
@@ -198,23 +198,21 @@ typedef enum
...
-}
-vpFlags_t;
+} vpFlags_t;
 
 /**
  * Queuing flags determining if speech is interrupted, played immediately,
  * whether prompts are queued for values, etc.
  */
 typedef enum
 {
```
silseva: Urgh! Also here: please, lint, leave the alignment as it was :)

openrtx/include/protocols/M17/M17CodePuncturing.hpp
```diff
@@ -36,22 +37,73 @@ namespace M17
 /**
  * Puncture matrix for linx setup frame.
  */
-static constexpr auto LSF_PUNCTURE = std::experimental::make_array< uint8_t >
```
silseva: Here the lint went crazy!

openrtx/include/protocols/M17/M17Constants.hpp
```diff
@@ -37,12 +37,15 @@ static constexpr size_t M17_FRAME_SYMBOLS    = 192;
...
 static constexpr size_t M17_FRAME_BYTES      = M17_FRAME_SYMBOLS / 4;
 
-static constexpr syncw_t LSF_SYNC_WORD    = {0x55, 0xF7};  // LSF sync word
-static constexpr syncw_t BERT_SYNC_WORD   = {0xDF, 0x55};  // BERT data sync word
-static constexpr syncw_t STREAM_SYNC_WORD = {0xFF, 0x5D};  // Stream data sync word
-static constexpr syncw_t PACKET_SYNC_WORD = {0x75, 0xFF};  // Packet data sync word
-static constexpr syncw_t EOT_SYNC_WORD    = {0x55, 0x5D};  // End of transmission sync word
+static constexpr syncw_t LSF_SYNC_WORD    = {0x55, 0xF7}; // LSF sync word
+static constexpr syncw_t BERT_SYNC_WORD   = {0xDF, 0x55}; // BERT data sync word
+static constexpr syncw_t STREAM_SYNC_WORD = {0xFF,
```
silseva: Why did the lint break this? Again for the 80 column overflow?

openrtx/include/protocols/M17/M17Demodulator.hpp
```diff
@@ -115,54 +114,61 @@ class M17Demodulator
...
-    static constexpr size_t  M17_SAMPLE_BUF_SIZE    = M17_FRAME_SAMPLES / 2;
-    static constexpr size_t  M17_SYNCWORD_SAMPLES   = M17_SAMPLES_PER_SYMBOL * M17_SYNCWORD_SYMBOLS;
-    static constexpr int8_t  SYNC_SWEEP_WIDTH       = 10;
-    static constexpr int8_t  SYNC_SWEEP_OFFSET      = ceil(SYNC_SWEEP_WIDTH / M17_SAMPLES_PER_SYMBOL);
-    static constexpr int16_t M17_BRIDGE_SIZE        = M17_SYNCWORD_SAMPLES + 2 * SYNC_SWEEP_WIDTH;
-
-    static constexpr float  CONV_STATS_ALPHA       = 0.005f;
-    static constexpr float  CONV_THRESHOLD_FACTOR  = 3.40;
-    static constexpr int16_t QNT_SMA_WINDOW        = 8;
+    static constexpr size_t M17_RX_SAMPLE_RATE = 24000;
```
silseva: Ok that we overflowed the 80 columns, but it was more readable... Probably we have to add lint exceptions here and there

openrtx/include/protocols/M17/M17Demodulator.hpp
```diff
@@ -115,54 +114,61 @@ class M17Demodulator
...
-    int8_t  stream_syncword[M17_SYNCWORD_SYMBOLS] = { -3, -3, -3, -3, +3, +3, -3, +3 };
+    int8_t lsf_syncword[M17_SYNCWORD_SYMBOLS]    = {+3, +3, +3, +3,
+                                                    -3, -3, +3, -3};
+    int8_t stream_syncword[M17_SYNCWORD_SYMBOLS] = {-3, -3, -3, -3,
+                                                    +3, +3, -3, +3};
 
     /*
      * Buffers
      */
-    std::unique_ptr< int16_t[] > baseband_buffer; ///< Buffer for baseband audio handling.
```
silseva: Also here we have a readability regression

silseva:
I did a round of revision and, while doing it, I remembered why at some point I gave up with adjusting the clang-format file: I found that in some occasions the lint behavior is way too rigid and leads to code worse than before in terms of readability.
Probably is better to have the clang-format ready and apply it by hand before opening a PR or merging a branch into master, instead of having it run as a build hook. In this way we have it doing all the heavy lifting and we just have to tweak the remaining spots of code style by hand. The alternative may be to place "no lint" directives around in the code, but it does not look very elegant to me...

Comments/opinions are welcome!

turnrye:
> I did a round of revision and, while doing it, I remembered why at some point I gave up with adjusting the clang-format file: I found that in some occasions the lint behavior is way too rigid and leads to code worse than before in terms of readability. Probably is better to have the clang-format ready and apply it by hand before opening a PR or merging a branch into master, instead of having it run as a build hook. In this way we have it doing all the heavy lifting and we just have to tweak the remaining spots of code style by hand. The alternative may be to place "no lint" directives around in the code, but it does not look very elegant to me...
> 
> Comments/opinions are welcome!

Thanks for the detailed feedback. I'll work on these items next.

As far as the rigidity of the formatter, I agree that at times that results in less readable code. The tradeoff, that contributors and reviewers no longer have to be concerned with formatting as they make changes, is quite empowering. In most of my past projects we've found that the compromise is worth it -- but that doesn't necessarily mean that it is here.

Let's see how close I can continue to make this PR before we really consider that question. If in the end it is still unclear what path to take, there are other options that we can take too (optional? just formatting the changed code?).

openrtx/include/core/cps.h
turnrye: Help me understand what I'm looking at here, is this called a "multiple adjacent bit field"?

silseva: Yes!

openrtx/include/calibration/calibInfo_Mod17.h
turnrye: Right now this is happening because there are no comments after the field definition. The other cases where the newlines are maintained are there because of the comments. I'm unclear on whether we can force these to new lines without comments (or disabling clang-format for the block).

Will investigate feasibility here more later.

openrtx/include/protocols/M17/M17CodePuncturing.hpp
turnrye: Haha, going to have to use a `//clang-format off` here.

openrtx/include/protocols/M17/M17Constants.hpp
turnrye: Yes, but we can force the comment to wrap instead.

openrtx/include/core/cps.h
turnrye: Haven't cracked how to do this yet.

turnrye:
Dusting this contribution off! @silseva shared with me that he wished for this project to migrate to linux kernel style, so I'm getting this branch up to date and incorporating that change into this PR.

.clang-format
```diff
@@ -1,35 +1,811 @@
...
+ConstructorInitializerAllOnOneLineOrOnePerLine: false
+ConstructorInitializerIndentWidth: 4
+ContinuationIndentWidth: 4
+Cpp11BracedListStyle: false
+DerivePointerAlignment: false
+DisableFormat: false
+ExperimentalAutoDetectBinPacking: false
+FixNamespaceComments: false
+
+# Taken from:
```
silseva: Actually we don't need these, they're very linux-specific

turnrye: Sounds good! I removed it and then noted the change in the header. Please take another look at your convenience, and resolve this conversation if I've addressed your feedback.

silseva:
Three observations:
1) I saw that, when bringing in the specification for the linux code style, some configs are gone: I have to go through them and see which ones should be kept;
2) I prefer not to have the commit passing the entire repository through the clang-format: even though it makes sense doing so, I think is a change too much intrusive. It will also generate an enormous pile of merge conficts with the branches currently under development and the PRs;
3) I'm afraid that automating the linting via github actions is again too much intrusive. The point is that clang-format is blindly applying the specified set of rules and, in doing so, it can make the the code look worse than it was before the linting pass.

turnrye:
> Three observations:
> 
>     1. I saw that, when bringing in the specification for the linux code style, some configs are gone: I have to go through them and see which ones should be kept;
The differences are noted in .clang-format L1-5, though I do encourage you to verify my work.
> 
>     2. I prefer not to have the commit passing the entire repository through the clang-format: even though it makes sense doing so, I think is a change too much intrusive. It will also generate an enormous pile of merge conficts with the branches currently under development and the PRs;
Ok, I can undo that part. Let's document a strategy then for how we'll work towards getting the whole project formatted and tackle that as a follow-up.
>     3. I'm afraid that automating the linting via github actions is again too much intrusive. The point is that clang-format is blindly applying the specified set of rules and, in doing so, it can make the the code look worse than it was before the linting pass.
The CI just verifies that all of the code is formatted. It is not making code changes. The goal is to use CI to indicate if the project is properly formatted or not. We can't do this in the same way due to the changes you've requested in observation 2, but perhaps we can find a way to incrementally validate as part of that. I've put my suggestions in https://github.com/OpenRTX/OpenRTX/issues/346.

.clang-format
```diff
@@ -1,35 +1,128 @@
...
-NamespaceIndentation: Inner
-
-BinPackParameters: 'false'
-BreakBeforeBraces: Custom
+AccessModifierOffset: -4
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlines: Left
+AlignOperands: true
```
silseva: Add
```
AlignConsecutiveBitFields: true
AlignOperands: AlignAfterOperator
AlignTrailingComments:
  Kind: Always
  OverEmptyLines: 1
```

.clang-format
```diff
@@ -1,35 +1,128 @@
...
+SpaceBeforeParens: ControlStatementsExceptForEachMacros
+SpaceBeforeRangeBasedForLoopColon: true
+SpaceInEmptyParentheses: false
+SpacesBeforeTrailingComments: 1
+SpacesInAngles: false
+SpacesInContainerLiterals: false
+SpacesInCStyleCastParentheses: false
+SpacesInParentheses: false
+SpacesInSquareBrackets: false
+Standard: Cpp03
```
silseva: Change to `Standard: c++14`

silseva:
I'd really like to find a way to prevent clang-format from playing shenanigans like this
```
typedef struct {
    bandCalData_t
        data[2]; // Calibration data for VHF (index 0) and UHF (index 1) bands
    freq_t vhfCalPoints
        [8];     // VHF calibration points for both TX power and mod1Amplitude
    freq_t uhfPwrCalPoints[16]; // UHF calibration points for TX power
    freq_t uhfCalPoints[8];     // UHF calibration points for mod1Amplitude
} gdxCalibration_t;
```

The workaround is to set `ColumnLimit` to something a bit more than 80, like 100. Otherwise, I don't know if there is a way to tell clang-format to not break a like if the comment goes out of the column limit.

silseva:
> Ok, I can undo that part. Let's document a strategy then for how we'll work towards getting the whole project formatted and tackle that as a follow-up.
> 
Thanks!

> The CI just verifies that all of the code is formatted. It is not making code changes. The goal is to use CI to indicate if the project is properly formatted or not. We can't do this in the same way due to the changes you've requested in observation 2, but perhaps we can find a way to incrementally validate as part of that. I've put my suggestions in #346.

Ah, ok. I thought it had a more invasive approach. If the format doesn't pass, the CI rises an error? If we can choose between an error and a warning, I'd probably choose the latter

turnrye:
> If the format doesn't pass, the CI rises an error? If we can choose between an error and a warning, I'd probably choose the latter

Github actions are pass/fail only. And in this case, I'd strongly encourage having a set of automated formatting rules that we strictly enforce rather than ones that are only partially enforced. Having the clang-format passing be a requirement helps alleviate any judgement from the developer or reviewer on the formatting. It may not be perfect all the time, but my idea is that the time spent manually formatting and reviewing code format is far more than the time lost to having to read imperfectly formatted code. If we make this check optional, then either we aren't planning to follow the format, or we are still burdening both parties with making judgements about formatting.

If we encounter a section where clang-format is just so wildly wrong that we cannot accept it, comments can be added to the code `//clang-format off` and `//clang-format on`. That makes that block manually formatted, and it's captured in the code itself that it is such. But making the job optional by way of being a log only and an always passing github action I think significantly reduces its value.

.clang-format
turnrye: Sounds good! I added/modified these and then noted the change in the header. Please take another look at your convenience, and resolve this conversation if I've addressed your feedback.

.clang-format
turnrye: Sounds good! I modified this and then noted the change in the header. Please take another look at your convenience, and resolve this conversation if I've addressed your feedback.

turnrye:
> I'd really like to find a way to prevent clang-format from playing shenanigans like this
> 
> ```
> typedef struct {
>     bandCalData_t
>         data[2]; // Calibration data for VHF (index 0) and UHF (index 1) bands
>     freq_t vhfCalPoints
>         [8];     // VHF calibration points for both TX power and mod1Amplitude
>     freq_t uhfPwrCalPoints[16]; // UHF calibration points for TX power
>     freq_t uhfCalPoints[8];     // UHF calibration points for mod1Amplitude
> } gdxCalibration_t;
> ```
> 
> The workaround is to set `ColumnLimit` to something a bit more than 80, like 100. Otherwise, I don't know if there is a way to tell clang-format to not break a like if the comment goes out of the column limit.

How frequent is this? May want to change our convention for comment placement or disable formatting for this section. Edit: my searching also yields that there isn't a solution, and it seems like the best alternative is to shift comments to their own line in this scenario.

silseva:
> How frequent is this? May want to change our convention for comment placement or disable formatting for this section. Edit: my searching also yields that there isn't a solution, and it seems like the best alternative is to shift comments to their own line in this scenario.

Is not very frequent, actually. We can just `clang-format off` these cases and call it a day

PR #170, opened 2023-08-18
Extend the M17 data structures

marcoSchr:
This will extend the data structures for M17 to include the data types for the encryption sub types and the data structures that can be represented inside the meta field for the null encryption.

This does not change any functionality.

Additional this changes the `decodeCallsign` function to use space instead of `x` and return `ALL` instead of `BROADCAST`

### DIFF
--- a/openrtx/include/protocols/M17/M17Datatypes.hpp
+++ b/openrtx/include/protocols/M17/M17Datatypes.hpp
@@ -126,7 +126,7 @@ typedef union
         m17_lsf_encryption_type encType : 2;  //< Encryption type
         uint8_t encryption_subtype      : 2;  //< Encryption subtype
         uint16_t CAN : 4;  //< Channel Access Number
-        uint16_t     : 4;  //< Reserved, padding to 16 bit
+        uint16_t     : 5;  //< Reserved, padding to 16 bit
     } fields;
 
     uint16_t value;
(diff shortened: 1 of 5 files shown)


openrtx/include/protocols/M17/M17Datatypes.hpp
```diff
@@ -46,19 +121,17 @@ typedef union
...
     {
-        uint16_t stream     : 1;    //< Packet/stream indicator: 0 = packet, 1 = stream
-        uint16_t dataType   : 2;    //< Data type indicator
-        uint16_t encType    : 2;    //< Encryption type
-        uint16_t encSubType : 2;    //< Encryption subtype
-        uint16_t CAN        : 4;    //< Channel Access Number
-        uint16_t            : 4;    //< Reserved, padding to 16 bit
-    }
-    fields;
+        m17_lsf_packet_stream_indicator stream : 1;  //< Packet/stream indicator: 0 = packet, 1 = stream
```
turnrye, line 124: Please align the bit field values in the same column like they were previously.

openrtx/src/rtx/OpMode_M17.cpp
```diff
@@ -224,8 +224,8 @@ void OpMode_M17::txState(rtxStatus_t *const status)
         if(!dst.empty()) lsf.setDestination(dst);
 
         streamType_t type;
-        type.fields.stream   = 1;             // Stream
-        type.fields.dataType = 2;             // Voice data
+        type.fields.stream   = M17_LSF_STREAM_MODE;             // Stream
```
turnrye: Please align the comments like they were before.

openrtx/include/protocols/M17/M17Datatypes.hpp
marcoSchr: This is currently only a draft. I will align the code, once it is finished.

Issue #173, opened 2023-08-21
CI: Testing the code by building it


marcoSchr:
It could be useful to test the code by compiling it for all targets.

# Current state

At the moment the firmware is only build when new code is pushed to master.

# Proposed Change

Build the firmware for all Pull requests

# Benefits

- We can see potential issues before a Pull request is merged
- We can see issues with specific targets because not everyone will test their changes with all targets

# Disadvantages

- Pull request pipelines will run longer
- More resources will be used for building the code every time

silseva:
Works for me!

silseva:
Implemented by PR #174, now merged in master.

PR #174, opened 2023-08-22
Add a "Build" job

turnrye:
# What

Add a CI job that creates builds. This replaces the existing GitLab actions and consolidates them into one place. Bins are still made for primary, but they're now made for all targets. They're also made for each push event and pull request, not just master.

# Why

This way its quicker for some users to test. It also could replace the nightly build system, though a limitation of this setup is that GitHub actions artifacts only supports creating a singular zip file (what an odd limitation).

# How

- Combine the YAMLs
- Add a new job, based on the other two, that creates builds

Ref https://github.com/OpenRTX/OpenRTX/issues/173


.github/workflows/main.yml
```diff
@@ -1,44 +1,102 @@
...
+      - name: Compile linux
+        run: meson compile -C build_linux openrtx_linux
+      - name: Compile arm targets
+        run: |
+          export PATH=$PATH:/tmp/OpenGD77/tools/Python/FirmwareLoader:/tmp/OpenGD77/firmware/tools
+          echo $PATH
+          meson compile -C build_arm openrtx_md3x0_wrap
+          meson compile -C build_arm openrtx_mduv3x0_wrap
+          meson compile -C build_arm openrtx_md9600_wrap
+          meson compile -C build_arm openrtx_gd77_wrap
```
turnrye, line 56: @n1zzo the gd77 wrapping isn't working despite [having these steps satisfied](http://openrtx.org/#/compiling?id=gd-77-and-dm-1801). Can you take a look at the commands that are being run here and the output of the build to shed some light on what may be happening?

antoniovazquezblanco:
From my experience, it is useful to make CI builds verbose. My rationale is that if everything builds you won't probably look at the verbose text and thus it doesn't bother you, but when it fails, you may want it to be there :D.

It is just adding `-v` to the meson compile command.

.github/workflows/main.yml
cifred98: Not sure what error you are getting but now there isn't the need anymore to have the OpenGD77 binaries in the PATH since they are in the scripts directory but the doc wasn't updated. Soon there should support for SGL files in radio_tool, so we'll move to that for the GD-77 & co. in the future.
You should remove the opengd77 step, the export of that PATH and fix the release-bins wildcard to include sgl files they have a ".sgl" extension, otherwise this PR looks good to me

turnrye:
Thanks so much for the feedback! I’ll make these changes. On Fri, Oct 6, 2023 at 5:34 PM, Fred ***@***.***> wrote:  
@cifred98 commented on this pull request.



In .github/workflows/main.yml:
> +          git clone https://github.com/open-ham/OpenGD77.git /tmp/OpenGD77
+      - name: setup meson
+        run: |
+          cd ${{github.workspace}}
+          meson setup build_linux
+          meson setup --cross-file cross_arm.txt build_arm
+      - name: Compile linux
+        run: meson compile -C build_linux openrtx_linux
+      - name: Compile arm targets
+        run: |
+          export PATH=$PATH:/tmp/OpenGD77/tools/Python/FirmwareLoader:/tmp/OpenGD77/firmware/tools
+          echo $PATH
+          meson compile -C build_arm openrtx_md3x0_wrap
+          meson compile -C build_arm openrtx_mduv3x0_wrap
+          meson compile -C build_arm openrtx_md9600_wrap
+          meson compile -C build_arm openrtx_gd77_wrap

Not sure what error you are getting but now there isn't the need anymore to have the OpenGD77 binaries in the PATH since they are in the scripts directory but the doc wasn't updated. Soon there should support for SGL files in radio_tool, so we'll move to that for the GD-77 & co. in the future.
You should remove the opengd77 step, the export of that PATH and fix the release-bins wildcard to include sgl files they have a ".sgl" extension, otherwise this PR looks good to me

silseva:
Thank you! :)

PR #176, opened 2023-08-23
WIP: Input lockout

tarxvftech:
Users sometimes need a way to prevent accidental channel changes. On many radios there is usually some mechanism to lock the keypad, often by holding a button or combination of buttons to lock and unlock. On some radios - like TYT OEM - this is done in such a way that the channel knob still changes the channel, utterly defeating the purpose.

This locks out all input except PTT and the macro menu, allowing for unlock the same way as lock. Volume is independent and therefore not effected, though I wouldn't mind having an option to lock that too at some point.

Draft because it still needs UI indications. I know there was a lot of discussion around that so I didn't want to jump ahead (and for my immediate usage, I don't need it :grin:  )

I was also considering the usage of the '*' key instead, as on TYT radios that has the lock icon, but I imagine that will change too often across platforms to be worth chasing.

### DIFF
--- a/openrtx/include/ui/ui_default.h
+++ b/openrtx/include/ui/ui_default.h
@@ -196,6 +196,7 @@ typedef struct ui_state_t
     uint8_t menu_selected;
     // If true we can change a menu entry value with UP/DOWN
     bool edit_mode;
+    bool input_lockedout;
     // Variables used for VFO input
     uint8_t input_number;
     uint8_t input_position;
--- a/openrtx/src/ui/default/ui.c
+++ b/openrtx/src/ui/default/ui.c
@@ -81,7 +81,7 @@
 
 /* UI main screen functions, their implementation is in "ui_main.c" */
 extern void _ui_drawMainBackground();
-extern void _ui_drawMainTop();
+extern void _ui_drawMainTop(ui_state_t* ui_state);
 extern void _ui_drawVFOMiddle();
 extern void _ui_drawMEMMiddle();
 extern void _ui_drawVFOBottom();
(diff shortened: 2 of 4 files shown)


tarxvftech:
2023-09-10-002539_480x384_scrot
2023-09-10-002524_480x384_scrot
2023-09-10-002455_480x384_scrot
2023-09-10-002443_480x384_scrot

tarxvftech:
I've moved the mode indicator to the channel details line, and massively cleaned up the channel details line. I am opinionated and like these, but these are presented for your consideration - the main thing is just about getting the space in the top left clear for the lock icon.

I also did a bunch in trying to make some of the frequency display a little more familiar for hand-input repeater usage - for example, 147.375+ will be rendered as such, with knowledge of standard repeater offsets per band.
This expects that standard offsets are global - I don't know if they are. I got carried away, because I actually like my channel details to be compressed and all on one line like I have for dmr.tools codeplug editing and this was a big step towards that. 

Perhaps this will be more a matter of taste! But, since I promised mockups these are them! They just have working code, too.

If the Ui changes are not immediately loved but moving the mode string down to the details is acceptable, I'll of course split the extra stuff to a different branch and make just the smallest changes necessary to make space for the lock icon.

n1zzo:
Merged into dedicated future branch to perform final testing, will move to master soon.

Thanks for the contribution! ;)

tarxvftech:
@n1zzo Closing since you have it and are working on the input side of things too, nizzo. 
I did find a bug in the frequency display - if you put in "430.000" you get "43" displayed. Oops!

PR #181, opened 2023-09-01
Better sync for M17 demodulator and Module17 dst & src update

Paulo-D2000:


sp5wwp:
I can't see any callsigns being displayed on my Module17. `SYNC` LED doesn't blink - that's good - but the speech is intermittent, still, even with a strong signal.

silseva:
The sync issue seems to be due to some bug in the demodulator and need more investigations. The display of src and dst callsign have been merged into master from Marco's pull requests. Thanks anyways for your work!

PR #183, opened 2023-09-05
Add UI and check for CAN on RX

### LINKED ISSUE #168: Feature: M17: Support CAN on RX
At the moment we only support CAN on tx M17.
When receiving M17 we ignore the CAN.

We should only play the M17 reception, when the CAN matches our configured CAN.
We should still show the green RX LED in order for the user to see, that a transmission is received, even when not outputting anything in order to prevent collision.
Additionally an option to disable this should be available.
This can be put inside the macro menu, similar to how RX and TX tones are handled for FM.

marcoSchr:
This adds the UI and logic to check the CAN on M17 RX.
It can be enabled and disabled in the settings menu.
When enabled, sound will only be played when the CAN matches the one set.
The LED will still light up.
The changes are for the default UI and the Module17 UI.
The Module17 UI needs to be tested.

This will fix #168

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -94,7 +94,8 @@ openrtx_ui_default = ['openrtx/src/ui/default/ui.c',
 
 openrtx_ui_module17 = ['openrtx/src/ui/module17/ui.c',
                        'openrtx/src/ui/module17/ui_main.c',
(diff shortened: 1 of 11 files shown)


silseva:
Thank you!

PR #193, opened 2023-09-22
Check if the battery voltage is below minimum

### LINKED ISSUE #192: Battery Calculation might underflow
My RT3S turned off today when I tried to transmit and it would then no longer turn on.
I measured the battery voltage and got 0V.
After putting the battery into the charger for a few seconds, it returned with about 6.2V.
When I then turned on the device, I could see in the Info screen a voltage of 6.2V and the battery percentage showed 100%.
I investigated the battery calculation source code and found that it could underflow.

When the battery voltage is lower than the assumed minimum voltage, the calculation will underflow. [1]
This leads to the percentage calculation assuming the battery is 100%.  [2]

[[1] ](https://github.com/OpenRTX/OpenRTX/blob/master/openrtx/src/core/battery.c#L71)
[[2]](https://github.com/OpenRTX/OpenRTX/blob/master/openrtx/src/core/battery.c#L76)

marcoSchr:
We check if the battery is below minimum and assume it is 0% This is done to prevent an underflow which would result in showing 100%.

This will fix #192

### DIFF
--- a/openrtx/src/core/battery.c
+++ b/openrtx/src/core/battery.c
@@ -68,6 +68,12 @@ uint8_t battery_getCharge(uint16_t vbat)
     vb          = vb / 1000;
     vb          = (vb + 256) >> 8;
 
+    /*
+     * If the voltage is below minimum we return 0 to prevent an underflow in
+     * the following calculation
+     */
+    if (vb < bat_v_min) return 0;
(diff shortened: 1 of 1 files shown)


silseva:
Thanks! :)

PR #194, opened 2023-09-25
Enable edit mode for dst callsign on mod17

marcoSchr:
This allows editing the destination callsign on the module17 When on the main screen, the right arrow will enter edit mode.


silseva:
Thanks!

PR #204, opened 2023-10-24
Fix module 17 coompilation with Callsign display in splash

marcoSchr:
This will fix compilation on the Module 17.


silseva:
Thanks!

PR #208, opened 2023-11-17
Add a volume bar in the bottom over the other bars

### LINKED ISSUE #190: Feature: Show Volume on Screen
It would be useful to show the current volume on screen.
This could be done in two ways.

1. Showing the Volume permanently on screen.
2. Show the Volume only when changing it.

marcoSchr:
This adds the current volume of the speaker as a bar above the S meter in the bottom.
The function to draw the meters has the option to not draw the volume bar, which at the moment is only used on the module 17 as we do not have a way to get the volume.

This will fix #190


n1zzo:
Reviewing this now.

n1zzo:
Tested working on T-TWR Plus, this is ready to be merged!

PR #212, opened 2023-11-19
Improving PPM setting menu

mdiepart:
This PR is in the continuation of #202 and improves the handling of negative PPM numbers and re-uses as much existing code as possible.


sp5wwp:
Imagine the calibration data being finally used properly. What would be the use case for this menu entry then?

silseva:
Probably of no use. Still worth keeping the code for possible targets requiring it.

PR #213, opened 2023-11-19
Add callsign squelch for incoming M17 transmissions

mdiepart:
This PR makes it so that M17 transmission with a destination callsign that is neither ALL nor broadcast will not be heard.

The way it checks for a callsign is that it removes an eventual prefix that may be there but should be less than or equal to 2 characters long. This makes it so that XY/AB1ABC will match AB1ABC and vice versa (but AB1ABC/P will not match AB1ABC).
If this behavior is not desired I can remove it.

### DIFF
--- a/openrtx/include/rtx/OpMode_M17.hpp
+++ b/openrtx/include/rtx/OpMode_M17.hpp
@@ -121,6 +121,19 @@ private:
      */
     void txState(rtxStatus_t *const status);
 
+    /**
+     * Computes if the destination callsign matches with the user's callsign.
+     * The comparison does not take into account the country prefixes (strips the '/'
+     * and whatever is in front from all callsigns). It does take into account the
+     * dash and whatever is after it. It will also return true if the destCallsign
+     * is "ALL".
+     *
+     * \param myCallsign plain text callsign from the user
+     * \param destCallsign plain text destination callsign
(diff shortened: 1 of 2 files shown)


lemielek:
Why not /p or /m or /mm ?

sp5wwp:
Add regex support for the most demanding users :)

marcoSchr:
There should also be an option to disable this feature, to receive all transmissions. Similar to how the M17 CAN check can be enabled/disabled.

mdiepart:
The "/p" case I gave is just an example, to further explain that only the prefix gets removed if it is 2 characters long or less. I have to say that I made this choice quite arbitrarily. Feel free to propose something else.
@sp5wwp Feel free to propose an implementation using regex 😄 
I agree that we should add a setting to enable/disable the CS squelch. I'll try to see if I can add that soon.

silseva:
> There should also be an option to disable this feature, to receive all transmissions. Similar to how the M17 CAN check can be enabled/disabled.

A possible alternative for this is disabling the callsign check if the destination callsign is set to "ALL".

mdiepart:
IMO this should be avoided as the feature are not really related. I.E. If I call CQ and a conversation between two people (with DEST set) is in progress, it does not make sense to open the squelch...

silseva:
Thanks!

PR #218, opened 2023-12-23
ttwrplus: Ignore revision from sa8x8-fw version

### LINKED ISSUE #217: T-TWR Plus: Init failure due to sa8x8-fw version mismatch
Symptom: A TWR-Plus programmed with latest sa8x8-fw release image (1.3.0-r0) will not fully boot and instead hangs at the OPNRTX logo. This effect is similar to one from #199 except the firmware is the expected one rather than the unsupported factory NiceRF firmware.

This issue is caused by how the firmware version string is evaluated, where the "revisions since last tag" field is used to determining compatibility rather than being used as a purely informational field.

edgetriggered:
The "revisions since last tag" version field is now accurately named and no longer checked. Extra parentheses also added to reduce compiler warnings.

Fix #217

### DIFF
--- a/platform/drivers/baseband/SA8x8.c
+++ b/platform/drivers/baseband/SA8x8.c
@@ -30,7 +30,6 @@
 #define SA868FW_MAJOR    1
 #define SA868FW_MINOR    3
 #define SA868FW_PATCH    0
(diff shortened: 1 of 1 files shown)


silseva:
Thanks!

PR #219, opened 2023-12-23
rt-890: Add UART code from OEFW, UART works

Tunas1337:


silseva:
Hi and thanks for your help!
I'm sorry but I'll not merge this PR because in the end I found out the root cause of the problem: for some reasons, the value of the `SystemCoreClock` variable (provided as per CMSIS specification) was remaining set to 8MHz even though the clock initialization function  updated it to 120MHz. Changing the default init value to 120'000'000 solved the issue.
We also have to consider that `SystemCoreClock` is also used to initialize the SysTick timer, so a wrong value would have caused even more problems. See 667f8aeb865e2956877f04edf103b445acc81583 for the details.
Your code gave me a great help to find out where the problem was, though!

Thanks,
IU2KWO.

Tunas1337:
No problem! Glad to have helped.

Closing as the mission was accomplished anyway :)

PR #223, opened 2023-12-30
Allow skipping key lock when only hash key is pressed

marcoSchr:
This improves 3a2bac3e7e0670d5e5f743d465cb10678c5cd7e8 by only skipping the key lock when the only key pressed is the hash key.

This fixes some issues where it was possible to get key presses past the key lock when holding the hash key.


silseva:
I'm sorry but your solution is logically equivalent to the previous implementation and doesn't solve the problem. To reproduce it, press the enter key while keeping pressed the hash key.
The correct fix is to have `msg.keys == KEY_HASH` which bypasses the lock only when the hash key is pressed alone. I'll merge your contribution anyways and amend the commit.

PR #224, opened 2023-12-30
Multiple linux targets

marcoSchr:
This PR separates the linux targets.
This allows to build 3 different linux executables.
The old linux target.
The `_smallscreen` target uses the smaller screen size and the black and white screen config.
The `_mod17` target behaves like the Module 17 0.1e.
For this the hardware config of the linux target was adjusted to not include the GPS and RTC defines.
They are now defined for each target separately, as the Module 17 does not have those.
Additionally the hwversion for the linux build is set to 1.
This does not change anything for the default UI, but in the mod17 UI it shows the menus like revision 0.1e.

I am open for better naming of the default ui targets.

Once this is merged, I can update the documentation for the website.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,8 +36,8 @@ jobs:
           meson setup --cross-file cross_arm.txt build_arm
       - name: Compile linux
         run: |
-          meson compile -C build_linux openrtx_linux_default_large
-          meson compile -C build_linux openrtx_linux_default_small
+          meson compile -C build_linux openrtx_linux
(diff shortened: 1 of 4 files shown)


silseva:
I suggest to do the following:
- leave the original build target `openrtx_linux` for the version with the large screen in order to keep things consistent with the past;
- rename the target for the small screen  to `openrtx_linux_smallscreen.

Thanks!

marcoSchr:
I have updated the `meson.build` file to use this naming.

silseva:
Thank you!

PR #225, opened 2023-12-30
Add button info to module 17 callsign entry

marcoSchr:
This  adds some text to the bottom of the screen when editing callsigns on the module 17 to show the function of the enter and escape buttons.


silseva:
Thanks!

PR #228, opened 2024-01-23
Add flash to ttwrplus

pwgen:
Enable NV/Flash write store  operations for ttwrplus


pwgen:
when no channel information exists on the flash  the device will not boot.
there must be some trick to identify incalid channel information .

PR #236, opened 2024-02-04
Update github action checkout and upload-artifect to v4

marcoSchr:
The v3 actions use Node 16 which is now deprecated and will be phased out. To fix this we will use the v4 actions which are based on Node20. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/


silseva:
Thank you!

PR #237, opened 2024-02-05
Allow git commit as version to fix artifacts not showing any version info

marcoSchr:
At the moment the github action artifacts have no version listed on the info screen. We use the `--always` flag, as this allows `git describe` to show only the commit if nothing else is available.

The alternative solution would be to always checkout more of the repository, instead of only a single commit.


silseva:
Thanks! :)

PR #238, opened 2024-02-11
Fix compiler warnings for linux and mduv3x0 targets

Derecho:
Whilst following the instructions for compiling for the targets mentioned in the title, I noticed some compilation warnings. This didn't prevent the compilation to succeed, but I figured I'd see if I can help by tidying these up.

All warnings during normal compilation for these targets are now gone for me with this changeset. If any clarification or rationale is needed for the changes I'll happily elaborate. If these's any contribution guidelines I've missed and violated, please let me know.

I do not currently have hardware to test on.


silseva:
Thank you!

PR #240, opened 2024-02-16
Added User Functions 

mdiepart:
This PR contains one commit that implements a user functions thread. 

This thread handles a set of functions added by the user.
Those functions can be either periodic (with a frequency of 1Hz, 10Hz, 20Hz, 50Hz, 100Hz or 1kHhz) or aperiodic (asynchronous).  
Periodic functions will be called periodically depending on their scheduling. Async functions need to be triggered using user_functions_trigger(id).

The maximum amount of user functions that can be handled have to be defined in hwconfig.h using the macro CONFIG_USER_FUNCTIONS. If this macro is not defined, the user functions thread will not be created.

For now, it is implemented by checking for async functions to call every 5ms. This is because the version of miosix used does not support `pthread_cond_timedwait()`.

Some code (untested, commented-out) have been voluntarily left in the `user_functions.c` waiting for the miosix kernel to be updated.


openrtx/include/core/user_functions.h
```diff
@@ -0,0 +1,60 @@
...
+
+#ifndef USER_FUNCTIONS_H
+#define USER_FUNCTIONS_H
+
+#include <hwconfig.h>
+#include <stdint.h>
+#include <stddef.h>
+#include <errno.h>
+
+#if defined(CONFIG_USER_FUNCTIONS) && (CONFIG_USER_FUNCTIONS > 0)
```
silseva: I think this guard is not needed, because:
- target-specific code using the user function system, will for sure need to define the `CONFIG_USER_FUNCTIONS`
- all the function calls in the common code have to be guarded with `#ifdef` to avoid linker errors in case the target does not support the user functions (see threads.c)
- functions not called anywhere and unused variables are garbage-collected at link time

openrtx/include/core/user_functions.h
```diff
@@ -0,0 +1,60 @@
...
+typedef enum {
+    USER_FUNCTION_ASYNC = 0,
+    USER_FUNCTION_100HZ = 10,           // 10 ticks
+    USER_FUNCTION_50HZ  = 20,           // 20 ticks
+    USER_FUNCTION_20HZ  = 50,           // 50 ticks
+    USER_FUNCTION_10HZ  = 100,          // 100 ticks
+    USER_FUNCTION_1HZ   = 1000,         // 1000 ticks
+} user_functions_sched_t;
+
+void user_functions_init();
```
silseva, line 57: Please add doxygen-style documentation to the function prototypes. :)

openrtx/include/core/user_functions.h
```diff
@@ -0,0 +1,60 @@
...
+void user_functions_terminate();
+void user_functions_task();
+error_t user_functions_add(void (*f)(), uint8_t id, user_functions_sched_t scheduling);
```
silseva: What about adding also a `void *arg` to be passed to the function, when called?

openrtx/include/core/user_functions.h
mdiepart: I can remove the guard if needed. But if user functions are expected to be used in common code then it is not really user functions anymore and we should also modify slightly the mechanism to account for functions added from common code.

openrtx/include/core/user_functions.h
mdiepart: Will do

openrtx/include/core/user_functions.h
mdiepart: Great idea, will do

openrtx/include/core/user_functions.h
silseva: I called them "user functions" following the terminology used in the code I used at work, but a more correct term should be "tasklet" or something like that. About adding periodic functions from common code, I see that the current `gps_task` is a good candidate for this mechanism: if there is the GPS hardware, or if the GPS is enabled, add a periodic function which retrieves and parses the NMEA sentences.

silseva:
This depends on miosix being updated to the latest version.

PR #239, opened 2024-02-16
Update Miosix Kernel

mdiepart:
While implementing user functions (functions that user code can register in OpenRTX to be called periodically or asynchronously to avoid creating a thread for minor functionalities) I realized that the function `pthread_cond_timedwait` is needed.

It looks like that function is available in recent versions of miosix. Updating the kernel would allow to implement user functions. It is also probably gonna be needed for OpenHT.

This PR updates the miosix kernel, then fixes the code for Module17. This PR does not work for all targets yet.

One of the main problem that was found in the update process is that the default miosix implementation for the system timer (TIM5) uses an uninitialized TimeConversion instance. Commit [86fce68](https://github.com/OpenRTX/OpenRTX/pull/239/commits/86fce684a41ec6bbb93bd081b8ace8e4d489a968) fixes it. 

As @silseva noted, a system timer implementation needs to be implemented for the MK22 processor.


mdiepart:
I tried flashing my RT3s radio (with a copycat MCU). It compiles and flashes alright, but it doesn't boot.

I explored a bit, and I find that there probably is a problem with miosix. In particular, the function `startKernel()` returns despite the comment right after saying that we should never reach there. I tested that by I turning on the green LED, turning on the PWR_SW pin and entering an infinite loop.

For now I don't have many clues on why it could be.

mdiepart:
I wrote a preliminary implementation for MK22 MCUs. It is untested as I don't have a GD-77 or MD-9600.

mdiepart:
I rebased the branch on the current Master branch. 
I completed/fixed the NXP timer implementation. For now, it is using FlexTimer 0 (FTM0) with a prescaler of 4.

mdiepart:
I rebased the branch on main. 
This PR now compiles on every target and runs fine on MD-UV3x0 and Module17. A simplified version of OpenRTX also runs fine on the NXP MK22FN512 devboard. I don't have any other targets to test it on though.

mdiepart:
I updated the miosix source files to the latest tagged release

mdiepart:
Miosix is now a submodule of OpenRTX. 
The successive commits describe the various changes that were brought to the code. 
The miosix fork is currently my fork and would have to be updated to the one OpenRTX uses.

All the targets compile. I have successfully tested OpenRTX on my Module17 (rev 1.0) and Retevis RT3s.

I haven't added the NXP MK22 serial driver to openrtx yet.

In the last commit, I realized that the vcom wasn't being initialized anymore (as it was initialized in bspInit2 function from the modified miosix kernel). 
I added a callback to the bsp in miosix and defined the callback in the same file as the SystemInitHook() function.

Let me know what you think of this and if there is anything I should do.

silseva:
Finally! :)

PR #246, opened 2024-02-18
Add smallscreen and mod17 linux binaries to artifacts

marcoSchr:
This adds the already build smallscreen and mod17 linux binaries to the github actions artifacts.


silseva:
Thanks!

PR #258, opened 2024-03-31
Change how gps data is stored and fix gps display

### LINKED ISSUE #241: GPS Display broken after switching to sniprintf
# Observed behavior 

When the GPS gets a fix, the display does not show the current position

# Expected behavior

When the GPS gets a fix, the display does show the current position

# Steps to reproduce

1. Enable GPS
2. Go to GPS screen
3. Wait for fix

# Additional Information

I have tested this on a Retevis RT3S, but I assume it is broken on all devices with GPS support.
I have used git bisect to find the commit, which introduced the bug: 1791e70eef0126c227519b0fec2749635b02ae65.

photo_2024-02-18_10-57-47

marcoSchr:
This will fix the display of the GPS data.

It changes the internal data types of the GPS data.
It adds a test for the conversion function.

This will fix #241

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -80,6 +80,8 @@ jobs:
         run: meson test -C build "M17 RRC Test"
       - name: Codeplug Test
         run: meson test -C build "Codeplug Test"
+      - name: minmea Conversion Test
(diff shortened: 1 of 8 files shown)


silseva:
Thank you!

PR #259, opened 2024-03-31
Adding support for Module 17 revision 1.0

mdiepart:
Hello !

This pull-request implements many changes needed on the Module17 implementation to support the upcoming revision 1.0.

I did my best to keep the code clean and follow the coding style in use in OpenRTX. Let me know if any changes are needed.

The commits have been squashed in a way to make clear the way I implemented everything and the changes brought on the codebase, module by module.

Most importantly:
- I faced the need to include more hardware info than the one specified in the HwInfo_t structure so I added a pointer to through which any device can add a custom structure containing additional info that is only relevant to this specific device.  
- I added hardware drivers for I2C (peripherals 1 and 2). I2C2 has additional SMBus support (incomplete for now as it does not support yet the SMBALERT# line assertion, it would be ignored as of now). Also, I implemented many functions described by the SMBus standard but was not able to test them all. The ones in use by the capacitive touch sensor for the HMI keyboard have been tested of course. The others have not. The version of the SMBus implemented is version 2.0 as the SMBus part of the I2C driver in the STM32 MCUs is not compatible with SMBus 3 (mostly, it is not able to manage variable size payloads that can be 0 bytes, Minimum variable size payload is 1 byte).
- I also split-up the display driver of Module 17 to be able to support different screens depending on the hardware configuration. The screen driver to use depends on the hardware revision and the presence or not of an HMI. 
- About menu shows the hardware revision but also whether or not an HMI was detected. It also shows if the soft potentiometers were detected on the I2C bus or not.

Future work on Module17 may include:
- measuring and displaying the voltage on the vbat pin and enable/disable RTC support based on that
- Supporting SMBALERT# signal assertion in SMBus driver

Thank you for the help provided along the way!


mdiepart:
The checks just revealed that the new pointer added in the HwInfo is not handled properly when compiling for linux with the Module 17 UI. I'll try to have a look into that and see if I can come up with a fix.

mdiepart:
I added two fixup commits dealing with the Linux compilation of the Module17 UI.

silseva:
Thank you!

PR #268, opened 2024-04-26
Module17: Fix I2C2 ER IRQ handler not linking

mdiepart:
This fixes an issue where Module17 would crash if I2C2 error interrupt was thrown. This was due to the function being declared in C and the linker not being able to link it to the symbol declared in the NVIC vector table. Thus the default handler was executed. 

This declares a symbol with the C++ mangled name and aliases it to the C function.

In the process it was found that the timeout variable used in the I2C2 SMB driver was never reset. This was fixed too.


silseva:
Thanks!

PR #278, opened 2024-07-17
Talkpod A36Plus support

Jamie793:
So far. I have implemented BK4819 and BK1080. It can TX and RX. The BK4819 function includes "CDCSS, CTCSS, Scan Frequency, VOX and so on". The BK1080 function includes  "Tune".


froqstar:
Can/will this target eventually support M17, and if so will it require a hardware mod?

Tunas1337:
> Can/will this target eventually support M17, and if so will it require a hardware mod?

It's only an analog radio, so it would be an interesting feat to make M17 work on it. The answer is "maybe." I'll try to get Codec2 working on it for the voice prompts first of all, and then check whether the MCU already has a mic path into the BK4819. If not, yes, it would require a hardware mod.

silseva:
To be evaluated what to do, not stale

turnrye:
I tried to look into parsing this out into pieces, but it seems like the key non-hardware-specific parts are actually already pulled out. BK4819 and BK1080 drivers are being added in https://github.com/OpenRTX/OpenRTX/pull/510, the framebufferless graphics is on a branch `RT-890` for investigation at some future time. I do not have an A36Plus on hand to b able to try to carry this work across the finish line or to help test and maintain it over time unfortunately.

So, at this point, I am closing this PR. But if anyone is interested in resurrecting the A36Plus hardware support part of this change, please do!

PR #279, opened 2024-09-17
pulse-simple is not needed to link openrtx_linux; implement CAT commands to set frequencies

mbalmer:


mbalmer:
Added CAT command to set frequencies.

PR #281, opened 2024-09-23
Define hex constants for CAT commands

mbalmer:
Define CAT command in an enum instead of hardcoding them in the code.  While here, bump the copyright year.


mbalmer:
Added a second commit to make the code more safe wrt the use of strncpy.

PR #284, opened 2024-10-04
Implement a CAT command to control PTT

mbalmer:
This implements the GPT (get-ptt) and SPT (set-ptt) CAT commands.

You can enable/disable PTT over CAT.  If you enabled it over CAT, and then press and release the PTT button on the real device, it will stop transmitting (button overrides CAT).  If you press the hardware PTT button, and send the CAT command to disable PTT, nothing happens, i.e. the hardware PTT button always "wins" over the CAT command.

Tested on openrtx_linux with trx-control.

### DIFF
--- a/openrtx/include/core/state.h
+++ b/openrtx/include/core/state.h
@@ -60,6 +60,7 @@ typedef struct
     bool       backup_eflash;
     bool       restore_eflash;
     bool       txDisable;
+    bool       ptt;
     uint8_t    step_index;
 }
 state_t;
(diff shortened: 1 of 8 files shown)


silseva:
Please propagate the change also in the M17 opmode handler.

mbalmer:
Sure.  I just need to first carefully read and try to understand that one.

mbalmer:
The PTT logic has been added to M17 and tested with openrtx_linux.

silseva:
Commits merged by hand, thank you!

PR #294, opened 2024-11-05
Fixed uninitialized move in Synchronizer

mdiepart:
While testing M17Netd code under valgrind (which uses OpenRTX's M17 demodulator), it appeared that there are some jumps or moves depending on uninitialized values. This all comes from the `Synchronizer.hpp` file.

I chased it down to:
1. Variable triggered never initialized in `Synchronizer`.
2. In function `Synchronizer::update()`: the return value for the sign of the sample index is read one past the end of the `values` array

The proposed fixes are:
1. Initialize `triggered` to false in the constructor
2. Return the sign of the `peak` variable (which is equivalent to the `values[sampIndex]` variable which is probably what the code should have been reading in the first place?

I'm not 100% sure on the proposed fix for the second issue so please counter-check my work here.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -82,6 +82,7 @@ openrtx_inc = ['openrtx/include',
                'platform/drivers/USB',
                'platform/drivers/GPIO',
                'platform/drivers/tones',
+               'platform/drivers/audio',
                'platform/drivers/baseband',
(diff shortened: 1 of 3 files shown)


silseva:
Thank you!

PR #295, opened 2024-11-13
RSSI Fix for +10/+20db - Tested in SIM

Trriss:
+10db/+20db now correctly shown
Added casting: (rssi_t)(127 + rssi) / 6 -> So that "rssi=-121" corresponds with "s_level=1"


silseva:
Thank you!

PR #298, opened 2024-11-19
MD9600: Fix Keyboard not working after ADC changes

marcoSchr:
The MD9600 microphone keyboard was non functional after the ADC platform changes, due to changes on how to calculate the row and column voltage.

This pull request fixes this by first calculating the value and then casting to uint16.
Before, the adc value was larger than uint16 and got cut off, which resulted in the keyboard not working.


silseva:
Thank you!

PR #300, opened 2024-12-24
Add MD-UV3x0 Plus target

marcoSchr:
This will add a target for the MD-UV3x0 Plus.

The plus version is the same as the non plus version, just with two pins swapped.
Therefor only a define for the plus is added and checked in the pinmap in order to flip the two pins.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -45,6 +45,7 @@ jobs:
           echo $PATH
           meson compile -C build_arm openrtx_md3x0_wrap
           meson compile -C build_arm openrtx_mduv3x0_wrap
+          meson compile -C build_arm openrtx_mduv3x0p_wrap
           meson compile -C build_arm openrtx_md9600_wrap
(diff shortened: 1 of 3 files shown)


silseva:
Thank you!
Just one thing: it would be interesting to compare the content of the radio's external flash security register at address 0x3000 with the one of a "plain" UV380. Probably this small hardware difference is codified in somewhere.

lemielek:
RT3S als6?
How to do that?

silseva:
The retevis RT3s is already supported by the `mduv3x0` target, the hardware is identical

silseva:
@marcoSchr should I merge the PR or we want to wait until also the GPS issue gets fixed?

marcoSchr:
We can merge it now in its own branch.
This should make it easier for others to test it and maybe find other issues beside the GPS and memory.

silseva:
Merged, thank you!

PR #303, opened 2024-12-28
Add github workflow for zephyr targets

### LINKED ISSUE #242: Add T-TWR Plus to actions build
At the moment we are building most targets with a GitHub action.
It would be useful to also add the T-TWR Plus to this list.
For this a new step in [.github/workflows/main.yml](https://github.com/OpenRTX/OpenRTX/blob/master/.github/workflows/main.yml) for the build would be needed.

marcoSchr:
This adds a new build step to the github workflow.
This build step will build the t-twr plus zephyr target.
It uses the `zephyrproject-rtos/action-zephyr-setup@v1` action to setup the zephyr sdk.

The resulting uf2 file is uploaded as an artifact.

This will fix #242


silseva:
Thank you!

PR #306, opened 2024-12-29
Keyboard handling improvements

sp6hfe:
This PR extend support for keyboard handling.

By resigning from (not used) KEY_F6 support 4 button events are accessible to get more grained TRX control.
Previous implementation allowed to differentiate between a press and long press (no support for single/multi-press and repeats)

Based on timings it is possible to detect
- a press (a key/keys press detected since last keyboard scan)
- a single-key long press (key depressed for some time)
- a multi-key long press (1+ keys depressed for some time)
- a repeat (which is a periodic event that happens after long press, but reserved for single keys only)

The idea behind is at first to distinguish between single/long presses and repeats to not widely take an action when we don't want to. Secondly a long multi-press event is useful for some special actions requiring keys combination (like entering a hidden menu or enabling/disabling an option/taking a defined action). Events allow for quick filtering of them when needed.

Knob movements are handled in parallel and due to the nature of the knob change signal (can not be detected as long press or repeat) is not involved much into the keyboard events logic. If you want to use knob direction just look up them in an event pressed keys.

Multi-key support is working by default in a way that whenever an additional (1+) key is pressed no event is generated until the long-press timeout occur. If any additional key is pressed before long-press is detected a timer restarts and upcoming long-press event (multi-press) will report all the keys pressed.
Max number of multi-pressed keys is 3. When more is detected there will be no further action until keyboard is fully released.
If during multi-press a change is detected by lowering number of pressed keys or by keeping the same amount but in a different combination no further action is possible until keyboard is fully released.

Additionally a special mode was added due to how Macro Menu handling is currently implemented. It is possible to generate press events for each newly depressed key when already in a multi-key situation. This allow for changing macro options without actually latching the menu itself.

Improvements proposed in this PR fix some behaviors detected like:
- pressing long a digit in VFO mode would add 2 digits to the frequency entering form instead of 1,
- pressing long up/down in the menus change menu position by 2 instead of that many as long as button is kept pressed,
- pressing ENTER long advances in sub-menus by 2 instead of 1,
- pressing and holding F2 latches Macro Menu even though it is used with a combination with a digit to quickly change an option without actually a need to latch it.

### DIFF
--- a/openrtx/include/core/input.h
+++ b/openrtx/include/core/input.h
@@ -1,5 +1,5 @@
 /***************************************************************************
- *   Copyright (C) 2020 - 2024 by Federico Amedeo Izzo IU2NUO,             *
+ *   Copyright (C) 2020 - 2025 by Federico Amedeo Izzo IU2NUO,             *
  *                                Niccolò Izzo IU2KIN,                     *
  *                                Silvano Seva IU2KWO,                     *
  *                                Grzegorz Kaczmarek SP6HFE                *
@@ -76,23 +76,26 @@ kbd_msg_t;
 
 /**
  * Turn on or off a support of press event generation on
- * new key pressed while 1+ keys is pressed at the time
- * This is used when some keys stay pressed while
- * others are changed
+ * new key pressed while 1+ keys pressed at the time
+ * This is used by UI when some key/s stay pressed while
+ * other/s change
  * 
- * @param isPressEventAllowed: press events support during multi-press
+ * @param isEnabled: feature enabling flag
  */
-void input_allowPressEventOnMultilpeKeysPressed(bool isPressEventAllowed);
+void input_enableMultiKeyPressEvent(bool isEnabled);
 
 /**
- * @brief 
+ * Looks up knob events only in provided keyboard_t data
+ * Useful in a situation where no real keys are pressed
+ * (keys detection is handled separately) 
  * 
  * @param keys keys pressed
  * @param msg keyboard event message to be filled
  * @return true msg contain valid event message
  * @return false msg contents are not valid
  */
-bool processKnobMovementDetection(keyboard_t keys, kbd_msg_t* msg);
+bool input_handleKnobEvents(keyboard_t keys, kbd_msg_t* msg);
+
 /**
  * Scan all the keyboard buttons to detect possible keypresses filling a
  * keyboard event data structure. The function returns true if a keyboard event
--- a/openrtx/include/core/input.h
+++ b/openrtx/include/core/input.h
@@ -1,5 +1,5 @@
 /***************************************************************************
- *   Copyright (C) 2020 - 2023 by Federico Amedeo Izzo IU2NUO,             *
+ *   Copyright (C) 2020 - 2024 by Federico Amedeo Izzo IU2NUO,             *
  *                                Niccolò Izzo IU2KIN,                     *
  *                                Silvano Seva IU2KWO,                     *
  *                                Grzegorz Kaczmarek SP6HFE                *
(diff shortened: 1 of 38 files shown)


openrtx/include/core/input.h
```diff
@@ -24,40 +25,84 @@
...
 
+/**
+ * Turn on or off a support of press event generation on
+ * new key pressed while 1+ keys is pressed at the time
+ * This is used when some keys stay pressed while
+ * others are changed
+ * 
+ * @param isPressEventAllowed: press events support during multi-press
+ */
+void input_allowPressEventOnMultilpeKeysPressed(bool isPressEventAllowed);
```
silseva: I'd rename this to `input_enableMultiKeyEvent` to reduce verbosity

openrtx/include/core/input.h
```diff
@@ -24,40 +25,84 @@
...
 
+/**
+ * @brief 
+ * 
+ * @param keys keys pressed
+ * @param msg keyboard event message to be filled
+ * @return true msg contain valid event message
+ * @return false msg contents are not valid
+ */
+bool processKnobMovementDetection(keyboard_t keys, kbd_msg_t* msg);
```
silseva: The `input_` prefix is missing. I also suggest to rename this to `input_handleKnobEvents` or something similar

openrtx/src/core/input.c
```diff
@@ -17,68 +18,169 @@
...
 #include <interfaces/delays.h>
 #include <inttypes.h>
 #include <stdbool.h>
-#include <input.h>
+#include <stddef.h>
 
-static long long  keyTs[KBD_NUM_KEYS];  // Timestamp of each keypress
-static uint32_t   longPressSent;        // Flags to manage long-press events
-static keyboard_t prevKeys = 0;         // Previous keyboard status
+bool pressEventAllowedForMultipleKeys;
```
silseva: The `static` keyword and the comment are missing. Also, please reduce the verbosity. I suggest, for example, `multiKeyEvent`

openrtx/src/core/input.c
```diff
@@ -17,68 +18,169 @@
...
+void input_allowPressEventOnMultilpeKeysPressed(bool isPressEventAllowed)
 {
-    msg->value     = 0;
-    bool kbd_event = false;
+    pressEventAllowedForMultipleKeys = isPressEventAllowed;
+}
+
+bool processKnobMovementDetection(keyboard_t keys, kbd_msg_t* msg)
+{
+    const bool isKeyboardEvent = true;      // Return value
```
silseva: I'd just return true/false instead of wrapping these into const variables

openrtx/src/core/input.c
```diff
@@ -17,68 +18,169 @@
...
+        msg->event = KEY_EVENT_PRESS;
+        return isKeyboardEvent;
+    }
+
+    return noKeyboardEvent;
+}
+
+bool input_scanKeyboard(long long timestamp, kbd_msg_t* msg)
+{
+    static keyboard_t previousKeys      = 0;     // Previously pressed keys
```
silseva: Please avoid defining static variables inside the function, if possible. These should be moved at the beginning of the file.

silseva (changes requested):

openrtx/src/core/input.c
sp6hfe: Intention here is to narrow the scope. input_scanKeyboard is the only user and it should be like that but I see your point in having all the globals in one place. Upcoming C++ class (separate PR) would handle scoping in a proper way.

openrtx/src/core/input.c
sp6hfe: Agree. This is my habit from C++ where for readability I use lots of static constexpr.

silseva:
@sp6hfe if you can please fix the various issues, I'll then proceed with merging the PR :)

openrtx/include/core/input.h
sp6hfe: fixed

openrtx/include/core/input.h
sp6hfe: fixed

openrtx/src/core/input.c
sp6hfe: fixed

openrtx/src/core/input.c
sp6hfe: fixed

openrtx/src/core/input.c
sp6hfe: fixed

sp6hfe:
Pls check fixes

silseva (approved):

PR #317, opened 2025-04-18
Add CS7000/CS700P/DM1701 to actions build

marcoSchr:
Add the new targets CS7000, CS7000P and DM1701 to the GitHub actions build.


silseva:
Just one note: the CS7000-PLUS is based on an ARM Cortex M7 CPU, its firmware needs to be built in a meson project configured with the cross_cm7.txt options

marcoSchr:
I have updated it, to have an additional step, where we compile the cortex m7 targets.

silseva:
Thank you!

PR #319, opened 2025-05-03
Create Dockerfile, devcontainer and vscode tasks for all targets

peterus:
What I did:

- add a Dockerfile to build a working docker container with all dependencies to build the project on
  - linux
  - arm
  - cm7
- add a devcontainer configuration to
  - automatically build and
  - open the docker container
- add a vscode tasks configuration for all [known targets](https://openrtx.org/#/compiling?id=compiling-for-a-radio) to run the build process in a faster way
- add a task to run the unit tests on linux


peterus:
please provide feedback! I am finished from my side with this PR.

silseva:
Everything looks good!
A couple of notes:
- you can use the pre-built miosix gcc instead of compiling it from sources, but probably you're doing so to cover all the architectures
- all the commits can be squashed into a single one :)

PR #324, opened 2025-06-29
Add library to generate QR Codes

marcoSchr:
This adds the [QRCode](https://github.com/ricmoo/QRCode) library.
With a small example to create a QRCode and render it to the screen, it takes 4512 Bytes of memory on the mduv3x0.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,6 +70,7 @@ target_include_directories(app
 
     lib/minmea/include
     lib/qdec/include
+    lib/QRCode/include
 )
 
 target_sources(app
@@ -137,4 +138,5 @@ target_sources(app
     subprojects/codec2/src/codebooknewamp1_energy.c
 
     lib/minmea/minmea.c
+    lib/QRCode/qrcode.c
 )
--- a/README.md
+++ b/README.md
@@ -78,6 +78,8 @@ This software is released under the GNU GPL v3.
 
 minmea is released under the DO WHAT THE FUCK YOU WANT TO PUBLIC LICENSE v2.
 
+[QRCode](https://github.com/ricmoo/QRCode) is released under the MIT License.
+
 Code for STM32F405 USB driver is released under the MCD-ST Liberty SW License Agreement V2.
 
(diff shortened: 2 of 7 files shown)


silseva:
Set as draft after our discussion, to be turned back to ready for merge after the function in the graphics layer is ready

marcoSchr:
This is not quite ready yet. There is still one todo in the code, to get a good value for the qr code version (The size of the qr code).
There is this function in the zxing library that solves this problem: https://github.com/zxing/zxing/blob/master/core/src/main/java/com/google/zxing/qrcode/encoder/Encoder.java#L210

It would be good, if we can implement something similar. Otherwise I would go with the version based on the size, where the qr code will be rendered. We have a certain number of pixels to work with. If we assume we use one pixel per data point of the qr code, then we can calculate the maximum version that fits in this space.
The formula would be the following:
Smallest dimension of our rectangular region minus 4 for the border.
Then subtract 17 and divide by 4 rounding down. This will be the maximum version of QR Code, that fits in this area.

More information about qr code versions: https://www.qrcode.com/en/about/version.html

silseva:
I'd go for the computation of the version based on the screen size. Given that the screen size is already known at compile time, all the computation can be done at compile time too.
I'll be curious to see if the compiler is smart enough to entirely pre-process and optimize away a `static inline` function computing the maximum QR code version, leaving only the final result in the code. With C++ this comes for sure if you use a `constexpr` function, but with C...

silseva:
@marcoSchr since you said that the code is not yet ready, I put back the PR to draft state. I'll let you change it to ready when everything is done! :)

marcoSchr:
@silseva The code is now ready for review.

I tested the `static inline` function with godbolt: https://godbolt.org/z/TPjKxPvsa
From what I saw, the function is still called, but the result is precomputed by the compiler (line 4 of the assembly).

silseva:
I played a little bit with godbolt and, with `-Os` (the optimization level we use), the compiler replaces the function call with a constant.

silseva (approved):

PR #328, opened 2025-07-11
DRAFT - example of call history implementation

agamos:
This version has my implementation of call history.  See openrtx/src/core/history.c for the underlying callsign store.  This is enabled/disabled in the Settings->M17 menu using the option Keep History.

It implements a notification mechanism which adds a H to the display when unread history is available.  It also flashes the LED, in case the backlight saver is switched on.  There is a Settings->Display setting for Notifications that will enable/disable the notifications.

The history can be found in the new History menu item, which will display a list of calls in time order along with their time stamp.  If you long click the green button it copies the callsign so as to have it set up as a direct call.

As an aside and to be ignored for now...
1)  this version of the code also has some optimisations to the RF loop when in M17 mode.  This, on my TYT MD-380UV HT, has increased the M17 standby time by 30% - although it is still being tested/debugged when I have time.
2)  it has re-implemented the screensaver logic so it only unlocks the screensaver with a key press and ignores any volume change.  I found it annoying that the volume control would switch the display on while I'm simply listening and don't want the super bright screen to light up the room.

Apologies, if this PR is not any standard format - I only really intended this to be for my personal use and don't have prior experience of swarm coding.


silseva:
I see a lot of valuable work in your commits, which I'd definitely like to merge in mainline.
If I saw correctly there is:
- favorites (but please tell me more about this)
- M17 last heard list
- a display pop-up on volume change
- dark theme
- RF cycling for power save
- screensaver logic fix

Do you mind splitting each contribution into separate pull requests? This will help me reviewing the code speeding up the merge process. Please make a commit for each logical change in the code and in the commit messages use a format like “make xyzzy do frotz” or “changed xyzzy to do frotz”. You can get an idea of the commit message format also by taking a look at the history of the main branch.
Sorry for being a bit picky on this, but having a uniform code style and commit message style helps lowering the entry barrier for everyone willing to contribute.

I'm really looking forward to merging your contributions!
Thank you,
Silvano IU2KWO

silseva:
This PR contains a lot of valuable features, do not close it as stale

agamos:
Thanks for keeping the thread alive.  I've been busy on other projects and had figured M17 was dead after Jonathan dropped support in the MMDVM firmware.  I'll aim to prioritise these splits in April.
Andy G3AGM

PR #332, opened 2025-08-26
Use `-std=gnu17` to allow building with gcc 15

marcoSchr:
gcc 15 uses C23 by default now.
The project does not yet compile with C23.
Thus we set the standard back to C17, as it was the default before gcc 15.


silseva:
Thanks!
I wasn't aware of this issue: at least on my machine GCC15 compiles happily

PR #335, opened 2025-09-02
Initial bringover from A36plus branch

Tunas1337:
This commit also includes some debug lines to see if/why/how stuff didn't draw.


openrtx/src/core/graphics.c
```diff
@@ -210,6 +210,18 @@ void gfx_fillScreen(color_t color)
     }
 }
 
+color_t gfx_getPixel(point_t pos)
+{
+    if (pos.x >= CONFIG_SCREEN_WIDTH || pos.y >= CONFIG_SCREEN_HEIGHT ||
```
silseva: Please isolate each logic condition within parenthesis. Is not _logically_ useful, but it improves readability

openrtx/src/core/graphics.c
```diff
@@ -210,6 +210,18 @@ void gfx_fillScreen(color_t color)
...
+        pos.x < 0 || pos.y < 0)
+        return (color_t){0}; // off the screen
+
+    return (color_t){framebuffer[pos.x + pos.y*CONFIG_SCREEN_WIDTH].r,
```
silseva: Generalize the function also for `CONFIG_PIX_FMT_BW`

openrtx/include/core/state.h
```diff
@@ -81,6 +92,9 @@ enum RtxStatus
     RTX_OFF = 0,
     RTX_RX,
     RTX_TX
+    #ifdef CONFIG_SPECTRUM
+    ,RTX_SPECTRUM
```
silseva: I'd call this state `RTX_RX_SWEEP` and make it independent from spectrum. This allows you to do a RX sweep also for other applications

openrtx/include/ui/EnglishStrings.h
```diff
@@ -32,6 +32,7 @@ const stringsTable_t englishStrings =
     .channels          = "Channels",
     .contacts          = "Contacts",
     .gps               = "GPS",
+    .spectrum          = "Spectrum",
```
silseva: Please move this to the bottom of the list, to match the position to the one in the struct definition.
Also, add the string also to the spanish string table, otherwise the firmware is going to crash as soon as we implement multi-language support

openrtx/src/core/state.c
```diff
@@ -80,6 +80,12 @@ void state_init()
     state.emergency     = false;
     state.txDisable     = false;
     state.step_index    = 4; // Default frequency step 12.5kHz
+    #ifdef CONFIG_SPECTRUM
+    state.spectrum_startFreq = 432100000; // Default start frequency for spectrum display
```
silseva: The initialization value should be dependent on the hardware configuration, otherwise on VHF radios it won't work. See for example https://github.com/OpenRTX/OpenRTX/blob/master/openrtx/src/core/cps.c#L46

openrtx/src/core/state.c
```diff
@@ -136,7 +142,15 @@ void state_task()
     state.volume = vol / 2;
 
     state.charge = battery_getCharge(state.v_bat);
+    // There is a bug where during spectrum operation, the RSSI reads collide.
```
silseva, line 136: Don't manage this here, do it in rtx.cpp to keep this code agnostic of what is going on

openrtx/src/rtx/rtx.cpp
```diff
@@ -24,7 +24,9 @@
 #include <rtx.h>
 #include <OpMode_FM.hpp>
 #include <OpMode_M17.hpp>
+#include <core/state.h>
 
+extern state_t state;
```
silseva: This is not needed, state.h already contains this statement.

openrtx/include/core/state.h
```diff
@@ -62,6 +62,17 @@ typedef struct
     bool       restore_eflash;
     bool       txDisable;
     uint8_t    step_index;
+    #ifdef CONFIG_SPECTRUM
+    bool       spectrum_shouldRefresh; // True if one scanning cycle is done.
+    uint32_t   spectrum_startFreq; // Start frequency for spectrum display, in 10Hz units
+    uint32_t   spectrum_peakFreq;  // Current peak frequency for spectrum display, in 10Hz units
```
silseva: I'd move all the return values from the rtx, that is peak values, samples, ... to a dedicated data structure like `struct rxSweep` defined inside rtx.h An instance of this struct struct can then be part of the `rtxStatus_t` struct.

Also, don't use magic numbers for the size of the array, use a macro

openrtx/src/rtx/rtx.cpp
```diff
@@ -126,6 +128,74 @@ void rtx_task()
...
+* the audio path is disabled. This allows to display the RSSI level of the
+* received signals across a frequency range in a waterfall-like display.
+* 
+* This block writes the received RSSI levels to the spectrum buffer.
+*/
+    #ifdef CONFIG_SPECTRUM
+    if(state.rtxStatus == RTX_SPECTRUM)
+    {
+        //state.spectrum_peakRssi = -160;
+        #define SPECTRUM_WF_LINES 64 // must match NUMBER_BARS in ui_menu.c!
```
silseva: See above, use a macro in rtx.h: this value is already conflicting with the buffer size set in state.h and, probably, with the one you're using in the UI.

openrtx/src/rtx/rtx.cpp
```diff
@@ -126,6 +128,74 @@ void rtx_task()
...
+        uint32_t spectrumStep = freq_steps[state.settings.spectrum_step];
+        // Get the current RSSI level
+
+        uint8_t peakIndex;
+        // Write the RSSI level to the spectrum buffer
+        for (int i = 0; i < SPECTRUM_WF_LINES; i++)
+        {
```
silseva: Please follow the new code style guide, that is  `for (int i = 0; i < SPECTRUM_WF_LINES; i++) {`

openrtx/src/rtx/rtx.cpp
```diff
@@ -126,6 +128,74 @@ void rtx_task()
...
+            // set peak value
+            if(rssi > state.spectrum_peakRssi)
+            {
+                state.spectrum_peakRssi = rssi;
+                state.spectrum_peakFreq = state.spectrum_startFreq + i * spectrumStep;
+                state.spectrum_peakIndex = i;
+            }
+            // stop scanning if the rssi is greater than the current squelch rssi,
+            // and listen to that frequency
+            if(radio_getRssi() > (-127 + (state.settings.sqlLevel * 66) / 15))
```
silseva: Do we really want this? By the way, the squelch is already managed in OpMode_FM and this is conflicting with that part

openrtx/src/rtx/rtx.cpp
```diff
@@ -185,25 +255,27 @@ void rtx_task()
      */
     if(rtxStatus.opStatus == RX)
     {
-
```
silseva, line 174: I don't see any functional change, please restore the proper indent

openrtx/src/ui/default/ui.c
```diff
@@ -1900,6 +1929,41 @@ void ui_updateFSM(bool *sync_rtx)
                     _ui_menuBack(MENU_TOP);
                 break;
 #endif
+#ifdef CONFIG_SPECTRUM
+            // Spectrum menu screen
+            case MENU_SPECTRUM:
+            // Go to the Spectrum Settings if we press ENTER
+                if(msg.keys & KEY_ENTER)
+                {
+                    display_init();
```
silseva: Here you are re-initializing the display driver, are you sure that is what you really want to do?

openrtx/src/ui/default/ui.c
```diff
@@ -1900,6 +1929,41 @@ void ui_updateFSM(bool *sync_rtx)
...
+                    state.rtxStatus = RTX_RX;
+                    state.ui_screen = SETTINGS_SPECTRUM;
+                }
+                if(msg.keys & KEY_ESC) {
+                    state.rtxStatus = RTX_RX;
+                    radio_enableRx();
```
silseva: Don't call the radio API outside the rtx thread, please.
You can trigger a radio reconfiguration by setting `sync_rtx = true`

openrtx/src/ui/default/ui.c
```diff
@@ -2155,6 +2224,54 @@ void ui_updateFSM(bool *sync_rtx)
...
+                            break;
+                    }
+                }
+                else if(msg.keys & KEY_RIGHT || (ui_state.edit_mode &&
+                        (msg.keys & KEY_UP || msg.keys & KNOB_RIGHT)))
+                {
+                    switch(ui_state.menu_selected)
+                    {
+                        case SP_MULTIPLIER:
+                            state.settings.spectrum_multiplier = (state.settings.spectrum_multiplier + 1) % 5;
```
silseva: Please use the `ARRAY_SIZE` macro instead of magic numbers

openrtx/src/ui/default/ui_menu.c
```diff
@@ -731,6 +761,168 @@ void _ui_drawMenuGPS()
...
 #endif
 
+#ifdef CONFIG_SPECTRUM
+void _ui_drawMenuSpectrum(ui_state_t* ui_state)
+{   
+    (void)ui_state;  // Suppress unused parameter warning
+    color_t bar_color = (color_t){255, 255, 255, 255};
+    uint32_t spectrumStep = freq_steps[state.settings.spectrum_step];
+    char freq_str[16];
+    #define NUMBER_BARS 64
```
silseva: Place those macros at the beginning of the source file, if you need them

openrtx/src/ui/default/ui_menu.c
```diff
@@ -731,6 +761,168 @@ void _ui_drawMenuGPS()
...
+            if(pos1.x < CONFIG_SCREEN_WIDTH && pos1.x < 128)
+                gfx_setPixel(pos1, bar_color);
+            if(pos2.x < CONFIG_SCREEN_WIDTH && pos2.x < 128)
+                gfx_setPixel(pos2, bar_color);
+        }
+        
+        state.spectrum_currentWFLine = ((state.spectrum_currentWFLine+1) % (waterfall_end_y - waterfall_start_y));
+    }
+    state.spectrum_currentPart = (state.spectrum_currentPart + 1) % 2;
+    ui_updateFSM(NULL);
```
silseva: Remove this function call, `ui_updateFSM` is already called by the main UI thread. Moreover, passing NULL as parameter may crash the firmware

openrtx/src/rtx/rtx.cpp
Tunas1337: Noted, will remove.

openrtx/src/core/state.c
Tunas1337: Hilariously, it was originally 146.550MHz and I changed it to this to test the CS7k 😅 

This default doesn't really apply in any real-world use anymore, as we reuse the current (valid, hopefully) VFO RX frequency as of the time the user enters the spectrum. However, it's still good to have a sane default. I'll work on this.

openrtx/src/ui/default/ui.c
Tunas1337: Artifact of A36plus code, when I needed to reinit the display to remove the scrolling register modifications and/or clear the GRAM, including the offscreen portions. Will remove.

openrtx/src/ui/default/ui_menu.c
Tunas1337: Some yes, some no. I'll remove the ones I don't need and move the rest.

openrtx/src/ui/default/ui.c
Tunas1337: This is a directly-used multiplier, there's no array associated with it. I do suppose I could #define the 5 as SPECTRUM_MULTIPLIER_MAX or something like that, so it's easily tweaked?

Tunas1337:
I accidentally undrafted the PR instead of re-requesting review. Oops 😂

openrtx/src/rtx/rtx.cpp
```diff
@@ -117,15 +127,63 @@ void rtx_task()
...
+    {
+        rtxStatus.rxSweep_data.sweepDone = false; // Clear the flag at the start of the sweep
+        uint32_t spectrumStep = freq_steps[state.settings.spectrum_step];
+        // Get the current RSSI level
+        // uint8_t peakIndex; // No longer needed
+        rtxStatus.rxSweep_data.peakRssi = -160; // Reset peak RSSI for each sweep
+        // Write the RSSI level to the spectrum buffer
+        uint32_t spanWidth = RX_SWEEP_POINTS * freq_steps[state.settings.spectrum_step];
+        uint32_t startFreq = state.rxSweep_start_freq;
+        for (int i = 0; i < RX_SWEEP_POINTS; i++) {
```
silseva, line 144: Does it make sense to have this scan part doing one step for each update of the rtx_task? In FM mode the update rate is 33Hz, so it will take around 2s to do a full scan

Tunas1337: I'm not sure I understand. As far as I can tell, it does a full sweep for every one execution of the rtx_task. Or did you mean that it _should_ do one step for each rtx_task update? I think 2s is too long for one full scan; currently, it updates 2-4 times a second.

silseva: I meant the second: do a single step for each task update. If 2s for a full scan is too much, then the code is fine as it is now :)

Tunas1337:
As soon as I resolve these conflicts, are we good to merge?

silseva:
Yes. Please rebase on top of master

silseva:
This contribution is planned to be integrated in OpenRTX release v0.5.0, merge will happen as soon as I have some time.

PR #336, opened 2025-09-02
Draft: CI builds using devcontainer

turnrye:
# What

- Update the devcontainer so that it uses miosix distributable rather than (slowly) building from source
- Update the CI job to use the devcontainer rather than setting up an ad-hoc build environment
- Update the artifacts that are being published to be more complete

# Why

- Trying to make the CICD and dev tools a bit simpler
- Wanting the CI job to be complete in building all platforms so that in the future it can be used as a source of nightlies

# Todo

- [ ] Confirm that the published files are the right ones, make naming consistent with those that are present in the existing releases
- [ ] Decide what to do with ttwrplus -- those builds havent been in the latest releases, do we need to modernize the ci for that or simply remove it


turnrye:
@silseva this is ready for review!

silseva:
Actually we have pre-built binaries of the miosix GCC at https://files.openrtx.org/toolchains/

turnrye:
Mmm thank you @cifred98 , this reminded me to test on a different architecture. The devcontainer doesnt work on a modern mac due to the architecture (it actually fails first on radio_tool). So, putting this in draft while I make tweaks to support arm64.

Edit: I've tried a few approaches to get either the full toolchain to work in arm64 or to use docker's emulator feature to build and run the devcontainer in amd4. Unfortunately I could get neither to reliably work, encountering various issues along the way. For now, I'm going to suggest that we move forward with the expectation that the devcontainer only works on amd64 hosts, and to flag that PRs are welcome to build support for other architectures. I think this is acceptable since the devcontainer for local env is a nice-to-have rather than a must-have.

turnrye:
Hey @silseva I just synced with master and resolved the conflict, but I see you changed the merge target branch. I'll leave it sitting from here so that you can land into that branch and tinker if you want to. You also have permissions to push to my fork's branch as well if you'd rather play within the PR itself -- I'll stop rebasing.

PR #341, opened 2025-09-08
Add contributing guide

turnrye:
# What

Add a contributing guide, that way new contributors have a shared understanding of expectations from the beginning.

# Why

To help contributors be more successful

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,101 @@
+# Contributing
+
+Thanks for your interest in contributing to OpenRTX! Due to the complexity of this project, strongly consider opening an issue to discuss your changes first, that way we can ensure that coordination with other ongoing changes and the project's roadmap. Help is always welcome, and there's plenty of room!
+
+## Where to start
+
+- Improving the documentation (both in this repository and in the adjacent website repository)
+- Triaging issues, especially if you have the means to reproduce them and add more details to them
+- Adding test coverage
+
+In general, create a fork, get your development environment setup (we recommend VS Code with the included devcontainer), make your first build, and start exploring!
+
+## Style Conventions
+
+Use the `scripts/clang_format.sh` script to ensure your contributions follow the intended style conventions.
+
+### Code Organization
+
+The main directories for the project are described below:
+
+```text
+openrtx/     # The core program, including its UI, common utilities, and features
+platform/    # Hardware specific implementations, including things like drivers for specific components, details of specific hardware targets, and low-level implementations of hardware specific needs
+tests/       # Tests, both unit tests and platform tests intended to be run on device (linux or otherwise)
+scripts/     # Build and utility scripts
+```
+
+More details about the individual components can be found on the [developer pages](https://openrtx.org/#/software).
+
+### Making Changes
+
+OpenRTX uses Git for source code management. Each commit should represent a distinct change, and as such often PRs will contain multiple commits. It isn't OK, though, for a commit to contain broken or incomplete changes. Keep your branch's revision history tidy using strategies like squashing.
+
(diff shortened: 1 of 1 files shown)


mdiepart:
Hi !
I think this is a good idea to have such a guide in the repository. The document from Contributor Covenant seems very good and well written.

Two suggestions:
1. Should we mention something about AI contributions?
2. I think we may go into a bit more details regarding the code organization

turnrye:
> Hi ! I think this is a good idea to have such a guide in the repository. The document from Contributor Covenant seems very good and well written.
> 
> Two suggestions:
> 
>     1. Should we mention something about AI contributions?
I think the maintainers should decide the project's stance on AI and separately update the contributing guide. I'd be happy to contribute to a discussion on this during the process.
 
>     2. I think we may go into a bit more details regarding the code organization
That sounds good, I'll add more detail.

lemielek:
About AI - something like that? https://www.phoronix.com/news/Cloud-Hypervisor-48

silseva:
Gentoo adopted this AI policy: https://wiki.gentoo.org/wiki/Project:Council/AI_policy
I think we should adopt a similar policy if not the same

mdiepart:
Both policies go in the same way as to reject contributions containing AI. Next question is: is it a separate issue from this one?

turnrye:
I think I'd like to advocate for a more permissive stance, but I don't want to cause delay to getting the basics of a contributing guide merged. 

So, I'd suggest for this to have a separate, dedicated issue where the maintainers either state a position or welcome feedback on AI usage. Then we can take the appropriate follow up actions.

silseva:
Ok, let's handle the AI part in a separate issue.
The part about coding style conventions is small, but I think is better to have a dedicated page on the website with all the rules for code style and commit message formatting.

What about citing/requiring/indicating the [Developer Certificate of Origin](https://developercertificate.org/) and requiring sign-off of commits? To me, is something we need. I need to better understand the exact meaning of point (b), though.

CONTRIBUTING.md
```diff
@@ -0,0 +1,101 @@
...
+## Style Conventions
+
+Use the `scripts/clang_format.sh` script to ensure your contributions follow the intended style conventions.
+
+### Code Organization
+
+The main directories for the project are described below:
+
+```text
+openrtx/     # The core program, including its UI, common utilities, and features
```
mdiepart: lib/    # External libraries such as the RTOS kernel or GNSS libraries

mdiepart: Regarding my previous comment about adding details to the code organization, I didn't realize how much work @silseva put into cleaning-up the code organization and it seems to me much more understandable just going through the project tree. I think this is good as-is.

mdiepart:
> Ok, let's handle the AI part in a separate issue. The part about coding style conventions is small, but I think is better to have a dedicated page on the website with all the rules for code style and commit message formatting.
> 
> What about citing/requiring/indicating the [Developer Certificate of Origin](https://developercertificate.org/) and requiring sign-off of commits? To me, is something we need. I need to better understand the exact meaning of point (b), though.

DCO seems like a good idea and wouldn't require much to implement, but could also be sorted-out in another issue so that we can get this one closed. 

For point (b), I think it means that you got the code from a project with a compatible license. Obviously the 4 points need to be OR'ed and not AND'ed ^^'

turnrye:
Hi @mdiepart @silseva , tagging you as my reviewers here. I've made some pretty hefty additions to this change, and I hope that they're in line with what you're expecting. Please give this a re-review and leave your feedback. If possible, I'd love to pair program in real time to help land this well before the end of the month. Thanks!

mdiepart:
Thank you very much for your contribution and the additional changes :)

mdiepart (approved):

CONTRIBUTING.md
silseva: Thank you :smile:

PR #314, opened 2025-04-01
Added M17 packet and SMS messaging.

kd0oss:
Added M17 Metatext.
Fixed Module17 HMI button lockup.

PR #347, opened 2025-09-13
Show M17 Meta Text

turnrye:
# What

When a received M17 transmission has metatext available, show it in a scrolling field.

Here's a demo gif, showing both scrolling and non scrolling text:

optimized_output

It's not immediately apparent to me why the meta is "flashing" when there's a multi frame meta text. Help with debugging is appreciated.

# How

This PR extracts portions of the M17 protocol changes and the UX feature from https://github.com/OpenRTX/OpenRTX/pull/314

- Update the M17 protocol to handle incoming M17 meta text, and a few stub-outs towards outgoing and gnss meta type
- Add a scrollingText UX control that _should_ be reusable in other contexts
- Add an "abc" icon to show beside metatext, the icon came from https://pictogrammers.com/library/mdi/icon/alphabetical-variant/ like the other existing icons and is already covered according to the license in the directory

# Why

This is part 1 of many to land the features that Rick et al built in https://github.com/OpenRTX/OpenRTX/pull/314. A separate PR will handle setting and transmitting meta text as well as meta GNSS location. Separate from that will be packet mode. Finally, separate from _that_ will be SMS send and receive. The goal throughout these contributions is to refactor UI code in place to try to create more reusable/DRY modules.


turnrye:
@silseva , this PR changes some of the big preexisting files. I'd like to format these files, but that will significantly inflate the diff. One thought I had is, I could reformat the files in a separate commit at the very beginning of this. That way the subsequent diff would still be small and manageable. What do you think about that approach?

turnrye:
I got off track here. Per the discussion [here](https://discord.com/channels/771492414120656907/771492414702747710/1416927264741785742), I'll be closing this PR and reopening one that is true to the original PR.

PR #348, opened 2025-09-14
build docker image and push to github package

peterus:


turnrye:
Hey @peterus , did you try using the official devcontainer's CI for this? If possible, I think we should take that route; that way we don't have as much to maintain ourselves in terms of CI. https://github.com/devcontainers/ci?tab=readme-ov-file#quick-start

This is specifically what I'm wondering about: can we just add this to the existing  with the proper image/cache values and "it works"?

```yaml
- name: Pre-build dev container image
  uses: devcontainers/ci@v0.3
  with:
    imageName: ghcr.io/example/example-devcontainer
    cacheFrom: ghcr.io/example/example-devcontainer
    push: always
```

peterus:
@turnrye I just checked the devcontainer part. It looks like it is possible to compile the image for different platforms, the downside is just that it will not be build in parallel. this means we will increase the build time with every architecture which we want to build more. in the current implementation everything will be build parallel.

peterus:
@turnrye Just for completeness I implemented your approach: https://github.com/OpenRTX/OpenRTX/pull/350
Currently the Action run is still in progress: https://github.com/peterus/OpenRTX/actions/runs/17738377386/job/50405870404

The action file is much more simpler, lets wait for the timing.

PR #261, opened 2024-04-05
Add FM Settings Menu Entry & improve Macro Menu Formatting V2

imostlylurk:
This is to create an FM settings entry in the menu list, creating a basis for consistent configuration across RF modes. This additionally solves a spacing issue in the macro menu regarding CTCSS tone configuration. Now with approximately 2000 less unnecessary style changes.

PR #353, opened 2025-09-18
FM Menu

turnrye:
This PR is my attempt at helping to land https://github.com/OpenRTX/OpenRTX/pull/261. 

From the original changes, I've:

- Rebased to get latest master and addressed conflicts
- Rewritten the commit history to have feature-oriented commits (but kept the author for credit)
- Added to the FM Settings menu the ability to manage encode/decode setting
- Refactored the setting and display of the CTCSS enabled option to have code sharing

## Verification
There are no automated tests covering this feature, so you must manually test. This feature does not apply to module17, and the UI code is shared across all targets, so a single platform is sufficient to test with.

1. Open meta menu pressing M
2. While holding M, Repeatedly press 1 to toggle through the encode/decode ctcss options
3. Periodically unpress to verify that the main VFO screen reflects the proper CTCSS option as specified in the meta menu
4. Ensure that when the meta menu selects "N" for "None", the entire CTCSS entry is hidden from the VFO screen
5. Press enter to open the menu and navigate to Settings -> FM Settings
6. Ensure that the options shown in the menu as selected are the same as in the meta menu and in the VFO screen
7. Change the CTCSS tone frequency and enablement setting, again confirming that the changes are reflected in the VFO screen

optimized_output_10

### DIFF
--- a/openrtx/include/ui/EnglishStrings.h
+++ b/openrtx/include/ui/EnglishStrings.h
@@ -98,6 +98,11 @@ const stringsTable_t englishStrings =
     .noGps             = "No GPS",
     .batteryIcon       = "Battery Icon",
     .fmSettings        = "FM Settings",
-    .CTCSSTone         = "CTCSS Frequency"
+    .CTCSSTone         = "CTCSS Frequency",
+    .CTCSSEnabled      = "CTCSS En.",
+    .CTCSSEnabledEncode = "Encode",
+    .CTCSSEnabledDecode = "Decode",
+    .CTCSSEnabledBoth  = "Both",
+    .CTCSSEnabledNone  = "None"
 };
 #endif  // ENGLISHSTRINGS_H
--- a/openrtx/src/ui/default/ui.c
+++ b/openrtx/src/ui/default/ui.c
@@ -119,7 +119,7 @@ extern void _ui_drawSettingsReset2Defaults(ui_state_t* ui_state);
 extern void _ui_drawSettingsRadio(ui_state_t* ui_state);
 extern bool _ui_drawMacroMenu(ui_state_t* ui_state);
 extern void _ui_reset_menu_anouncement_tracking();
-
+// TODO: get these from ui strings / currentLanguage
 const char *menu_items[] =
 {
     "Banks",
@@ -188,7 +188,8 @@ const char * settings_m17_items[] =
 
 const char* settings_fm_items[] =
 {
-    "CTCSS Tone"
+    "CTCSS Tone",
+    "CTCSS On"
 };
 
 const char * settings_accessibility_items[] =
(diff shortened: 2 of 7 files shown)


openrtx/src/ui/default/ui.c
```diff
@@ -890,6 +891,26 @@ static bool _ui_exitStandby(long long now)
     return true;
 }
 
+// TODO: find a better home for this function
```
turnrye, line 894: Flagging this as existing tech debt. Yes I added these methods new, but we need a better place for these feature-oriented UI chunks. I don't think this PR is the place to introduce that quite yet though. So, TODO it is.

openrtx/src/ui/default/ui.c
```diff
@@ -119,7 +119,7 @@ extern void _ui_drawSettingsReset2Defaults(ui_state_t* ui_state);
 extern void _ui_drawSettingsRadio(ui_state_t* ui_state);
 extern bool _ui_drawMacroMenu(ui_state_t* ui_state);
 extern void _ui_reset_menu_anouncement_tracking();
-
+// TODO: get these from ui strings / currentLanguage
 const char *menu_items[] =
```
turnrye, line 123: Flagging this as existing tech debt. We need to refactor string handling here and in a few other places in order to support internationalization.

openrtx/include/ui/EnglishStrings.h
```diff
@@ -97,5 +97,16 @@ const stringsTable_t englishStrings =
...
     .noGps             = "No GPS",
     .batteryIcon       = "Battery Icon",
+    .fmSettings        = "FM Settings",
+    .CTCSSTone         = "CTCSS Frequency",
+    .CTCSSEnabled      = "CTCSS En.",
+    .CTCSSEnabledEncode = "Encode",
+    .CTCSSEnabledDecode = "Decode",
+    .CTCSSEnabledBoth  = "Both",
+    .CTCSSEnabledNone  = "None",
+    .CTCSSEnabledEncodeShort = "E",
```
silseva: There is no real reason to have the short symbols for the macro menu in the string table, actually. Its purpose is to have a direct correspondence between strings and voice prompts, and CTCSS settings in the macro menu are in any case announced using the full text

turnrye: I think its unsafe to assume that the first char of the full word is what should be displayed, though I haven't looked into all of the planned languages to really know that. I don't think we want to force every language to use the English abbreviations. So, If remove this from the string table, how would I update the feature to work?

silseva: So far we have only English and Spanish, but in general the idea is to have many more other languages.
But for the macro menu itself I believe we can just leave the current abbreviations: they are not localized, but in the worst case you can resort to the user guide.

turnrye: I pushed a fixup with the changes, does this address your concerns? Even without strings, I think centralizing this representation to functions is an improvement.

openrtx/src/ui/default/ui_main.c
```diff
@@ -78,10 +78,34 @@ void _ui_drawBankChannel()
               b, last_state.channel_index + 1, last_state.channel.name);
 }
 
+// TODO: find a better home for this function
```
silseva: This function can be simplified adopting this approach:

```
char* _ui_getToneEnabledString(bool tone_tx_enable, bool tone_rx_enable) {
    char *strings[] = {
        currentLanguage->CTCSSEnabledNone,
        currentLanguage->CTCSSEnabledEncode,
        currentLanguage->CTCSSEnabledDecode,
        currentLanguage->CTCSSEnabledBoth,
    };

    uint8_t index = (tone_rx_enable << 1) | tone_tx_enable;
    return strings[index];
}
```

The same for the other one

turnrye: Thanks -- I've adopted this pattern and flattened these into a single function now since the signature is the same. At the same time, I took your advice from our offline convo and switched the abbreviations to be in line with the user expectations ("T", "R", "N", "B"). "B" is a holdover till we redesign the macro menu because only one char fits. At the same time I realized that the empty string feature I had put in the char function wasn't needed, so I removed that.

silseva:
Thank you!

PR #358, opened 2025-09-20
ui: power: change power level to more granular value

romanatrn:
Add option for 1,2,3 and 5w power level.

### DIFF
--- a/openrtx/src/ui/default/ui.c
+++ b/openrtx/src/ui/default/ui.c
@@ -990,6 +990,10 @@ static void _ui_fsm_menuMacro(kbd_msg_t msg, bool *sync_rtx)
             break;
         case 6:
             if (state.channel.power == 1000)
+                state.channel.power = 2000;
+            else if (state.channel.power == 2000)
+                state.channel.power = 3000;
+            else if (state.channel.power == 3000)
                 state.channel.power = 5000;
             else
                 state.channel.power = 1000;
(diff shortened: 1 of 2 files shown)


openrtx/src/ui/default/ui.c
```diff
@@ -990,6 +990,10 @@ static void _ui_fsm_menuMacro(kbd_msg_t msg, bool *sync_rtx)
             break;
         case 6:
             if (state.channel.power == 1000)
+                state.channel.power = 2000;
```
silseva, line 993: What about having an array of values and cycling through it modulo its size? I think is better than having a cascade of `if else` branches

romanatrn: will work on it :)

silseva:
Please set it back to ready once done

juliadin:
Thanks for working on this. This would be really really nice to have for hotspot work as well. 1W (which on my RT3s comes out at more like 2W) is way too hot for a hotspot right next to me. :)

juliadin:
Any chance this can still happen and be made ready for merge?

silseva:
If someone can take care of it, for sure!
"Stale" does not mean that a gravestone fell on it forever :)

juliadin:
If everyone is ok with it I can take the spirit of @romanatrn s work and see if I can work my request in and credit them.

silseva:
Yes, sure

PR #360, opened 2025-09-22
M17 GNSS Meta

turnrye:
# What

This change adds support for sending M17 GNSS meta.

# How

This change is based on the work originally done in https://github.com/OpenRTX/OpenRTX/pull/314, current as of https://github.com/OpenRTX/OpenRTX/pull/314/commits/e8c3ea54b93fe894b86d69c29448fdcdb236bb7d. The changes there had whitespace changes removed, then were squash merged onto a new branch. From there, changes that were unrelated to the M17 GNSS meta feature were removed. This yielded the changes in this PR.

The work in this PR was done by @kd0oss Rick. I am helping pick and land the changes.

I've left this in an un-squashed state intentionally so that both the reviewers and Rick can "follow along" with how this branch was created. This PR is trying to avoid rebasing since there are multiple contributors. It should be squashed _after review_ but before merging. This strategy will be followed again for picking additional changes from this fork into the mainline repo.

### DIFF
--- a/openrtx/include/rtx/OpMode_M17.hpp
+++ b/openrtx/include/rtx/OpMode_M17.hpp
@@ -135,6 +135,7 @@ private:
     bool compareCallsigns(const std::string& localCs, const std::string& incomingCs);
 
 
+    static constexpr int16_t MINIMUM_GPS_TIME_TICKS = 5 * 30; // Wait at least 5 seconds between GPS transmissions
     bool startRx;                      ///< Flag for RX management.
     bool startTx;                      ///< Flag for TX management.
     bool locked;                       ///< Demodulator locked on data stream.
@@ -150,9 +151,7 @@ private:
     M17::M17FrameDecoder decoder;      ///< M17 frame decoder
     M17::M17FrameEncoder encoder;      ///< M17 frame encoder
     M17::M17LinkSetupFrame lsf;        ///< M17 link setup frame
-
-
-    static constexpr int16_t MINIMUM_GPS_TIME_TICKS = 5 * 30; // Wait at least 5 seconds between GPS transmissions
+    int16_t gpsTimer;                  ///< Timer used to manage GPS data transmission intervals.
 };
 
 #endif /* OPMODE_M17_H */
--- a/openrtx/src/protocols/M17/M17LinkSetupFrame.cpp
+++ b/openrtx/src/protocols/M17/M17LinkSetupFrame.cpp
@@ -161,38 +161,43 @@ uint16_t M17LinkSetupFrame::crc16(const void *data, const size_t len) const
 }
 
 void M17LinkSetupFrame::setGnssData(const gps_t *position, const M17GNSSStationType stationType) {
-    gnssData_t gnssData;
-    memset(&gnssData, 0, sizeof(gnssData));
+    if(position->fix_type < FIX_TYPE_2D) {
+        return;
+    }
 
-    gnssData.data_src = M17_GNSS_SOURCE_OPENRTX;
-    gnssData.station_type = stationType;
+    streamType_t streamType = getType();
+    streamType.fields.encType    = M17_ENCRYPTION_NONE;
+    streamType.fields.encSubType = M17_META_GNSS;
+    setType(streamType);
 
-    gnssData.coords_valid = 1;
-    gnssData.alt_valid = 1;
-    gnssData.velocity_valid = 1;
-    gnssData.radius_valid = 0;
+    data.meta.gnss_data.data_src = M17_GNSS_SOURCE_OPENRTX;
+    data.meta.gnss_data.station_type = stationType;
 
-    gnssData.radius = position->hdop;
+    data.meta.gnss_data.coords_valid = 1;
+    data.meta.gnss_data.alt_valid = 1;
+    data.meta.gnss_data.velocity_valid = 1;
+    data.meta.gnss_data.radius_valid = 0;
 
-    gnssData.bearing_1 = (position->tmg_true >> 8) & 0x01; // MSB (1 bit)
-    gnssData.bearing_2 = position->tmg_true & 0xFF; // Lower 8 bits
+    data.meta.gnss_data.radius = position->hdop;
+
+    data.meta.gnss_data.bearing_1 = (position->tmg_true >> 8) & 0x01; // MSB (1 bit)
+    data.meta.gnss_data.bearing_2 = position->tmg_true & 0xFF; // Lower 8 bits
 
     // Encode the coordinates in Q1.24 format, flip the byte order for M17's big endian standard
     int32_t lat_encoded, lon_encoded;
     rtx_to_q(&lat_encoded, &lon_encoded, position->latitude, position->longitude);
     for(uint8_t i = 0; i < 3; i++)
     {
-        gnssData.latitude_bytes[i] = *((uint8_t*)&lat_encoded + 2 - i);
-        gnssData.longitude_bytes[i] = *((uint8_t*)&lon_encoded + 2 - i);
+        data.meta.gnss_data.latitude_bytes[i] = *((uint8_t*)&lat_encoded + 2 - i);
+        data.meta.gnss_data.longitude_bytes[i] = *((uint8_t*)&lon_encoded + 2 - i);
     }
 
     uint16_t speed = (uint16_t)position->speed * 2;
-    gnssData.speed_1 = speed >> 4; // MBS
-    gnssData.speed_2 = speed & 0x0F; // Lower 4 bits
+    data.meta.gnss_data.speed_1 = speed >> 4; // MBS
+    data.meta.gnss_data.speed_2 = speed & 0x0F; // Lower 4 bits
 
     // Structure numeric fields that need offset and steps
     uint16_t alt = (uint16_t)1000 + position->altitude * 2;
-    gnssData.altitude = __builtin_bswap16(alt);
+    data.meta.gnss_data.altitude = __builtin_bswap16(alt);
 
-    data.meta.gnss_data = gnssData;
 }
--- a/openrtx/src/rtx/OpMode_M17.cpp
+++ b/openrtx/src/rtx/OpMode_M17.cpp
@@ -313,14 +313,12 @@ void OpMode_M17::rxState(rtxStatus_t *const status)
 
 void OpMode_M17::txState(rtxStatus_t *const status)
 {
-    static streamType_t type;
     frame_t m17Frame;
-    static int16_t gpsTimer;
 
     if(startTx)
     {
         startTx = false;
-        gpsTimer = -1;
+        gpsTimer = 0;
 
         std::string src(status->source_address);
         std::string dst(status->destination_address);
@@ -328,7 +326,14 @@ void OpMode_M17::txState(rtxStatus_t *const status)
         lsf.clear();
         lsf.setSource(src);
         if(!dst.empty()) lsf.setDestination(dst);
+        
+        if(state.settings.gps_enabled)
+        {
+            lsf.setGnssData(&state.gps_data, M17_GNSS_STATION_HANDHELD);
+            gpsTransmitting = true;
+        }
 
+        streamType_t type = lsf.getType();
         type.fields.dataMode = M17_DATAMODE_STREAM;     // Stream
         type.fields.dataType = M17_DATATYPE_VOICE;      // Voice data
         type.fields.CAN      = status->can;             // Channel access number
@@ -349,16 +354,11 @@ void OpMode_M17::txState(rtxStatus_t *const status)
         modulator.sendFrame(m17Frame);
     }
 
-    gps_t gps_data;
-    gps_data = state.gps_data;
-    if(encoder.canUpdateLsf() && state.settings.gps_enabled && (gpsTimer == -1 || gpsTimer >= MINIMUM_GPS_TIME_TICKS) && gps_data.fix_type > FIX_TYPE_NOT_AVAIL)
+    if(encoder.canUpdateLsf() && state.settings.gps_enabled && gpsTimer >= MINIMUM_GPS_TIME_TICKS)
     {
         gpsTimer = 0;
         gpsTransmitting = true;
-        lsf.setGnssData(&gps_data, M17_GNSS_STATION_HANDHELD);
-        type.fields.encType    = M17_ENCRYPTION_NONE;
-        type.fields.encSubType = M17_META_GNSS;
-        lsf.setType(type);
+        lsf.setGnssData(&state.gps_data, M17_GNSS_STATION_HANDHELD);
         lsf.updateCrc();
         encoder.encodeLsf(lsf, m17Frame);
     }
@@ -393,7 +393,7 @@ void OpMode_M17::txState(rtxStatus_t *const status)
         encoder.encodeEotFrame(m17Frame);
         modulator.sendFrame(m17Frame);
         modulator.stop();
-        gpsTimer = -1;
+        gpsTimer = 0;
         gpsTransmitting = false;
     }
 }


openrtx/src/rtx/OpMode_M17.cpp
```diff
@@ -172,7 +179,9 @@ void OpMode_M17::offState(rtxStatus_t *const status)
 
 void OpMode_M17::rxState(rtxStatus_t *const status)
 {
```
turnrye, line 187: Feedback on feature picking for myself: I dont think any of the changes within the rxState are related to this MR, given that the gnss feature has no RX behavior. This is probably prime to be reverted from this MR and reintroduced as part of the meta text feature PR.

kd0oss: That seems reasonable.

kd0oss:
Looks good to me. There is one small change that could made. In OpMode_M17.hpp the "#if" lines regarding Repeater/Hotspot could be removed as they are irrelevant for now.

turnrye:
> Looks good to me. There is one small change that could made. In OpMode_M17.hpp the "#if" lines regarding Repeater/Hotspot could be removed as they are irrelevant for now.

Thanks, Rick! I'll remove that plus the other rxState change, and then we can float this up for review with the leads.

kd0oss:
Changes look fine to me.

openrtx/include/protocols/M17/M17Demodulator.hpp
```diff
@@ -165,7 +165,9 @@ class M17Demodulator
     filter_state_t                 dcrState;        ///< State of the DC removal filter
 
     Correlator   < M17_SYNCWORD_SYMBOLS, SAMPLES_PER_SYMBOL > correlator;
+    Synchronizer < M17_SYNCWORD_SYMBOLS, SAMPLES_PER_SYMBOL > lsfSync   {{ +3, +3, +3, +3, -3, -3, +3, -3 }};
```
silseva: This and the line below are adding LSF and EOT syncword detection to the demodulator, they should become a separate PR

openrtx/include/rtx/OpMode_M17.hpp
```diff
@@ -16,6 +16,8 @@
  *                                                                         *
  *   You should have received a copy of the GNU General Public License     *
  *   along with this program; if not, see <http://www.gnu.org/licenses/>   *
+ *                                                                         *
```
silseva: Authors should be added to the list above

openrtx/include/rtx/OpMode_M17.hpp
```diff
@@ -26,8 +28,15 @@
 #include <M17/M17Demodulator.hpp>
 #include <M17/M17Modulator.hpp>
 #include <audio_path.h>
+#include <vector>
 #include "OpMode.hpp"
 
+#define M17_GNSS_SOURCE_M17CLIENT        ((uint8_t)0)
```
silseva: Please move these to M17LinkSetupFrame and change them to `enum`s or `enum class`

openrtx/include/rtx/OpMode_M17.hpp
```diff
@@ -134,20 +143,34 @@ class OpMode_M17 : public OpMode
...
-    pathId rxAudioPath;                ///< Audio path ID for RX
-    pathId txAudioPath;                ///< Audio path ID for TX
-    M17::M17Modulator    modulator;    ///< M17 modulator.
-    M17::M17Demodulator  demodulator;  ///< M17 demodulator.
-    M17::M17FrameDecoder decoder;      ///< M17 frame decoder
-    M17::M17FrameEncoder encoder;      ///< M17 frame encoder
+    /**
+     * Convert OpenRTX latitude/longitude format into M17 format and vice versa.
+     */
+    void rtx_to_q(int32_t* qlat, int32_t* qlon, int32_t lat, int32_t lon);
```
silseva: These are actually general-purpose conversion functions: please move them to gps.h or utils.h
Also, there is no "OpenRTX latitude/longitude" and "M17" formats. OpenRTX is using fixed point, while M17 is using Q1.24

openrtx/src/protocols/M17/M17Demodulator.cpp
```diff
@@ -265,17 +265,24 @@ bool M17Demodulator::update(const bool invertPhase)
                 case DemodState::UNLOCKED:
                 {
                     int32_t syncThresh = static_cast< int32_t >(corrThreshold * 33.0f);
-                    int8_t  syncStatus = streamSync.update(correlator, syncThresh, -syncThresh);
```
silseva, line 268: These changes are a separate PR, as above

openrtx/src/protocols/M17/M17Demodulator.cpp
```diff
@@ -300,7 +307,34 @@ bool M17Demodulator::update(const bool invertPhase)
                     }
                     else
                     {
-                        demodState = DemodState::UNLOCKED;
```
silseva, line 303: ditto

openrtx/src/protocols/M17/M17LinkSetupFrame.cpp
```diff
@@ -76,6 +76,12 @@ void M17LinkSetupFrame::setType(streamType_t type)
     data.type  = type;
 }
 
+// Added by KD0OSS (2025)
```
silseva: This is adding meta text, but this PR deals with GNSS data. I'd remove these changes and add them to the other incoming PRs

openrtx/src/rtx/OpMode_M17.cpp
```diff
@@ -16,21 +16,27 @@
...
+#include <vector>
 #include <errno.h>
+#include <stdlib.h>
+#include <math.h>
+#include <gps.h>
 #include <rtx.h>
 
 #ifdef PLATFORM_MOD17
 #include <calibInfo_Mod17.h>
-#include <interfaces/platform.h>
```
silseva: Restore this line

openrtx/src/rtx/OpMode_M17.cpp
```diff
@@ -335,6 +347,93 @@ void OpMode_M17::txState(rtxStatus_t *const status)
...
+    	lsfFragCount++;
+
+
+    if(gpsStarted)
+    {
+        gpsStarted = false;
+        gpsTimer = 0;
+
+        gps_t gps_data;
+        pthread_mutex_lock(&state_mutex);
```
silseva: Locking the state mutex in the rtx threads may introduce delays disrupting the correct timings for M17 modulation. GPS data is being fetched for read access and is actually safe to retrieve it without a critical section

openrtx/src/rtx/OpMode_M17.cpp
```diff
@@ -335,6 +347,93 @@ void OpMode_M17::txState(rtxStatus_t *const status)
...
+        gps_data = state.gps_data;
+        pthread_mutex_unlock(&state_mutex);
+
+        if(gps_data.fix_type > 0) //Valid GPS fix
+        {
+            platform_ledOn(YELLOW); // Blink LED yellow when sending GNSS
```
silseva: LED management should be done only in the dedicated FSM

openrtx/src/rtx/OpMode_M17.cpp
```diff
@@ -335,6 +347,93 @@ void OpMode_M17::txState(rtxStatus_t *const status)
...
+        	uint8_t gnss[14] = {0};
+
+        	gnss[0] = (M17_GNSS_SOURCE_OPENRTX<<4) | M17_GNSS_STATION_HANDHELD; //OpenRTX source, portable station
```
silseva: This code should be moved into a dedicated member function of the M17LinkSetupFrame class. For example `void setGnssData(const struct gps_t *position)`
Additionally, I think is better to handle data using a struct with bitfields rather than a raw array of bytes with mask-and-shift operations

openrtx/include/protocols/M17/M17Demodulator.hpp
turnrye: Thanks, I can move this out! This isn't needed till the PR for packet mode for SMS.

openrtx/src/protocols/M17/M17Demodulator.cpp
turnrye: Thanks, I can move this out! This isn't needed till the PR for packet mode for SMS.

openrtx/src/protocols/M17/M17Demodulator.cpp
turnrye: Thanks, I can move this out! This isn't needed till the PR for packet mode for SMS.

openrtx/src/protocols/M17/M17LinkSetupFrame.cpp
turnrye: Thanks, I can move this out!

openrtx/src/protocols/M17/M17LinkSetupFrame.cpp
```diff
@@ -158,3 +159,55 @@ uint16_t M17LinkSetupFrame::crc16(const void *data, const size_t len) const
 
     return crc;
 }
+
+void M17LinkSetupFrame::setGnssData(const struct gps_t *position) {
+    // TODO: shift this to a struct with bitwise fields
+    uint8_t gnss[14] = {0};
```
lwvmobile: Does the compiler for this allow an array to be initialized like this and guarantee a zero fill on all elements? Might be wise to use a memset on this, because for gnss[1] you aren't assigning an initial value, you are doing some bit operations with &=, which I don't quite understand, if the goal is to set gnss[1] to 0, why not just straight assign it an initial value of zero, then fill in the validity bits and radius instead of the weird &= ~0xF0 stuff you got going on.

silseva: Yes, is fine. When you do this on a type which is not a basic one, the compiler is replacing that line with a `memset`.

openrtx/src/rtx/OpMode_M17.cpp
```diff
@@ -335,6 +345,56 @@ void OpMode_M17::txState(rtxStatus_t *const status)
...
 
+    if(lsfFragCount == 6)
+    {
+	   lsfFragCount = 0;
+       gpsStarted = false;
+
+       // reset metadata if no text message
+       if(last_text_blk == 0xff)
+       {
+           uint8_t buf[14];
```
turnrye: 14 magic number, sizeof(type.raw_data)?

openrtx/include/protocols/M17/M17Datatypes.hpp
```diff
@@ -28,6 +28,13 @@
 #error This header is C++ only!
 #endif
 
+// Constants for the GNSS Meta field
```
silseva: Is better to have these as `enum`s instead of `#define`s, also for consistency with the rest

openrtx/include/protocols/M17/M17Datatypes.hpp
```diff
@@ -74,24 +81,40 @@ enum M17ScramblingType
...
+    uint8_t     data_src        : 4;  //< Data source
+    uint8_t     bearing_1       : 1;  //< MSB of bearing
+    uint8_t     radius          : 3;  //< estimate of lateral uncertainty, based on HDOP value
+    uint8_t     radius_valid    : 1;  //< radius data valid
+    uint8_t     velocity_valid  : 1;  //< speed and bearing valid
+    uint8_t     alt_valid       : 1;  //< Altitude data valid
+    uint8_t     coords_valid    : 1;  //< Coordinate data valid
+    uint8_t     bearing_2       : 8;  //< Lower 8 bits of bearing, heading direction for the velocity in degrees, never to exceed 359; e.g. north is 0
+    union {
+        struct {
```
silseva: I see you're accessing these fields using the array: then we can probably drop the `union` and `struct` and leave only the array

openrtx/include/protocols/M17/M17LinkSetupFrame.hpp
```diff
@@ -140,6 +141,15 @@ class M17LinkSetupFrame
...
     lich_t generateLichSegment(const uint8_t segmentNum);
 
+    /**
+     * @brief Set the GNSS data for the LSF's meta feature. This method not only
+     * handles converting the datatypes, but it also handles the byte swapping
+     * necessary since M17 is big endian but the runtime is little endian.
+     *
+     * @param position
+     */
+    void setGnssData(const gps_t *position);
```
silseva: I'd pass the station type as the second parameter: not all the OpenRTX targets with M17 are (or will be) handelds

openrtx/src/rtx/OpMode_M17.cpp
```diff
@@ -335,6 +354,47 @@ void OpMode_M17::txState(rtxStatus_t *const status)
         modulator.sendFrame(m17Frame);
     }
 
+    if(lsfFragCount == 6)
```
silseva: @kd0oss can you please give me a brief explanation about this logic? :)

openrtx/include/protocols/M17/M17Datatypes.hpp
turnrye: Can't use the enum within a bit field though, so they have to stay this way

silseva: You can :)
See here: https://github.com/OpenRTX/OpenRTX/blob/4b9f75fe81092947eaef586d304cf10662cbb6ff/openrtx/include/core/settings.h#L57
`display_timer` is a bitfield but it its assigned a value from an `enum`

openrtx/src/rtx/OpMode_M17.cpp
kd0oss: This just makes sure six frames have been sent to make a complete LSF before we change anything.

openrtx/include/protocols/M17/M17Datatypes.hpp
turnrye: Thanks! Forgot that I didnt have to set the enum type on the bitfield itself. Eventually since we want to use these in a target hwconfig file, which is C, I think I need to move these enums out of this file. Where would they belong best?

openrtx/src/rtx/OpMode_M17.cpp
silseva: Thanks. Actually this logic can be moved inside the FrameEncoder, to have it set up once for all the code touching the LSF.

turnrye: So is your thought that lsfFragCount should be a member of M17FrameEncoder, handling tracking internally, and then there's a public method `shouldSendLsf()` that then we use for this if statement?

silseva: Even better: a variable equivalent to `lsfFragCount` is already present inside the M17FrameEncoder class because you have to cyclically append different LICH chunks alongside an M17 stream. This means that we can add a logic there to automatically re-generate all the chunks from the LSF when the counter is zero. From the application layer, the only thing to do is to send a new LSF to the the FrameEncoder.
I'll implement this myself and then ask for testing.

openrtx/include/protocols/M17/M17Datatypes.hpp
silseva: Let's leave the things as they're for now, the PR is already good :)

silseva:
All good, will merge it soon into a dedicated branch so that I can implement the automatic LSF chunk update.
Throwing there also an idea: should we add a settings entry to enable/disable the GNSS transmission?

turnrye:
> All good, will merge it soon into a dedicated branch so that I can implement the automatic LSF chunk update. Throwing there also an idea: should we add a settings entry to enable/disable the GNSS transmission?

When we add the meta text option, I think we should also add a settings entry for controlling what meta is sent (none, text, gnss, text+gnss). But since GNSS already has a dedicated menu option, I'd think this is OK to go to master for now, and we just make sure to add that change before the feature is in a release.

turnrye:
I'll work on squashing and rebasing this, since at this point the implementation is relatively stable and @silseva plans to take up the final steps.

Separate note, a few of us have noticed that at times the GPS state is set as having a fix yet the lat and lon are incorrectly 0. There may be a bug around this, perhaps outside of this PR, that needs to be addressed.

jancona:
> When we add the meta text option, I think we should also add a settings entry for controlling what meta is sent (none, text, gnss, text+gnss). But since GNSS already has a dedicated menu option, I'd think this is OK to go to master for now, and we just make sure to add that change before the feature is in a release.

Can't you get that by setting meta text (or not) and turning the GPS on (or not)? What's the use case for setting meta text and then not transmitting it? I guess you could turn on the GPS to see your lat/long or set the clock, while not wanting it to transmit, but that seems like a pretty marginal case.

turnrye:
> > When we add the meta text option, I think we should also add a settings entry for controlling what meta is sent (none, text, gnss, text+gnss). But since GNSS already has a dedicated menu option, I'd think this is OK to go to master for now, and we just make sure to add that change before the feature is in a release.
> 
> Can't you get that by setting meta text (or not) and turning the GPS on (or not)? What's the use case for setting meta text and then not transmitting it? I guess you could turn on the GPS to see your lat/long or set the clock, while not wanting it to transmit, but that seems like a pretty marginal case.

Yes, you're right that it can be managed through those other settings. Let's discuss more on the next PR once that's up (~ end of this week hopefully).

silseva:
Agreed, there is no need for an additional setting. We just need to make it clear that, when you press PTT on M17 and your GPS is enabled, you're also transmitting your position.

turnrye:
I've connected with @silseva, and we agreed for me to give the LSF automated chunk update changes a try. I'll be working on those next and continuing to rewrite history a bit as the sole developer on this branch.

turnrye:
Hey @silseva this is ready for review again!

Re identifying when it's safe to update LSF, I kept with the simple flag based approach rather than adding a second buffer to the encoder. I felt this performed well enough in my testing. I also recognize we have the chance to revisit this with my next PR for text metadata.

openrtx/src/rtx/OpMode_M17.cpp
```diff
@@ -300,21 +313,22 @@ void OpMode_M17::rxState(rtxStatus_t *const status)
 
 void OpMode_M17::txState(rtxStatus_t *const status)
 {
+    static streamType_t type;
     frame_t m17Frame;
+    static int16_t gpsTimer;
```
silseva: Make it a private member variable, please :)

openrtx/src/rtx/OpMode_M17.cpp
```diff
@@ -335,6 +349,28 @@ void OpMode_M17::txState(rtxStatus_t *const status)
         modulator.sendFrame(m17Frame);
     }
 
+    gps_t gps_data;
+    gps_data = state.gps_data;
+    if(encoder.canUpdateLsf() && state.settings.gps_enabled && (gpsTimer == -1 || gpsTimer >= MINIMUM_GPS_TIME_TICKS) && gps_data.fix_type > FIX_TYPE_NOT_AVAIL)
```
silseva: Two things:

1. For the first initialization, I'd just put
```
if(state.settings.gps_enabled)
    lsf.setGnssData(&gps_data, M17_GNSS_STATION_HANDHELD);
```
in the block preparing the LSF for the first transmission. This saves you from having to set `gpsTimer` to -1 and have a cumbersome condition checking.

2. I'd move the `gps_data.fix_type > FIX_TYPE_NOT_AVAIL` check inside the `setGnssData` function: there is the place where you're composing the data block and setting the various validity flags

openrtx/src/rtx/OpMode_M17.cpp
```diff
@@ -335,6 +349,28 @@ void OpMode_M17::txState(rtxStatus_t *const status)
...
+    {
+        gpsTimer = 0;
+        gpsTransmitting = true;
+        lsf.setGnssData(&gps_data, M17_GNSS_STATION_HANDHELD);
+        type.fields.encType    = M17_ENCRYPTION_NONE;
```
silseva: I'd move the setting of the LSF type field inside the `setGnssData` function itself: being it a member function of the M17LinkSetupFrame class, it has access to all the fields of the LSF data structure.

openrtx/src/protocols/M17/M17LinkSetupFrame.cpp
```diff
@@ -158,3 +159,40 @@ uint16_t M17LinkSetupFrame::crc16(const void *data, const size_t len) const
...
+
+    uint16_t speed = (uint16_t)position->speed * 2;
+    gnssData.speed_1 = speed >> 4; // MBS
+    gnssData.speed_2 = speed & 0x0F; // Lower 4 bits
+
+    // Structure numeric fields that need offset and steps
+    uint16_t alt = (uint16_t)1000 + position->altitude * 2;
+    gnssData.altitude = __builtin_bswap16(alt);
+
+    data.meta.gnss_data = gnssData;
```
silseva: Instead of creating a local copy of  `gnssData` on the stack, filling it data and then copying it into the corresponding field of the LSF, I'd just access `data.meta.gnss_data` directly.

openrtx/include/rtx/OpMode_M17.hpp
```diff
@@ -148,6 +149,10 @@ class OpMode_M17 : public OpMode
     M17::M17Demodulator  demodulator;  ///< M17 demodulator.
     M17::M17FrameDecoder decoder;      ///< M17 frame decoder
     M17::M17FrameEncoder encoder;      ///< M17 frame encoder
+    M17::M17LinkSetupFrame lsf;        ///< M17 link setup frame
+
+
+    static constexpr int16_t MINIMUM_GPS_TIME_TICKS = 5 * 30; // Wait at least 5 seconds between GPS transmissions
```
silseva: Do you mind moving this above all the variables' declarations? Usually I tend to have:
- private member functions
- private constants
- private variables

turnrye:
@silseva thanks for the quick feedback! I've addressed the issues you raised and left it as a separate fixup commit so that you can trace the individual changes if you wish. I've also verified that the changes still work with a quick "happy path" test. Please take another look at your convenience and let me know if there is any further feedback.

PR #96, opened 2022-07-17
WIP: Remove default include paths

carpikes:
This PR removes all the ambiguous include path, leaving only `openrtx/` and `platform/`. This PR in WIP, do not merge.

PR #361, opened 2025-09-25
Remove default include paths (take 2)

### LINKED ISSUE #359: [TG-730] Address ambiguous and inconsistent include paths
AS A developer,
I WANT simple, consistent, and clear include paths
SO THAT I can make my contributors more easily and review other peoples commits more easily.

## Acceptance Criteria

- The only include paths in build.meson should be openrtx/ and platform/
- The conventions should be followed for <> vs "" as such:
  > 
  >    #include <..> indicates an include of system libraries, miosix libraries, external libraries
  >    #include ".." indicates an include of openrtx libraries, local (current directory) libraries.

# Rationale

Meson is configured to default include a number of specific directories, which is problematic because it:

- makes imports ambiguous and harder to understand
- forces filenames to be unique-ish, which leads to some odd conventions to avoid collisions
- obfuscates important feedback for contributors where jumping across "modules" in the project isnt immediately obvious, though in many places it's discouraged or even forbidden

# Existing WIP

@carpikes in https://github.com/OpenRTX/OpenRTX/pull/96 addressed this, but the PR is quite outdated and this point and therefore will need some additional work. This issue represents reviving that work and chasing this down to closure.

turnrye:
This work is inspired by https://github.com/OpenRTX/OpenRTX/pull/96. I re-did the effort though as there were some substantial changes to the project structure since this PR was opened and last rebased.

Unlike #96, I chose _not_ to clang-format the include blocks, as our current clang-format preset doesn't have the `SortIncludes` or `IncludeBlocks` set to make any changes. If we do think that the imports should have a specific sort or block structure, we should revise the clang-format config at the same time.

Closes #359.


silseva:
Looks good! I'll merge it later

PR #362, opened 2025-09-27
hr-c6000 driver: add method to read an spi 16 bit register

turnrye:
Implement the helper method for reading 16 bit registers from the HR-c6000. This is based on the datasheet's instructions about how to read. I've done some light testing of this, showing that when I write to a register and then read it back, it has the expected value.

This may be helpful to developers implementing more features on this baseband chip. I'm using this to debug my DTMF code.

### DIFF
--- a/platform/drivers/baseband/HR_C6000.h
+++ b/platform/drivers/baseband/HR_C6000.h
@@ -92,9 +92,6 @@ public:
 
 private:
 
-    static const uint8_t SPI_IS_READ_FLAG = 0x80;
-    static const uint8_t SPI_EXPAND_FLAG = 0x40;
-
     /**
      * Write a register with 16-bit address.
      *
@@ -106,7 +103,7 @@ private:
     {
         uint8_t data[4];
 
-        data[0] =  static_cast< uint8_t >(opMode) | SPI_EXPAND_FLAG;
+        data[0] =  static_cast< uint8_t >(opMode) | static_cast< uint8_t >(SpiFlags::EXPAND);
         data[2] = (addr >> 8) & 0x07;
         data[1] = addr & 0xFF;
         data[3] = value;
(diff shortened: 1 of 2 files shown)


platform/drivers/baseband/HR_C6000.h
```diff
@@ -103,14 +106,39 @@ class HR_C6000 : public HR_Cx000 < C6000_SpiOpModes >
...
+
+        const uint8_t READ_FLAG = 0x80;
+        const uint8_t EXPAND_FLAG = 0x40;
+
+        data[0] =  static_cast< uint8_t >(opMode) | SPI_EXPAND_FLAG | SPI_IS_READ_FLAG;
+        data[2] = (addr >> 8) & 0x07;
+        data[1] = addr & 0xFF;
+
+        ScopedChipSelect cs(uSpi, uCs);
+        spi_send(uSpi, data, 4);
```
turnrye: Should be len(data) aka 3, not 4.

platform/drivers/baseband/HR_C6000.h
```diff
@@ -92,6 +92,9 @@ class HR_C6000 : public HR_Cx000 < C6000_SpiOpModes >
 
 private:
 
+    static const uint8_t SPI_IS_READ_FLAG = 0x80;
```
silseva: These can become an `enum class` and be moved to https://github.com/OpenRTX/OpenRTX/blob/master/platform/drivers/baseband/HR_Cx000.h

turnrye: Done! I did _not_ implement the | operator (like exists https://github.com/OpenRTX/OpenRTX/blob/master/platform/drivers/baseband/HR_Cx000.h#L277), so static casts had to be added. I can tinker with this if you think it'd be good to implement.

PR #363, opened 2025-09-30
Draft: M17 protocol callsign class

turnrye:
This PR refactors to add a class to handle M17 callsigns. Ref https://tasks.openrtx.org/project/openrtx/us/733

TODO: 
- dig in to openrtx/src/rtx/OpMode_M17.cpp to see where types there should just change altogether to the new class rather than instantiating so many objects repeatedly
- review class from a memory management perspective, also what can be made const


silseva:
Do you mind changing the code to use `char *` instead of `std::string`? Having `std::string` as a constructor parameter or return value is perfect, but I'd not use it for the internals. This makes the code a bit more simple and, probably, even a bit smaller.

turnrye:
> Do you mind changing the code to use `char *` instead of `std::string`? Having `std::string` as a constructor parameter or return value is perfect, but I'd not use it for the internals. This makes the code a bit more simple and, probably, even a bit smaller.

Got it! I'll update the private methods so that internally they use cstrings.

turnrye:
I've updated this now to not use strings internally for the callsign comparison and for decoding. Let me know if you have any other feedback!

openrtx/src/protocols/M17/Callsign.cpp
```diff
@@ -0,0 +1,217 @@
...
+{
+    call_t encodedCall;
+    encode_callsign(std::string(this->call), encodedCall);
+    return encodedCall;
+}
+
+/**
+ * NOTE! since only incomingCs is checked for special values, the second arg must be the incoming station
+ */
+bool compareCallsigns(const char *localCs, const char *incomingCs)
```
silseva: Please make this function a private member of the `Callsign` class or embed it into the `operator==` implementation

openrtx/include/protocols/M17/Callsign.hpp
```diff
@@ -0,0 +1,88 @@
...
+            * value first, leading to 0-padding of callsigns shorter than nine characters.
+            *
+            * \param callsign the callsign to encode.
+            * \param encodedCall call_t data structure where to put the encoded data.
+            * \param strict a flag (disabled by default) which indicates whether invalid
+            * characters are allowed and assigned a value of 0 or not allowed, making the
+            * function return an error.
+            * @return true if the callsign was successfully encoded, false on error.
+            */
+    bool encode_callsign(const std::string &callsign, call_t &encodedCall,
```
silseva: Please rename this function to `encode` and realign the comments

openrtx/include/protocols/M17/Callsign.hpp
```diff
@@ -0,0 +1,88 @@
...
+                         bool strict = false) const;
+
+    /**
+            * Decode a base-40 encoded callsign to its text representation. This decodes
+            * a 6-byte big-endian value into a string of up to 9 characters.
+            *
+            * \param encodedCall base-40 encoded callsign.
+            * \return a string containing the decoded text.
+            */
+    const char *decode_callsign(const call_t &encodedCall, char *out) const;
```
silseva: Same as the other. :)

openrtx/src/protocols/M17/Callsign.cpp
```diff
@@ -0,0 +1,217 @@
...
+    std::strncpy(call, callsign, CALLSIGN_MAX_CHARS);
+    call[CALLSIGN_MAX_CHARS] = '\0';
+}
+
+Callsign::Callsign(const call_t encodedCall)
+{
+    decode_callsign(encodedCall, this->call);
+}
+
+bool Callsign::encode_callsign(const std::string &callsign, call_t &encodedCall,
```
silseva: Can you change the function so that i takes an `char *` as input argument? I'd like to completely eradicate the use of `std::string` inside this class.
The for loop is using a reverse iterator, scanning the string from the end towards the beginning.
We can also remove the `bool` return value, because this function now is being called only from inside the `Callsign` class and this ensures that the maximum length of the input string is 9 characters.

openrtx/src/protocols/M17/Callsign.cpp
```diff
@@ -0,0 +1,217 @@
...
+        }
+    }
+
+    auto *ptr = reinterpret_cast<uint8_t *>(&encoded);
+    std::copy(ptr, ptr + 6, encodedCall.rbegin());
+
+    return true;
+}
+
+const char *Callsign::decode_callsign(const call_t &encodedCall,
```
silseva: Also here, we can make it return a `void`

openrtx/include/protocols/M17/Callsign.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+#include <string>
+#include "M17Datatypes.hpp"
+
+namespace M17
+{
+
+class Callsign {
+public:
+    static constexpr size_t CALLSIGN_MAX_CHARS = 9;
```
silseva: This can be private

openrtx/include/protocols/M17/Callsign.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+    Callsign(std::string callsign);
```
silseva, line 46: Please add headers to these functions

openrtx/include/protocols/M17/Callsign.hpp
```diff
@@ -0,0 +1,88 @@
...
+     * Answer if a callsign is empty
+     * 
+     * @return true if the first char of the callsign is a NUL
+     */
+    inline bool empty() const
+    {
+        return call[0] == '\0';
+    }
+
+    bool operator==(const Callsign &other) const;
```
silseva, line 79: Ditto

turnrye:
Thank you for providing such specific feedback. I've addressed each point, and I added a few unit tests just to make sure that my changes to the existing callsign encoding utility didn't break it. Please give this another lookover and let me know what additional feedback you have. Thanks again!

turnrye:
Hey @silseva ! I think I've addressed all of your feedback, please let me know if there are any other things that I need to consider. I also added some naive unit tests since this saw me changing the encoder/decoder behaviors a bit (before this, all I did was move code around). Thank you!

turnrye:
Erm, hang on, I need to add docs for these too:
```
/home/turnrye/workspace/OpenRTX/openrtx/include/protocols/M17/Callsign.hpp:34: warning: Compound M17::Callsign is not documented.
/home/turnrye/workspace/OpenRTX/openrtx/include/protocols/M17/Callsign.hpp:68: warning: parameters of member M17::Callsign::operator== are not documented
```

Edit: OK, added those too. Putting this out of draft.

PR #367, opened 2025-10-04
DTMF Support

turnrye:
# What

Enable sending DTMF tones when on the FM mode. As part of this, it meant juggling a few keys in the UI to make room.

    - 1750 Hz access tone now occurs by pressing the F1 key while pressing PTT
    - F1 key still is used for accessibility announcements otherwise
    - # key now is DTMF #, as are the rest of the DTMF characters; this is currently only implemented for the GDx and UVx platforms
    - You can no longer change VFO frequency while pressing PTT

Note that while these changes are all across all targets besides module17, the DTMF send capability is only implemented in this PR for the GDx and UV3x0 platforms. Additional baseband drivers need to be contributed for other platforms separately.

# How

- Use state and rtxSync to copy the pressed key between UI and RTX threads
- Use a FM protocol DTMF namespace to convert DTMF chars to tones
- Implement driver methods for at1846s to send DTMF tones similar to how the 1750 hz access tone is sent

# Future Steps

- Implement this for more platforms
- Implement a tone monitor so that the action is more apparent to the user


openrtx/src/protocols/FM/DTMF.cpp
```diff
@@ -0,0 +1,41 @@
...
+static constexpr DtmfKeyEntry dtmfKeyMap[] = {
+    { '1', 697, 1209 }, { '2', 697, 1336 }, { '3', 697, 1477 },
+    { 'A', 697, 1633 }, { '4', 770, 1209 }, { '5', 770, 1336 },
+    { '6', 770, 1477 }, { 'B', 770, 1633 }, { '7', 852, 1209 },
+    { '8', 852, 1336 }, { '9', 852, 1477 }, { 'C', 852, 1633 },
+    { '*', 941, 1209 }, { '0', 941, 1336 }, { '#', 941, 1477 },
+    { 'D', 941, 1633 }
+};
+
+DtmfKeyEntry lookupDtmfFreq(char symbol)
```
silseva: I think we can move everything in this file, the code is not that much: https://github.com/OpenRTX/OpenRTX/blob/master/platform/drivers/baseband/radioUtils.h

openrtx/include/core/input.h
```diff
@@ -86,6 +86,14 @@ bool input_isCharPressed(kbd_msg_t msg);
  */
 uint8_t input_getPressedNumber(kbd_msg_t msg);
 
+/**
+ * This function returns the DTMF character associated with a key.
+ *
+ * @param key: the key to convert
+ * @return the DTMF character associated with the key
+ */
+char input_keyToDtmfChar(uint32_t key);
```
silseva: Do you mind renaming the function to `input_getDtmfChar`?

turnrye:
This is in draft because despite the tones it produced decoding properly using various tools , it is _not_ working to signal my allstar node for control. I want to debug this further to determine if this is a "just me" problem or with the feature overall.

silseva:
Put back to draft, mark it ready once you finish! :)

turnrye:
Closing because Im not actively working on this -- hope to pick this back up at some time in the future when I have more time to troubleshoot and research.

PR #368, opened 2025-10-08
M17 meta text receiving

turnrye:
# What

This change adds support for reassembling M17 meta text and displaying it.

# How

This change is based on the work originally done in https://github.com/OpenRTX/OpenRTX/pull/314, current as of https://github.com/OpenRTX/OpenRTX/pull/314/commits/e8c3ea54b93fe894b86d69c29448fdcdb236bb7d. The changes there had whitespace changes removed, then were squash merged onto a new branch. From there, changes that were unrelated to the M17 meta text feature were removed. Then, changes consistent with feedback from previous commits were made to try to simplify code review.

The work in this PR was done by @kd0oss Rick. I am helping pick, address feedback, and land these changes.

# Potentially controversial decisions

- When for a given transmission both an extended callsign and a meta text are received, only display the meta text (ui code decision); this is to avoid having to introduce another line of text, or squash the text together into the same line; I think this should be revisited though when there's a redesign of the VFO receive mode
- No attempt is made to deal with received meta text changing mid transmission (i.e. some blocks are from meta text A but then some blocks are from meta text B); this is ok per the spec since meta text isn't expected to change, but edge cases are certainly possible
- If meta text blocks are received out of order, they'll still be inserted, but the display behavior may have excessive empty characters where the missing blocks should be; e.g. if the last block is the only received block, the previous blocks will be shown on the UI as empty rather than trimming whitespace or hiding the incomplete text; again this is generally consistent with the M17 spec but this may not be the best UX; would love user feedback on how often this happens and what their expectations would be

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -40,17 +40,17 @@ jobs:
             ls build_cm7/ &&
             ls build_arm/ &&
             ls build_linux/ &&
-            mv build_cm7/openrtx_cs7000p_wrap openrtx_cs7000p_${{ env.BRANCH_NAME }}.bin &&
-            mv build_cm7/openrtx_cs7000p_dfu openrtx_cs7000p_${{ env.BRANCH_NAME }}.dfu &&
-            mv build_arm/openrtx_cs7000_bin openrtx_cs7000_${{ env.BRANCH_NAME }}.bin &&
-            mv build_arm/openrtx_cs7000_dfu openrtx_cs7000_${{ env.BRANCH_NAME }}.dfu &&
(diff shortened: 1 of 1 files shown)


turnrye:
The build failure appears to be a CI platform network blip, it just needs to be re-run.

turnrye:
Setting as draft to try to minimize code reviewer work-in-progress. This is a lower priority change compared to my other PRs. Please help me get those through review first.

turnrye:
I've synced this with the latest master. The last thing I need to do before taking this back out of draft is to split up the commit into smaller, more focused changes that stack to build this feature. Will do that soon.

turnrye:
Moved to https://github.com/OpenRTX/OpenRTX/pull/390; the CI was complaining about the branch name having "/" in it here.

PR #371, opened 2025-10-12
Update License and Copyright Header

turnrye:
This change:

- Updates the copyright across this project to cover years 2020-2026
- Makes the license header the same on every source file that can have it
- Adds the license header in a few places where it was missing
- Adds an inclusive copyright statement rather than individual authors, that way it's clear that copyrights continue to be individually held despite not having each specific contributor listed
- Adopts a tool for checking and setting the copyright header
- Adds a CI job to ensure that files continue to have the proper header (including new files)
- Codifies in the tool's configuration what files are _not_ original works of this project and therefore not having the license header


silseva:
Now you made me think: given that we're changing all the copyright headers, what if we update everything to use the SPDX licence identifier?

turnrye:
> Now you made me think: given that we're changing all the copyright headers, what if we update everything to use the SPDX licence identifier?

@silseva are you suggesting to add this to the top in addition to the changes this branch has, or to replace the header altogether with this one line?
```C
// SPDX-License-Identifier: GPL-3.0-or-later
```
?

silseva:
Something like this, maybe?
```
/*
 * SPDX-License-Identifier: GPL-3.0-or-later
 * Copyright (C) 2020-2025 OpenRTX Contributors
 *
 * This file is part of OpenRTX.
 */
```
Just throwing ideas

turnrye:
Lovely, that was what I was imagining too! I'll make this change.

silseva:
I'll wait to merge this after we close a bit of the current pending PRs, so that we do not have to do the same work twice

silseva:
What about using this tool from FSFE for license checking? https://github.com/fsfe/reuse-tool

turnrye:
> What about using this tool from FSFE for license checking? https://github.com/fsfe/reuse-tool

I'll look into it, thanks!

turnrye:
Hey @silseva ! I think this is in a decent spot. The `toml` file reveals a lot of things to be improved, which may or may not be suitable for this PR. I wonder if you could take a cursory glance at this and let me know your thoughts on the mergeability of it? 

I have the following doubts:
- Do you want to wait a few days till 2026 and update the headers for that?
- Do you want me to add the header to files that were missing it previously that are part of the project (i.e. project copyright and gpl3+)?

silseva:
Hi Ryan!
Everything looks good! I think we can ignore the licenses of the files in the `lib` and `subprojects` directories, consistently with the reuse specification. The files in `lib` are not git submodules, but in the long run they will become.

PR #376, opened 2025-10-21
Add memory utils

Supermagnum:


openrtx/include/core/memory_utils.h
```diff
@@ -0,0 +1,224 @@
...
+    }
+    
+    return new_ptr;
+}
+
+/**
+ * @brief Safe memory deallocation with null pointer check
+ * @param ptr Pointer to memory to free
+ */
+static inline void safe_free(void *ptr)
```
silseva, line 93: There is no need for this. From [man free](https://linux.die.net/man/3/free):
```
 If ptr is NULL, no operation is performed. 
```

openrtx/include/core/memory_utils.h
```diff
@@ -0,0 +1,224 @@
...
+}
+
+/**
+ * @brief Safe memory copy with bounds checking
+ * @param dest Destination buffer
+ * @param src Source buffer
+ * @param size Number of bytes to copy
+ * @return 0 on success, -1 on error
+ */
+static inline int safe_memcpy(void *dest, const void *src, size_t size)
```
silseva, line 107: We're in a case where pointer set to zero causes a segmentation fault (Linux) or is aliased to some read-only memory region (STM32), so wrong parameters passed to a `memcpy` will cause the program to crash or notably misbehave. Moreover, a `memcpy` of size 0 does nothing.
All in all, I think we can safely skip these checks. We can discuss about that, in any case

silseva:
I'm sorry, but we're currently unable to accept this PR. I also see that it adds new functions but there is no change in the codebase to use them.

PR #377, opened 2025-10-21
Add crypto utils

Supermagnum:


silseva:
Currently we're not using any cryptographic system, but we'll reopen this PR when necessary.

PR #378, opened 2025-10-21
Add string utils

Supermagnum:


silseva:
We're in a case where pointer set to zero causes a segmentation fault (Linux) or is aliased to some read-only memory region (STM32), so wrong parameters passed to a copy functions will cause the program to crash or notably misbehave. Moreover, a copy of size 0 does nothing. All in all, I think we can safely skip these checks.

PR #382, opened 2025-10-22
M17 packet: add net-new protocol definitions for packet support

turnrye:
This commit adds the various M17 protocol definitions that are unique to packet mode. Subsequent commits will implement the updates to the protocol mechanisms to utilize these.

Ref https://tasks.openrtx.org/project/openrtx/task/756


silseva:
Thinking out loud: do we really need a packet class?

turnrye:
> Thinking out loud: do we really need a packet class?

No, but it's there for consistency with M17StreamFrame.

turnrye:
Closing because I'm not actively working on this at the moment. I hope to resume this once some of my higher priority work gets complete first.

Issue #369, opened 2025-10-08
Sporadic errors while decoding M17 speech frames on various devices


sp5wwp:
It seems like sometimes the M17 decoder outputs wrong data. That causes Codec2's decoder to receive two invalid frames (as there are two per a single M17 frame) and output 80ms of noise (so-called *blerp* or *bird tweeting*). A similar issue has been reported already with the Module17: #254 The underlying cause has not yet been definitely identified.

A suggested workaround: keep track of the frame number (FN) and only decode the speech frame contents if the received FN is in immediate succession to the last received one (i.e. FN increment of 1). In case of sequence error, replace the missing frame data with zeros. It offers the most comfortable alternative, as compared with replacing it with the last known Codec2 data (repeating) or fiddling with the energies to mimic exponential decay (sounds irritating and unnatural).

Caveat: this might cause problems when a single frame goes "missing", then the rest of the stream will be muted. A simple fix would employ using a timer synchronized with syncwords, incrementing the expected FN

lwvmobile:
> (so-called _blerp_ or _bird tweeting_). A similar issue has been reported already with the Module17: [#254](https://github.com/OpenRTX/OpenRTX/issues/254) The underlying cause has not yet been definitely identified.

This is just my own personal theory, for what it worth, but anytime there is a bit error, or any number of bit errors, the bird chirping or squeaky sound occurs. It could be from a complete garbage frame being decoded as valid, or from enough errors in a valid frame. Since the RF application is using Viterbi, perhaps using the error or cost metric should be a determining factor into whether or not to decode the audio frame, or to substitute silence or drop the audio for that short segment if the metric exceeds a certain threshold where the validity of the data or voice frames cannot be guaranteed.

silseva:
The source of the issue is exactly the presence of bit errors not recovered by Viterbi. From past experiments done by Wojciech, we know that we can set a threshold on the cost metric and discard bad frames quite reliably. If someone can help implementing this, it would be nice

<img width="570" height="327" alt="Image" src="https://github.com/user-attachments/assets/4c7483eb-ed54-4d2d-90a6-03ae4376f56a" />

turnrye:
@silseva , please assign this issue to me. I'll give this a try!

Edit: https://github.com/OpenRTX/OpenRTX/compare/master...turnrye:OpenRTX:tryfix/discard-high-error-audio is where I have put a very basic implementation of this up. Will enlist some testers and see what we find.

[...]

Issue #346, opened 2025-09-10
[Epic] Code is formatted uniformly


turnrye:
The project is adopting code formatting conventions that are the same as the linux kernel. The clang-format tool has been integrated for some time in the project, but the project is not uniformly formatted today. Let's address that. Our vision is a project that has an updated, uniform code format across the project with tools that make it easy for developers to follow.

## Step 1. Formally adopt the new format

https://github.com/OpenRTX/OpenRTX/pull/166

## Step 2. Incrementally "ratchet up" coverage

As contributors add new features and fix bugs, code reviewers should encourage them to add their files changed to scripts/clang_format.sh's FILE_LIST. That way, the formatter formats their file, and it is covered by the CI to avoid regressions / enforce the formatting.

In parallel, there may be dedicated efforts to "ratchet up" coverage in specific portions of the codebase during opportune times; e.g. ratcheting up infrequently changed code at the same time, when it would be unlikely to cause conflicts for contributors

## Step 3. The final push

Eventually, the difference between the old FILE_LIST and the new FILE_LIST will be small, and a dedicated effort to tackling ratcheting up the last files can take place. Once this is done, the clang-format script can be updated to remove the manual FILE_LIST, and at that point the entire project will be considered formatted.

silseva:
Tracked in https://tasks.openrtx.org/project/openrtx/us/731

silseva:
Progress tracking continues on Taiga

PR #383, opened 2025-10-24
Tryfix/discard high error audio

turnrye:
Dont forward on voice frames that have viterbi errors so high. This way, noise artifacts as a result of high error rates (but still established lock) happen much less often. A value of 15 is set based on trial and error and user feedback.

Also, prior to this, format the sources for the changing files (ref https://github.com/OpenRTX/OpenRTX/issues/346). This way I can't introduce formatting issues as part of my change, and neither can future contributors to these files :)

Closes https://github.com/OpenRTX/OpenRTX/issues/369

### DIFF
--- a/openrtx/src/protocols/M17/M17FrameDecoder.cpp
+++ b/openrtx/src/protocols/M17/M17FrameDecoder.cpp
@@ -153,8 +153,13 @@ void M17FrameDecoder::decodeStream(const std::array<uint8_t, 46> &data)
     begin += lich.size();
     std::copy(begin, data.end(), punctured.begin());
 
-    viterbi.decodePunctured(punctured, tmp, DATA_PUNCTURE);
-    memcpy(&streamFrame.data, tmp.data(), tmp.size());
+    uint16_t bitErrorCount =
+        viterbi.decodePunctured(punctured, tmp, DATA_PUNCTURE);
+
+    // If Viterbi bit error is too high, skip copying the frame altogether; this avoids passing bad audio
+    if (bitErrorCount < MAX_VITERBI_ERRORS) {
+        memcpy(&streamFrame.data, tmp.data(), tmp.size());
+    }
 }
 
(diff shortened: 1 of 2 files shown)


silseva:
I prefer to have brace wrapping after class, we have to change the corresponding parameter in the .clang-format file to `AfterClass: true`

openrtx/src/protocols/M17/M17FrameDecoder.cpp
```diff
@@ -72,93 +75,95 @@ M17FrameType M17FrameDecoder::decodeFrame(const frame_t& frame)
...
-M17FrameType M17FrameDecoder::getFrameType(const std::array< uint8_t, 2 >& syncWord)
+M17FrameType
+M17FrameDecoder::getFrameType(const std::array<uint8_t, 2> &syncWord)
 {
     // Preamble
-    M17FrameType type   = M17FrameType::PREAMBLE;
-    uint8_t minDistance = hammingDistance(syncWord[0], 0x77)
-                        + hammingDistance(syncWord[1], 0x77);
+    M17FrameType type = M17FrameType::PREAMBLE;
+    uint8_t minDistance =
```
silseva, line 83: I would _really_ like to know why that damn thing of clang-format is doing these disgraces. It is also interesting that it changed this line, but the others remained untouched.
If you have the patience to try understanding how to make the tool formatting the lines like it was before, it'd be super. Otherwise, we can do a clang-format pass, fix these flaws by hand and do not add the C++ files to the list of automatic checks.

turnrye: How about for now for this PR, I drop this commit and _not_ include formatting of these files? That way we can talk about continuing to tweak the formatter's behavior in the formatting task.

silseva: As you prefer. Apart from that borked line, the reformat is good: I'd keep it but not add the file to the clang-format check script

PR #384, opened 2025-10-27
README: add section mentioning weblate

turnrye:
As we adopt Weblate as the tool for providing translations, one of their hosting terms is that the project's README references their support. This change adds that as part of a section referencing the translation effort and sharing a graphic with translation stats.

Ref https://tasks.openrtx.org/project/openrtx/task/743


silseva:
Thank you!

PR #386, opened 2025-10-29
Voice prompt generation script + CI

turnrye:
Voiceprompts today are generated in a separate project that has a lot of capabilities that we don't use. Most contributors don't understand how to generate voice prompts, and the docs over there aren't very detailed. Rather than getting that to work really well over there, this change instead brings the key script to this repo, focuses just on the steps that we use today, and then introduces CI to validate that voice prompts are being kept up-to-date (which double as documentation for how to generate voice prompts).

Note that there are only two changes to the voiceprompt python script, for the purposes of hardcoding variables (L405, L430). The remainder of the file matches what is in the https://github.com/OpenRTX/vpcgen repo.

# How

- Copy the script from the vpcgen repo here
- Overwrite some of its values so that it can be run "naked" and just work
- Check in the sound sources
- Update the comment on `stringsTable_t` to give folks more clue as to how this works
- Add GitHub action that generates the file **just for the purpose of validating that it was done properly in the commit**; it will fail if the user adds a string to the wordlist csv but hasnt run the script

# Future steps

- Refactor how the tooling works so that the linkage between the wordlist csv and the stringsTable_t type is enforced (or maybe consolidated into one thing?)


turnrye:
Closing this PR -- notes were added to https://tasks.openrtx.org/project/openrtx/task/742 so that this isnt lost.

PR #387, opened 2025-10-31
M17 tx gain

turnrye:
# What

- Move audio_codec's tx multiplier values to setting
- Add settings UI entries to allow it to be configured

# Why

So that users who experience low level issues on M17 TX can address them

# How

This is a pretty straight-forward change.

# TODO

- [ ] Add "migration" concept to ensure a sensible default is set
- [ ] The menu says "gain" because that's what users expect, but this is actually a multiplier instead; 0 = dead silence, not "no gain"; find a better name for this, and consider if that changes what the scale should be
- [ ] Consider if the entire approach is wrong from a UX perspective -- for example "TX Gain" as built here is a range from 0-255 which is a lot of (too much?) granularity; also having defaults like "32" are unintuitive; maybe the fix for both is to change this to a log scale?

### DIFF
--- a/openrtx/include/core/settings.h
+++ b/openrtx/include/core/settings.h
@@ -64,6 +64,7 @@ typedef struct
     char    m17_dest[10];         // M17 destination
     bool    showBatteryIcon;      // Battery display true: icon, false: percentage
     bool    gpsSetTime;           // Use GPS to ajust RTC time
+    uint8_t m17_tx_level;          // M17 TX level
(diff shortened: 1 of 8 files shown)


turnrye:
This is blocked on settings persistence. Leaving in draft.

turnrye:
Closing because I'm not actively working on this at the moment. I hope to resume this once some of my higher priority work gets complete first.

PR #388, opened 2025-11-01
ci: fix artifact path for cs7000p build target

turnrye:
The build artifacts were looking for a file ending with , and so it was never picked up. This fixes that problem.

Draft till I confirm that this build includes the missing file.

https://tasks.openrtx.org/project/openrtx/task/778


silseva:
All good!

silseva (approved):

PR #389, opened 2025-11-01
ci: rename build artifacts to align with naming convention

turnrye:
OpenRTX releases have had their artifacts follow a historical pattern, which has both set users expectations and had documentation (both official and unofficial) built. The CI builds don't follow those conventions though, which causes confusion with users who don't have experience doing local builds.

This change moves the build artifacts to match the file naming convention users are accustomed to, and it uses the branch name as the name of the build name that is listed.

Other notable fixes:
- Add missing dm1801 and gd77 artifacts (they were previously looking for "_wrap" which for these platforms doesn't create a file with "wrap" in the name)

Fixes: https://tasks.openrtx.org/project/openrtx/task/779

# But does it work?

This branch has a build on it with successful results. The artifacts contained in the zip are:
```
openrtx_cs7000_ci-artifact-renaming.bin
openrtx_cs7000_ci-artifact-renaming.dfu
openrtx_cs7000p_ci-artifact-renaming.bin
openrtx_cs7000p_ci-artifact-renaming.dfu
openrtx_dm1701_ci-artifact-renaming.bin
openrtx_dm1801_ci-artifact-renaming.sgl
openrtx_gd77_ci-artifact-renaming.sgl
openrtx_linux_ci-artifact-renaming
openrtx_md3x0_ci-artifact-renaming.bin
openrtx_mduv3x0_ci-artifact-renaming.bin
openrtx_mod17_ci-artifact-renaming.bin
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,7 +6,7 @@ on:
 
 env:
   RADIO_TOOL_VERSION: 0.2.2
-
+  BRANCH_NAME: ${{ github.head_ref || github.ref_name }} 
 jobs:
   build:
     runs-on: ubuntu-24.04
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -25,7 +25,7 @@ jobs:
             meson compile -C build_linux openrtx_linux &&
             meson compile -C build_linux openrtx_linux_smallscreen &&
             meson compile -C build_linux openrtx_linux_mod17 &&
-            meson compile -C build_arm openrtx_cs7000_bin &&
```
silseva, line 28: Please restore this. CS7000 firmware _must_ be a plain binary.
CS7000 does not have a secondary bootloader requiring obfuscated ("wrapped") firmware images: this radio uses the DFU bootloader from the bootrom.

silseva:
Once done please squash the fixup commits

silseva:
I did a bit of a mess, sorry.

.github/workflows/main.yml
turnrye: What should L45 be?

silseva: I didn't understand the question, sorry.

silseva (approved):

PR #390, opened 2025-11-12
M17 rx and tx meta text

turnrye:
# What

This change adds support for both receiving and transmitting M17 meta text.

# How

This change is based on the work originally done in https://github.com/OpenRTX/OpenRTX/pull/314, current as of https://github.com/OpenRTX/OpenRTX/pull/314/commits/e8c3ea54b93fe894b86d69c29448fdcdb236bb7d. The changes there had whitespace changes removed, then were squash merged onto a new branch. From there, changes that were unrelated to the M17 meta text feature were removed. Then, changes consistent with feedback from previous commits were made to try to simplify code review. This resulted in some substantial changes in the end, 

The work in this PR was done by @kd0oss Rick. I am helping pick, address feedback, and land these changes.

# Potentially controversial decisions

- When for a given transmission both an extended callsign and a meta text are received, only display the meta text (ui code decision); this is to avoid having to introduce another line of text, or squash the text together into the same line; I think this should be revisited though when there's a redesign of the VFO receive mode
- No attempt is made to deal with received meta text changing mid transmission (i.e. some blocks are from meta text A but then some blocks are from meta text B); this is ok per the spec since meta text isn't expected to change, but edge cases are certainly possible
- If meta text blocks are received out of order, they'll still be inserted, but the display behavior may have excessive empty characters where the missing blocks should be; e.g. if the last block is the only received block, the previous blocks will be shown on the UI as empty rather than trimming whitespace or hiding the incomplete text; again this is generally consistent with the M17 spec but this may not be the best UX; would love user feedback on how often this happens and what their expectations would be

The corresponding vpcgen PR is https://github.com/OpenRTX/vpcgen/pull/2

### DIFF
--- a/openrtx/src/protocols/M17/MetaText.cpp
+++ b/openrtx/src/protocols/M17/MetaText.cpp
@@ -52,6 +52,7 @@ void MetaText::setText(char* srcText)
     }
     memset(text, 0, META_TEXT_MAX_LEN);
     strncpy(text, srcText, META_TEXT_MAX_LEN);
+    resetBlockPaging();
 }
 
 void MetaText::getBlock(meta_t &meta, uint8_t index)
--- a/openrtx/src/rtx/OpMode_M17.cpp
+++ b/openrtx/src/rtx/OpMode_M17.cpp
@@ -322,6 +322,11 @@ void OpMode_M17::txState(rtxStatus_t *const status)
 
         lsf.setType(type);
 
+        if(strlen(state.settings.M17_meta_text) > 0) {
+            metaText.setText(state.settings.M17_meta_text);
+            metaText.getNextBlock(lsf.metadata());
+        }
+
         if(state.settings.gps_enabled) {
             lsf.setGnssData(&state.gps_data, M17_GNSS_STATION_HANDHELD);
             gpsTimer = 0;
@@ -363,6 +368,13 @@ void OpMode_M17::txState(rtxStatus_t *const status)
     // the upcoming superframe.
     if(encoder.superframeBoundary())
     {
+        if(strlen(state.settings.M17_meta_text) > 0)
+        {
+            auto lsf = encoder.getCurrentLsf();
+            metaText.getNextBlock(lsf.metadata());
+            encoder.updateLsfData(lsf);
+        }
+
         if(state.settings.gps_enabled)
         {
             gpsTimer++;


silseva:
The chained `if` clauses processing the content of the metadata field makes me think that, probably, we have to first refactor how the metadata as a whole is managed.
To do so, I believe we should introduce a `Metadata` class representing the generic M17 metadata containing the raw bytes and a type field providing information on how to interpret the content. The LSF class should not care about what the metadata content is except for the type flag (too bad that the information is scattered around all the LSF block, preventing us from adopting a more clean approach). Then, should we subclass the `Metadata` to have a specific class for each metadata type? Or should we stick to have a single class with multiple methods?
Actually this is a case where `dynamic_cast` and RTTI can be helpful, if they're not too heavy.
Given that features like gnss and meta text are more of the _nice to have_ type, I believe we can take the occasion to introduce substantial code improvements alongside with them.

turnrye:
I've rebased this and will be unwinding some of the previously introduced complications (e.g. dynamic casting, OOP metadata) in order to make this step more manageable to review.

turnrye:
Hey @silseva I believe this is ready for review again.

openrtx/include/protocols/M17/MetaText.hpp
silseva, line 1: Don't hate me, but I think is cleaner to convert the code in this file to a C++ class. In the end the `MetaTextState` is basically a class without being `class`.
Also, for meta text reassembly, there is no real need of storing each block _and_ the final string: you can just have the character buffer and an `uint8_t` with the bitmap of the blocks added. The M17 specification provides an hint:
```
The Control Byte is split into two 4-bit fields. The most significant four bits are a bit map of
the message length indicating how many Text Data blocks are required for a complete message.
[...]
The least significant four bits indicate which of the Text Data blocks this text corresponds to.
[...]
Any received Control Byte is OR-ed together by the receiving station, and once the most significant
and least significant four bits are the same, a complete message has been received.
```

openrtx/src/rtx/OpMode_M17.cpp
```diff
@@ -203,32 +203,41 @@ void OpMode_M17::rxState(rtxStatus_t *const status)
...
+                        strncpy(status->M17_src, src, 10);
+                    }
+                    else
+                    {
+                        strncpy(status->M17_src, src, 10);
+                    }
+                }
+                else
+                {
                     strncpy(status->M17_src, src, 10);
```
silseva: The `strncpy(status->M17_src, src, 10)` can be moved to line 202 and done only once.
The code was as it was to avoid copying into `status->M17_src` twice in case the LSF metadata contained an extended callsign, but in the end doing otherwise does not really make any difference.

openrtx/src/rtx/OpMode_M17.cpp
```diff
@@ -203,32 +203,41 @@ void OpMode_M17::rxState(rtxStatus_t *const status)
...
-                    // receiving extended callsign data: in order to always store
-                    // the true source of a transmission, we need to store the first
-                    // extended callsign in M17_src.
-                    //
-                    strncpy(status->M17_src,  exCall1, 10);
-                    strncpy(status->M17_refl, exCall2, 10);
-                    strncpy(status->M17_link, src, 10);
-                } else {
+
+                    if(streamType.fields.encSubType == M17_META_EXTD_CALLSIGN)
```
silseva: A `switch` statement instead of a chain of `if`s is probably easier to read

openrtx/src/ui/default/ui_main.c
```diff
@@ -156,15 +161,48 @@ void _ui_drawModeInfo(ui_state_t* ui_state)
...
+                // Meta text (if present)
+                if(rtxStatus.M17_meta_text[0] != '\0')
+                {
+                    char msg[14];
+                    
+                    // Reset scroll on first display of meta text (after being empty)
+                    if(meta_text_was_empty)
+                    {
+                        ui_state->m17_meta_text_scroll_position = 0;
+                        ui_state->m17_meta_text_last_scroll_tick = 0;
```
silseva: The value returned by `getTick` is monotonically increasing, there is no real need to reset `m17_meta_text_last_scroll_tick` each time. Because of this you can also replace all the `meta_text_was_empty = true` lines with just `m17_meta_text_scroll_position = 0` and simplify the code a little bit.

turnrye:
@silseva thank you for the detailed PR feedback! Ill work on this tonight and have it back to you ASAP.

openrtx/include/protocols/M17/MetaText.hpp
```diff
@@ -0,0 +1,154 @@
...
+    /**
+     * Get total block count from metadata byte 0.
+     */
+    static uint8_t getTotalBlocks(const meta_t &meta)
+    {
+        uint8_t mask = (meta.raw_data[0] & 0xf0) >> 4;
+        // mask is 0x1 => 1, 0x3 => 2, 0x7 => 3, 0xF => 4
+        return __builtin_popcount(mask);
+    }
+
```
turnrye, line 77: I plan to add the setters for these in the next tx meta PR.

openrtx/include/protocols/M17/MetaText.hpp
```diff
@@ -0,0 +1,154 @@
...
+        if (idx >= META_TEXT_MAX_BLOCKS)
+            return false;
+
+        // Set bit for this block and copy data directly to buffer
+        blockMask |= (1 << idx);
+        memcpy(text + (idx * META_TEXT_BLOCK_LEN), meta.raw_data + 1,
+               META_TEXT_BLOCK_LEN);
+        return true;
+    }
+
```
turnrye, line 88: Planning to add getBlock and setText methods for the tx meta pr.

openrtx/include/protocols/M17/MetaText.hpp
silseva: Is not prohibited to add them now :)
This PR introduces the full MetaText class _and_ the RX functionality, while the next one will add the necessary glue code for TX

openrtx/include/protocols/M17/MetaText.hpp
silseva: Same as above! :)

turnrye:
@silseva Alrighty, this is ready for review again! Now it includes tx as well :)

openrtx/include/core/settings.h
```diff
@@ -50,6 +50,7 @@ typedef struct
     char    m17_dest[10];         // M17 destination
     bool    showBatteryIcon;      // Battery display true: icon, false: percentage
     bool    gpsSetTime;           // Use GPS to ajust RTC time
+    char    M17_meta_text[53];    // M17 Meta Text to send
```
silseva, line 53: I'd not put the meta text inside the settings struct, because is not really a configuration parameter like the screen brightness. For the time being, I think we can live for a while having it volatile until we set up proper VFO state persistence. In this regard, also the M17 CAN and destination callsign are not configuration parameters and will be moved from settings to VFO state.

openrtx/include/protocols/M17/M17Datatypes.hpp
```diff
@@ -94,6 +94,14 @@ typedef struct __attribute__((packed))
 }
 gnssData_t;
 
+typedef struct __attribute((packed))
```
silseva: I think we can do without this struct.
The `MetaText` class takes care of handling the metadata information, and its methods take as input parameter the raw `meta_t` type: all the byte manipulation is done inside the methods' implementation and nothing is exposed outside the class. In my opinion is better to not expose "how the sausage is made", because this can lead to confusion and, potentially, misuse of the code.
As I may have already said, the content of this file is not the best and I already plan to refactor it.

openrtx/include/protocols/M17/MetaText.hpp
```diff
@@ -0,0 +1,98 @@
...
+ * 
+ * Meta text can span up to 4 stream frames. This module provides both
+ * single-frame parsing (layer 2) and multi-frame accumulation (layer 3)
+ * for simplicity, as the reassembly logic is straightforward.
+ */
+
+namespace M17
+{
+
+static constexpr uint8_t META_TEXT_MAX_BLOCKS = 4;
```
silseva: Probably is better to have these constants as private members of the class. There is no real need to expose them in the M17 namespace

openrtx/include/protocols/M17/MetaText.hpp
silseva: Adding a comment here, as a follow-up: thanks for having converted the code into a class! There are just a couple of details that need to be fixed:
- please make the order of methods and variables consistent with the other classes. The convention is to have, from top to bottom: public functions, private functions and finally private variables. The [M17 demodulator class](https://github.com/OpenRTX/OpenRTX/blob/master/openrtx/include/protocols/M17/M17Demodulator.hpp) is a good reference
- please add documentation comments to all the methods

openrtx/include/protocols/M17/MetaText.hpp
```diff
@@ -0,0 +1,98 @@
...
+    void reset();
+
+    meta_t getBlock(meta_t &meta, uint8_t index);
+
+    void setText(char *meta);
+
+    /**
+     * Get the block index (0-based) from metadata byte 0.
+     */
+    uint8_t getBlockIndex(const meta_t &meta);
```
silseva: This method can be private, I think. Application code only needs to get/set the whole text, while the lower layer does get/set of raw metadata blocks

openrtx/include/protocols/M17/MetaText.hpp
silseva: Same for `getBlockIndex`: I'd make it a private method.

openrtx/include/core/settings.h
turnrye: I think making the tx metatext value non-persistent would be confusing and unintuitive to the user, since they set this from the settings menu alongside other things that do get persisted. Making this non-persistent would require  moving the menu entry out of settings page and into the new vfo settings page. 

We should also improve this distinction by adding a transmit screen that shows the various fields, that way the fact that it isnt persistent would be discoverable. 

But until those things, I think its best to keep it consistent with the rest of the menu item's behavior. And having a breaking change then when it no longer was persisted seems like a reasonable introduction to the new "volatile" concept, when those other things are in place.

kd0oss: In my mind metatext was suppose to be persistent. This is the way it is in DSTAR. It's kind of a simple greeting or extended identification that you normally set once and forget.

turnrye:
Hey @silseva ! I've added a fixup commit changes addressing your feedback. @kd0oss and I are pushing back on one of your feedbacks, please see the thread for discussion about that. This is ready for another review.

openrtx/include/core/settings.h
silseva: Is it meant to be something independent from the radio channel you're transmitting in? Said otherwise: when we'll have the codeplug system, are we going to possibly have a meta text for each channel stored there or just one for all of them?

turnrye: Meta text would be something set independently of the memory channel, but still a persistent setting. I'd imagine that having one value is fine for now, and a potential future enhancement would be to let the user predefine multiple values and then select from them. I would expect a similar capability for APRS status text as well. My rationale here is, if we take a comparative analysis, we identify these similar features on other modes:

- APRS with its comment status fields (both as a part of the mic-e definition and the standalone status; see [APRS Protocol Ref 1.0](https://www.aprs.org/doc/APRS101.PDF) Chapters 10 and 16)
- DStar has a 4 character "Note" field that is associated with the station callsign, often used to indicate a status or distinguish between multiple radios. There is also a TX Message option for a 20-char text field that is sent in the background during voice transmissions, which users often use to include things like name and location

To look into how these are treated in UIs, lets take a couple of examples. 

For D-star: For the ID-5100 which is a large touchscreen display, the user is able to specify 5 message options. They are able to select from these as they use the radio regardless of vfo or memory mode (see [p. 7-2 - p. 7-4](https://static.dxengineering.com/global/images/instructions/ico-id-5100a-d.pdf)). The ID-52a, which is an HT, has a similar treatment (see [p. 5-17](https://www.icomeurope.com/wp-content/uploads/2024/07/ID-52A_E_PLUS_ENG_AM_0.pdf)).

For APRS: The TH-D75 is a well-regarded APRS capable HT, which allows the sending of position comments. The user is able to select from 8 predefined or 7 custom position comments. Additionally, there is a status text menu option that permits specifying up to 5 comments, with one selected at a time to be included in beacons. In general, the radio treats these as mode settings, and they are set separately from the memory channel (see p. [14-3](https://kasc.kenwood.com/files/prod/3692/5/THD75E_User_manuals_English.pdf)).

kd0oss: Since I was not around when the Metatext field was defined I have no idea what it was actually meant for. I use DSTAR regularly and  the 20 char limit for it's text field limits it's ability to carry any more useful data then what @turnrye mentioned, so most users set it to something simple and rarely touch it again. M17 allows for a little over double that so maybe it would have more uses, although Packet mode offers much more potential for sophisticated communication.

openrtx/src/protocols/M17/MetaText.cpp
```diff
@@ -0,0 +1,163 @@
...
+ * Set what the MetaText text should be
+ */
+void MetaText::setText(char* srcText) 
+{
+    if (!srcText) {
+        text[0] = '\0';
+        return;
+    }
+    strncpy(text, srcText, META_TEXT_MAX_LEN);
+    text[META_TEXT_MAX_LEN] = '\0'; // Always null-terminate
```
silseva: Lines 79-82 can be simplified by doing `memset(text, 0, META_TEXT_MAX_LEN)` before the `memcpy`

openrtx/include/core/settings.h
silseva: I don't really know what the purpose of M17 meta text is, either...
If the main usage is the one of transmitting a status message alongside the voice stream, with this message being independent from the radio being in VFO or memory mode, then we can keep it as part of the radio settings.

turnrye: Yes, that is the main purpose. The use case is similar to those that I referenced in the comparative analysis above, but I am happy to describe this in more detail if needed.

silseva: Is fine, thank you. I already marked the conversation as resolved. :smile:

turnrye:
Hey @silseva this is ready for another review! Thanks!

turnrye:
Hey @silseva I squashed this and addressed the conflict with some of the new things on master. This is ready for review.

silseva:
Perfect! Now that I finished with the NVM work, I'll go back to reviewing the PRs

openrtx/include/protocols/M17/MetaText.hpp
```diff
@@ -0,0 +1,191 @@
...
+     * @brief Return a single meta block at a given index.
+     * 
+     * Generates a properly formatted M17 metadata block containing the control byte
+     * and text data for the specified block index.
+     * 
+     * @param meta Reference to meta_t structure to populate with block data
+     * @param index The 0-based index of the block being fetched (0-3)
+     * @return The populated meta object
+     */
+    meta_t getBlock(meta_t &meta, uint8_t index);
```
silseva: Given that the metadata block to be filled is passed by reference, I think we can avoid to return it as well.

openrtx/include/ui/utils.h
```diff
@@ -0,0 +1,136 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* Number of extra ticks to pause at the beginning and end of scroll */
+#define SCROLL_PAUSE_TICKS 5
+
+/**
+    * Convert scroll position (with pauses) to actual display offset.
```
silseva: The indentation here is broken :)

openrtx/include/ui/utils.h
```diff
@@ -0,0 +1,136 @@
...
+    * @param buf The input buffer containing the text to scroll.
+    * @param printBuf The output buffer where the scrolled text will be stored.
+    * @param printBufSize Size of printBuf in bytes (including room for the terminating NUL).
+    * @param scrollPosition The current scroll position (includes pause ticks).
+    * @return true if the text requires scrolling, false otherwise.
+ */
+static inline bool scrollTextPeek(const char *buf, char *printBuf,
+                                  size_t printBufSize, size_t scrollPosition)
+{
+    if (buf == NULL || printBuf == NULL || printBufSize == 0) {
```
silseva: I think we can skip the NULL pointer checks: is not very elegant, I know. This is an helper function called from the UI, so I'm expecting it to always receive sane values.

openrtx/include/ui/utils.h
```diff
@@ -0,0 +1,136 @@
...
+        return false;
+    }
+
+    /* usable window length (space for characters, excluding NUL) */
+    size_t window = printBufSize - 1;
+
+    /* handle empty window */
+    if (window == 0) {
```
silseva: You can change this to `if (printBufSize <= 1)` and move the assignment of line 58 after the end of the clause

openrtx/include/ui/utils.h
```diff
@@ -0,0 +1,136 @@
...
+        printBuf[0] = '\0';
+        return false;
+    }
+
+    size_t bufLen = strlen(buf);
+
+    /* no scrolling needed: copy text (and pad with spaces to keep stable width) */
+    if (bufLen == 0) {
+        /* empty source */
+        memset(printBuf, ' ', window);
```
silseva: Here we're creating an empty string by setting `printBuf[0] = '\0';`: filling it with spaces can be avoided

openrtx/include/ui/utils.h
```diff
@@ -0,0 +1,136 @@
...
+        printBuf[0] = '\0';
+        return false;
+    }
+    if (bufLen <= window) {
+        memcpy(printBuf, buf, bufLen);
+        if (bufLen < window) {
```
silseva: As above: if we end the string at `window`, do we need to fill the rest?
I'm probably missing something, though

openrtx/include/ui/utils.h
```diff
@@ -0,0 +1,136 @@
...
+ */
+static inline void scrollTextAdvance(const char *buf, size_t printBufSize,
+                                     size_t *scrollPosition)
+{
+    if (buf == NULL || scrollPosition == NULL || printBufSize == 0) {
+        return;
+    }
+
+    size_t window = printBufSize - 1;
+    if (window == 0) {
```
silseva: Also here you can return `if (printBufSize <= 1)` and remove the check at line 105

openrtx/src/protocols/M17/MetaText.cpp
```diff
@@ -0,0 +1,152 @@
...
+#include <new>
+#include <cstddef>
+#include <cstring>
+#include "protocols/M17/MetaText.hpp"
+
+using namespace M17;
+
+MetaText::MetaText()
+{
+    memset(text, 0, sizeof(text));
```
silseva: Is better to call `reset()` here to avoid code duplications: if another variable is added in the future, you have to reset it in a single place

openrtx/src/protocols/M17/MetaText.cpp
```diff
@@ -0,0 +1,152 @@
...
+        text[0] = '\0';
+        return;
+    }
+    memset(text, 0, META_TEXT_MAX_LEN);
+    strncpy(text, srcText, META_TEXT_MAX_LEN);
+}
+
+meta_t MetaText::getBlock(meta_t &meta, uint8_t index)
+{
+    meta = {};
```
silseva: I think is more clear to directly have a `memset` here. (This already compiles to a call to `memset` in assembly)

openrtx/src/protocols/M17/MetaText.cpp
```diff
@@ -0,0 +1,152 @@
...
+    setBlockIndex(meta, index);
+    uint8_t totalBlocks = (strlen(text) + META_TEXT_BLOCK_LEN - 1) / META_TEXT_BLOCK_LEN;
+    setTotalBlocks(meta, totalBlocks);
+    
+    // Fill text portion with spaces (raw_data[1] through raw_data[13])
+    // Do NOT overwrite raw_data[0] which contains the control byte
+    memset(meta.raw_data + 1, ' ', META_TEXT_BLOCK_LEN);
+
+    size_t textLen = strlen(text);
```
silseva: You can put this assignment before line 78, so that you call `strlen` only once.

openrtx/src/protocols/M17/MetaText.cpp
```diff
@@ -0,0 +1,152 @@
...
+        return;
+    }
+    memset(text, 0, META_TEXT_MAX_LEN);
+    strncpy(text, srcText, META_TEXT_MAX_LEN);
+}
+
+meta_t MetaText::getBlock(meta_t &meta, uint8_t index)
+{
+    meta = {};
+    setBlockIndex(meta, index);
```
silseva: I see that, in the end, `setBlockIndex` and `setTotalBlocks` are called only by this function: given that their body is made of few lines, perhaphs we can just put the code here and remove the functions entrely?

openrtx/src/protocols/M17/MetaText.cpp
```diff
@@ -0,0 +1,152 @@
...
+{
+    // According to M17 spec: OR the control bytes together, and once
+    // the most significant and least significant four bits are the same,
+    // a complete message has been received.
+    if (blockMask == 0) {
+        text[0] = '\0';
+        return text;
+    }
+
+    // Find the length based on complete blocks
```
silseva: Should we return the text only when is completely reassembled? That is, partial results are not returned

openrtx/src/ui/default/ui.c
```diff
@@ -1711,6 +1719,37 @@ void ui_updateFSM(bool *sync_rtx)
...
+                        {
+                            if(msg.keys & KEY_ENTER)
+                            {
+                                _ui_textInputConfirm(ui_state.new_message);
+                                // Save selected message and disable input mode
+                                strncpy(state.settings.M17_meta_text, ui_state.new_message, 52);
+                                ui_state.edit_message = false;
+                                *sync_rtx = true;
+                            }
+                            else if(msg.keys & KEY_HASH)
```
silseva: Isn't this going to conflict with M17 destination callsign input?

turnrye:
@silseva thanks for the PR feedback -- Ill work on this tomorrow and get it back to you for review. I'll be out of town Friday - Monday so any additional feedback will be addressed middle of next week.

silseva:
No worries, take your time. I know is a bunch of small, nitpicking, comments which will take a bit of time to deal with.
I just merged your work on the unit tests, so you'll need to rebase onto master and update the unit test as well... :sweat_smile:

turnrye:
Hey @silseva nudge here -- this is ready for review again. I just got back, and I rebased as well.

turnrye:
I found a bug where the Meta Txt menu wasn't getting unselected properly in some scenarios, and I've just fixed this. This is ready for review.

openrtx/include/ui/utils.h
```diff
@@ -0,0 +1,122 @@
...
+ * @param buf The input buffer containing the text to scroll.
+ * @param printBuf The output buffer where the scrolled text will be stored.
+ * @param printBufSize Size of printBuf in bytes (including room for the terminating NUL).
+ * @param scrollPosition The current scroll position (includes pause ticks).
+ * @return true if the text requires scrolling, false otherwise.
+ */
+static inline bool scrollTextPeek(const char *buf, char *printBuf,
+                                  size_t printBufSize, size_t scrollPosition)
+{
+    if (printBufSize <= 1) {
```
silseva: I'm wondering if here we can't just have
```
if (printBufSize <= 1)
    return false;
```

If the `printBuf` is only a single char, we can't really do much in any case.

openrtx/src/protocols/M17/MetaText.cpp
```diff
@@ -0,0 +1,132 @@
...
+    return true;
+}
+
+const char *MetaText::getText()
+{
+    // According to M17 spec: OR the control bytes together, and once
+    // the most significant and least significant four bits are the same,
+    // a complete message has been received.
+    if (blockMask == 0 || blockMask != expectedBlockMask) {
+        static const char empty[] = "";
```
silseva: I'd just return a NULL pointer instead of an empty string

openrtx/src/rtx/OpMode_M17.cpp
```diff
@@ -309,6 +323,12 @@ void OpMode_M17::txState(rtxStatus_t *const status)
 
         lsf.setType(type);
 
+        if(strlen(state.settings.M17_meta_text) > 0) {
+            metaText.setText(state.settings.M17_meta_text);
```
silseva, line 326: Perhaps we can have that `resetBlockPaging` is called inside `setText`, to make it automatic?

silseva:
I added a couple of very small notes, for the rest is good to go!

silseva (approved):

silseva:
I'll try to merge the PR this afternoon/evening.

PR #396, opened 2026-01-29
APRS Receive Support for OpenRTX

rxt1077:
This PR adds APRS receive support for the OpenRTX firmware. It implements a new OpMode and adds functions to the macro menu to change between modes a view the saved APRS packet list. You can see [a video of what it does here.](https://youtu.be/qKSnOaKfMxw)

# Thread Communication Changes

Currently information passes from  _UI thread_ to _RTX Thread_ by modifying _State_ and signalling with _sync_rtx_ that _RTX Config_ needs to be updated (`core/threads.c`). Information passes back from _RTX Thread_ to _UI Thread_ by having _UI Thread_ call `rtx_getCurrentStatus()`. This is used a few times in the M17 mode.

For APRS packet operations, as packets are decoded in _RTX Thread_ they are added head-first to a linked list in _RTX Config_.
In the main thread loop (`core/threads.c`) if APRS packets are available in _RTX Config_ and the OpMode is APRS, _RTX Config_ and _State_ are locked and APRS packets are moved from the _RTX Config_ to _State_. The packets are then cleared from _RTX Config_. In this way _UI Thread_ only has to deal with _State_. _UI Thread_ can read and delete packets as it see's fit without worrying about multiple access issues.

```mermaid
block
    columns 6
    %% row 1
    eventQueue(("Event Queue"))
    space
    uiThread["UI Thread"]
    space:2
    rtxConfig(("&nbsp;RTX Config&nbsp;"))

    %% row 2
    space:3
    syncRtx{"sync_rtx?"}
    space:2

    %% row 3
    devThread["DEV Thread"]
    space
    state(("&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;State&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;"))
    space
    aprsPkts{"aprsPkts?"}
    rtxThread["RTX Thread"]

    %% connections
    eventQueue --> uiThread
    uiThread --> state
    state --> uiThread
    rtxConfig --"rtx_getCurrentStatus"--> uiThread
    syncRtx --> rtxConfig
    state --- syncRtx
    rtxConfig --> rtxThread
    rtxThread --> rtxConfig
    devThread --> eventQueue
    devThread --> state
    rtxConfig --- aprsPkts
    aprsPkts --> state
```
# AFSK1200

The demodulator, slicer, and decoder are designed to exclusively use integer arithmetic and were [prototyped in Python](https://github.com/rxt1077/aprs_prototype). Testing shows they seem to decode around 940 packets on track 1 of the [TNC Test CD](http://www.wa8lmf.net/TNCtest/). Most of it is written in C++, but the final step of converting an HDLC frame to an APRS packet is done in C and stored in a C struct for easier interop between the RTX and UI threads. This code can be found in `src/protocols/APRS`.

# Code Style

I tried to follow the [coding style guidelines](https://openrtx.org/#/coding_style) to the best of my ability and to use the clang-format script. I did end up choosing to pull the UI code out and put it in its own file, `src/ui/default/ui_aprs.c`, so that it could be formated by clang-format and be a little easier to work on. There is a daunting number of lines, #ifdefs, ifs, switch statements, etc. in the UI files for this project.

# Testing

There is a comprehensive test that uses raw samples collected from my radio and completely decodes them.
It can be found in `tests/unit/APRS.c`. I'm also currently running the firmware on my modded Retevis RT3S (MD-UV380). I don't have any of the other OpenRTX-compatible hardware to test on.

# Debugging

The demodulator, slicer, and decoder will print out debug information if `APRS_DEBUG` is defined.


silseva:
Hi!
I had a look at the code and you did a really good work: thanks a lot for it!
In the past weekend I was at FOSDEM with two other OpenRTX developers and we ran your code on a Baofeng DM-1701 for the two days of the event: the radio received a good amount of packets and it was really nice to see it working. We also found some bugs, but this is normal. :)
Before we can do a proper code review, however, please split the PR in more than two commits. The principle is to have a commit dedicated to each logical change applied to the code (addition of new modules, code changes, ...). The commit messages should be written in imperative mood, with the first line providing a short summary and, if necessary, a detailed description in the lines below. Don't forget to sign-off your commits too!

To have an overview of how the commits should be separated, you can have a look at the history of the master branch: https://github.com/OpenRTX/OpenRTX/commits/master/

Regarding commit messages and splitting, you can have a look here (take them as general references, not to be applied literally) :
- https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53
- https://www.kernel.org/doc/html/v4.10/process/submitting-patches.html#describe-your-changes (sections 2 and 3)

Thank you,
Silvano IU2KWO

rxt1077:
That's really cool to hear that you ran it at FOSDEM!

I tried to clean up my commits and I think I have them formatted in a way that complies and logically makes sense.

In terms of bugs, I think the biggest issue for me is that it'll keep collecting packets until it runs out of memory. It really needs a settings menu where you can specify how many to keep and then it'll just replace the oldest when new packets come in.

I'm currently preparing for the Utah Digital Communications Conference, so my time is a bit limited. I wanted to get the PR out before I got too deep into that though.

Thanks for taking a look.

Ryan N2BP

silseva:
Thank you!
I'll review the code in these days

openrtx/src/core/state.c
```diff
@@ -72,10 +73,17 @@ void state_init()
...
         state.settings.brightness = 100;
     }
+
+    state.aprsStoredPkts     = NULL;
+    state.aprsStoredPktsSize = 0;
 }
 
 void state_terminate()
 {
+    /* free APRS packets that have been stored */
```
silseva: Here there is a missing `#ifdef CONFIG_APRS`: the CI fails to compile for the ttwrplus because of this.

openrtx/src/protocols/APRS/Demodulator.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+    // Create the correlation products
+    const int16_t *markIResult = markI.convolve(filteredAudio);
+    const int16_t *markQResult = markQ.convolve(filteredAudio);
+    const int16_t *spaceIResult = spaceI.convolve(filteredAudio);
+    const int16_t *spaceQResult = spaceQ.convolve(filteredAudio);
+
+    // The demodulated signal is mark - space
+    for (size_t i = 0; i < 64; i++)
+        diff[i] = ((abs(markIResult[i]) + abs(markQResult[i]))
```
silseva: Too many parentheses here: I'd split the code in something like
```
int16_t tmp = abs(markIResult[i]) + abs(markQResult[i]);
tmp -= (abs(spaceIResult[i]) + abs(spaceQResult[i]);
diff[i] = tmp >> 3;
```

openrtx/src/protocols/APRS/Demodulator.cpp
```diff
@@ -0,0 +1,118 @@
...
+    const int16_t *filteredAudio = bpf.convolve(input);
+
+    // Create the correlation products
+    const int16_t *markIResult = markI.convolve(filteredAudio);
+    const int16_t *markQResult = markQ.convolve(filteredAudio);
+    const int16_t *spaceIResult = spaceI.convolve(filteredAudio);
+    const int16_t *spaceQResult = spaceQ.convolve(filteredAudio);
+
+    // The demodulated signal is mark - space
+    for (size_t i = 0; i < 64; i++)
```
silseva: Magic number here: can it be replaced with a macro/constexpr variable?

openrtx/src/protocols/APRS/HDLC.cpp
```diff
@@ -0,0 +1,162 @@
...
+                        bit, prevBit, nrziBit, onesCount, workingByte, bitPos);
+
+            prevBit = bit;
+
+            // unstuff bits and watch for flags/errors
+            if ((onesCount == 5) && (nrziBit == 0)) {
+                // if we have 5 ones and get a zero, it's a stuffed bit, disregard
+                onesCount = 0;
+                continue;
+            } else if ((onesCount == 6) && (nrziBit == 0)) {
```
silseva: Perhaps this chain of  `if else` statements can be turned into a state machine based on a `switch`/`case`?
More or less like what I did for the [M17 demodulator](https://github.com/OpenRTX/OpenRTX/blob/1c6725fa5e6e61fdfe56aaa573c76f5401a74a25/openrtx/src/protocols/M17/M17Demodulator.cpp#L256)

openrtx/include/protocols/APRS/Demodulator.hpp
```diff
@@ -0,0 +1,123 @@
...
+
+#include <cstdint>
+#include <cstddef>
+#include <array>
+#include "protocols/APRS/constants.h"
+
+namespace APRS
+{
+
+class Convolver
```
silseva: I didn't go a lot into the details, but is the `Convolver` similar to an IIR filter? If so, is probably better to directly use the [IIR class](https://github.com/OpenRTX/OpenRTX/blob/master/openrtx/include/core/iir.hpp). It uses floating point math, but we can always create a fixed point version as well.

Also, please have a look at the [Correlator class](https://github.com/OpenRTX/OpenRTX/blob/master/openrtx/include/protocols/M17/Correlator.hpp) for M17. My point here is that, if we can have a single implementation covering both cases, it'd be better (less code to maintain).

silseva:
Please add the correct copyright headers to all the files, the reuse CI stage is failing because of that.

openrtx/src/core/state.c
rxt1077: Fixed in fixup commit bb28000

openrtx/src/protocols/APRS/Demodulator.cpp
rxt1077: Fixed in commit 1745683

openrtx/src/protocols/APRS/Demodulator.cpp
rxt1077: Fixed in fixup commit 1745683

rxt1077:
Added copyright headers to all new files in fixup commits 0bc494f 77ed825 9d73aa6 2c285c1 917e0c2 1745683

silseva:
To resolve the various review comments, please use [fixup commits](https://moodledev.io/general/development/abc/fixups) in order to keep the history clean.
The idea is that, when the branch is merged into master, there should be no trace of fixes/reworks/changes. Not having rework commits also makes debugging easier when bisecting through the commit history.

openrtx/src/protocols/APRS/HDLC.cpp
rxt1077: I'm not sure this makes sense for HDLC. I'm not as knowledgeable about M17, but it looks like it has distinct states in the demodulator, SYNCED, INIT.

HDLC is really just always counting ones and IF there are six of them it's a flag. If there's five of them and a zero you ignore it (it's been stuffed in there to avoid a flag), if there's seven, something went wrong. 

To implement a switch statement you would need some logic to set the state based on what bit was received and how many ones have been counted (basically what's in there already) and then you'd take actions. I'm not sure it would be any more readable than comments in the if statements.

openrtx/include/protocols/APRS/Demodulator.hpp
rxt1077: Convolver is performing the convolution operation in chunks of APRS_BUF_SIZE needed for the FIR filters and correlators. There's a lot of overlap with that Correlator class you wrote. I agree that having a common set of DSP functions as opposed to custom ones for each protocol would be the best/most maintainable route. It looks like an integer version of [FIR.hpp](https://github.com/OpenRTX/OpenRTX/blob/master/openrtx/include/core/fir.hpp) in core would be best.

I'll take a look at making that work in the coming week.

If it would result in a new core integer FIR class would you want that class also in core/fir.hpp or would you want a new file? Would you want it in this PR or in a separate one?

openrtx/src/protocols/APRS/HDLC.cpp
silseva: I see what you mean. We can keep it as it is.

openrtx/include/protocols/APRS/Demodulator.hpp
silseva: As we discussed yesterday, having a class for fixed-point FIR inside core/fir.hpp is fine. And also there is no need for a separate PR: just isolate the addition in a dedicated commit.

rxt1077: Just FYI, I wanted to test the FIR implementation to see if there was any benefit to doing it in blocks with overlap-add the way I implemented it, or doing it one at a time like OpenRTX currently has. Long story short, for an 81 tap FIR used to do the RRC calculation on a 1920 sample block (like you're currently using in the M17 code) there really isn't any noticeable benefit.

I put [my test](https://github.com/rxt1077/OpenRTX/blob/27fd914ff6975d93a2f90d81a251ef29c2e6227b/tests/unit/FIR.cpp) and [FIR implementation](https://github.com/rxt1077/OpenRTX/blob/27fd914ff6975d93a2f90d81a251ef29c2e6227b/openrtx/include/core/fir.hpp) in a separate branch, [fir-test](https://github.com/rxt1077/OpenRTX/tree/fir-test). You may be able to [eek out some more performance from that convolution in a similar fashion using an FFT](https://en.wikipedia.org/wiki/Overlap%E2%80%93add_method).

Other things to consider are whether that FIR code can be modified/made more generic and used in the correlation work you do for sync symbols. I think there's also some ARM-optimized convolution functions from CMSIS-DSP I saw reference is the Miosix kernel.

This is really a long winded way of saying I haven't forgotten about this and I think the best solution is to add a type to the Fir class template, update the M17 code to use the float type when setting up that class, and modify the APRS code to use the Fir class too, which I realize is exactly what you asked for!

rxt1077: OK, I've finished switching over to using the FIR code currently in OpenRTX. I refactored the way Demodulator, Slicer, and Decoder work to match the way Fir works, namely to operate on one sample at a time and use operator() since they're all basically a class that does one thing. This makes things simpler: less bit packing and unpacking, Decoder will now emit _one_ frame or nothing, no need for a vector of frames. The OpMode and test had to be refactored as well.

In the process I also fixed a missing test case (no packets decoded) and a memory leak in Decoder (currentFrame needs to be freed if the program ends before we get the final flag).

openrtx/src/rtx/OpMode_APRS.cpp
```diff
@@ -0,0 +1,119 @@
...
+        for (size_t i = 0; i < APRS_BUF_SIZE; i++) {
+            // apply the DC filter and scale the samples by 4 so our convolution
+            // operations don't overflow. This seems to work well with the baseband
+            // audio level
+            int16_t sample = dsp_dcBlockFilter(&dcBlock, baseband.data[i]) >> 2;
+
+            // demodulate
+            sample = demod(sample);
+
+            // slice
```
silseva: I'd move slicing and decoding inside the `demod` method of the demodulator.
This function, then, returns true if a new full packet is ready. For reference, see how the M17 demodulator is implemented

openrtx/src/rtx/OpMode_APRS.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+    audioStream_terminate(basebandId);
+    audioPath_release(rxAudioPath);
+
+    radio_disableRtx();
+
+    enterRx = true;
+}
+
+void OpMode_APRS::update(rtxStatus_t *const status, const bool newCfg)
```
silseva, line 55: Please move the RX logic to a dedicated function: this will allow to keep the code cleaner when TX will be added.
See: https://github.com/OpenRTX/OpenRTX/blob/778e5b6d4b2a01cc2bf39bba4a00b0237fbde6c8/openrtx/src/rtx/OpMode_M17.cpp#L97

openrtx/src/rtx/rtx.cpp
```diff
@@ -144,6 +151,9 @@ void rtx_task()
                 #ifdef CONFIG_M17
                 case OPMODE_M17:  currMode = &m17Mode; break;
                 #endif
+		#ifdef CONFIG_APRS
```
silseva: Broken indentation :)

openrtx/src/rtx/OpMode_APRS.cpp
silseva: > Do you also want APRS::Demodulator to have a readyFrame/demodFrame swapping memory management setup like M17Demodulator with a getter for the frame? You wouldn't have to remember to delete frames outside of the class.

Yes, please.
Having dynamic memory allocation is not a problem, but in embedded targets having a tight allocate-release cycle like in this case is not the best.

> Do you want dsp_dcBlockFilter in there too?

Why not? In this way we make the demodulator taking a chunk of samples and returning a block of bits or a frame.

> Would you be opposed to having another class (I don't know what to call it, context?) handle this so that all Demodulator does is demodulate, etc. The end result would still have more simplified code in the OpMode section.

I don't really know if it does make sense to introduce another class with the only purpose of is to wrap the demodulation, slicing and decoding steps. Given that the body of the operator() in the demodulator contains only a few lines of code, I think we can move everything there and have something which translates a block of baseband samples into a frame.
In the case of M17 we have the advantage that the frame size is fixed, so the demodulator does the baseband > bits conversion and then there is the frame decoder which does the rest of the protocol processing. I don't know if here we can do something similar

rxt1077: I'm going to make _slice_, _decode_, and _demod_ be member functions of _APRS::Demodulator_. Then I'll add an _update_ function that returns a boolean if a frame is ready. I'll also double buffer the frame (_readyFrame_, _demodFrame_) the same way _M17Demodulator_ does so that the class can handle all the memory management.

rxt1077: LOL, we must have been typing this at the same time. I realized the error of my ways after thinking about it for a bit and tried deleting my first response!

silseva: > I'm going to make _slice_, _decode_, and _demod_ be member functions of _APRS::Demodulator_. 

You can keep them as standalone classes and instantiate them inside the `Demodulator` class, actually.

openrtx/src/protocols/APRS/HDLC.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+bool Decoder::checkCRC()
+{
+    if (currentFrame->size() < 3)
+        return false;
+
+    // CRC16 via table lookup by nibble
+    // source: https://gist.github.com/andresv/4611897
+    uint16_t crc = 0xFFFF;
+    uint16_t crcTable[] = { 0x0000, 0x1081, 0x2102, 0x3183, 0x4204, 0x5285,
```
silseva: This table can be made `static constexpr`

openrtx/src/protocols/APRS/HDLC.cpp
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    crc = ~crc;
+
+    DEBUG_PRINT("CRC Data %02x %02x Calculated %02x %02x\n",
+                currentFrame->at(currentFrame->size() - 2),
+                currentFrame->at(currentFrame->size() - 1), (uint8_t)(crc >> 8),
+                (uint8_t)(crc & 0xff));
+
+    if ((currentFrame->at(currentFrame->size() - 2) == (crc & 0xff))
```
silseva: To improve readability is better to have something like:
```
uint16_t pktCrc = (currentFrame->at(currentFrame->size() - 1) << 8)
                  | (currentFrame->at(currentFrame->size() - 2));
if (pktCrc == crc)
    return true;
```

openrtx/src/protocols/APRS/HDLC.cpp
```diff
@@ -0,0 +1,147 @@
...
+        && (currentFrame->at(currentFrame->size() - 1) == ((crc >> 8) & 0xff)))
+        return true;
+    return false;
+}
+
+std::vector<uint8_t> *Decoder::operator()(const uint8_t bit)
```
silseva: Some of the comments in this function are quite obvious, I'd remove them

openrtx/src/protocols/APRS/packet.c
```diff
@@ -0,0 +1,164 @@
+/*
+ * SPDX-FileCopyrightText: Copyright 2020-2026 OpenRTX Contributors
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+#include "protocols/APRS/packet.h"
+#include "interfaces/platform.h"
+
+/**
```
silseva: Please move the documentation of functions to the header file

openrtx/src/protocols/APRS/Slicer.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+Slicer::Slicer()
+{
+}
+
+Slicer::~Slicer()
+{
+}
+
+int8_t Slicer::operator()(const int16_t input)
```
silseva: Also here comments can be removed, imo

openrtx/src/ui/default/ui.c
```diff
@@ -2673,6 +2705,11 @@ bool ui_updateGUI()
         case LOW_BAT:
             _ui_drawLowBatteryScreen();
             break;
+#ifdef CONFIG_APRS
+        case APRS_PKT:
+            ui_aprsPktDraw(&ui_state);
```
silseva: I'd call this function `ui_printAprsPkt`, `ui_showAprsPkt` or something similar: you do not really _draw_ the content of a packet

openrtx/include/protocols/APRS/packet.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include <string.h>
+#include <stdlib.h>
+#include "interfaces/platform.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct address {
```
silseva: Please change this to `struct address`, without the  `typedef`: is not yet explicitly stated in the coding conventions, but I'm moving all the codebase towards using typedefs only for data types which are not structs or enums.
See https://www.kernel.org/doc/html/v5.6/process/coding-style.html#typedefs

openrtx/include/protocols/APRS/packet.h
```diff
@@ -0,0 +1,42 @@
...
+    char addr[7]; /* include null termination */
```
silseva, line 20: I suggest to restructure the fields as follows:
```
struct address {
    struct address *next;
    char addr[7];
    uint8_t ssid : 4;
    uint8_t commandHeard : 1;
    uint8_t __unused : 3;
}; ```

In this way the fields are properly aligned and there is no extra-cost due to padding

File: openrtx/include/protocols/APRS/packet.h [hunk 17]
```diff
@@ -0,0 +1,42 @@
...
+    uint8_t ssid : 4;
+    bool commandHeard;
+    struct address *next;
+} aprsAddress_t;
+
+typedef struct packet {
+    aprsAddress_t *addresses;
```
silseva: How many addresses a packet can have? I'm expecting it to have at most two: source and destination.
If so, is better to replace the list with a fixed-size array and allocate all the `struct packet` in a single chunk.

openrtx/include/protocols/APRS/packet.h
```diff
@@ -0,0 +1,42 @@
...
+#endif
+
+typedef struct address {
+    char addr[7]; /* include null termination */
+    uint8_t ssid : 4;
+    bool commandHeard;
+    struct address *next;
+} aprsAddress_t;
+
+typedef struct packet {
```
silseva: I suggest to a have this struct defined like this:
```
struct packet {
    struct packet *prev;
    struct packet *next;
    struct address addresses[N];
    datetime_t ts;
    char info[0];
}
```
This allows to allocate the entire struct with `malloc(sizeof(struct packet) + info_size);`, reducing memory fragmentation and malloc overhead.

See also https://gcc.gnu.org/onlinedocs/gcc/Zero-Length.html

openrtx/include/protocols/APRS/packet.h
```diff
@@ -0,0 +1,42 @@
...
+    aprsAddress_t *addresses;
+    char *info;
+    datetime_t ts;
+    struct packet *prev;
+    struct packet *next;
+} aprsPacket_t;
+
+void aprsPktFree(aprsPacket_t *pkt);
```
silseva: These functions are dealing with packet lists, more than with single packets. Additionally, we have a packet list both in the radio state data structure and in the rtx status.
I suggest to do the following:
- create a new data structure representing a list of packets
```
struct aprsPktList {
    struct aprsPacket *packets;
    size_t nbPackets;
}
```
- rename the functions to be `aprsPktList_insert`, `aprsPktList_free`, `aprsPktList_delete`,...

Eventually, I'd split the file in two translation units: one named `packet.h/.c` which contains the definition of the `aprsPacket` struct and functions to do the packet <> raw frame conversion, and one named `packet_list.h/.c` which contains only the list implementation.

openrtx/src/core/state.c
```diff
@@ -72,10 +73,19 @@ void state_init()
...
+
+    state.aprsStoredPkts     = NULL;
+    state.aprsStoredPktsSize = 0;
 }
 
 void state_terminate()
 {
+#ifdef CONFIG_APRS
+    /* free APRS packets that have been stored */
+    for (aprsPacket_t *pkt = state.aprsStoredPkts; pkt; pkt = pkt->next)
```
silseva: These two lines of code can become a nice function inside `packet_list.h/.c`:
```
void aprsPktList_release(struct pktList *list)
{
    struct aprsPkt *head = list->packets;

    for (size_t i = 0; i < list->nbPackets; i++) {
         struct aprsPkt *next = head->next;
         free(head);
         head = next;
    }
}
```

openrtx/src/core/threads.c
```diff
@@ -59,6 +59,31 @@ void *ui_threadFunc(void *arg)
         }
 
         pthread_mutex_lock(&state_mutex);   // Lock r/w access to radio state
+
+#ifdef CONFIG_APRS
+        // copy APRS packets from RTX to radio state
+        if (state.channel.mode == OPMODE_APRS) {
+            rtxStatus_t rtx_status = rtx_getCurrentStatus();
+            if (rtx_status.aprsPkts) {
+                pthread_mutex_lock(&rtx_mutex); // Lock r/w access to RTX configuration
```
silseva: The critical section can be reduced to this:
```
pthread_mutex_lock(&rtx_mutex);
rtx_cfg.aprsPkts = NULL;
rtx_cfg.aprsPktsSize = 0;
pthread_mutex_unlock(&rtx_mutex);
```

As long as you access the list in read mode, is fine also without a mutex.

openrtx/src/rtx/OpMode_APRS.cpp
rxt1077: I actually think they all belong as member functions in _APRS::Demodulator_ and the _APRS::Slicer_ and _APRS::Decoder_ classes should be scrapped. Here's why:

1. _APRS::Demodulator_ will be handling the _readyFrame_ and _demodFrame_ pointers to store frames that go to OpMode_APRS. If I put an instance of _APRS::Decoder_ class inside _APRS::Demodulator_ I'd have to deal with moving or referencing those pointers from a class within a class. If I take out the _decode_ function (operator() in _APRS::Decoder_ right now) and put it directly in the _APRS::Demodulator_ class I can deal with it more directly and the net effect is OpMode_APRS still looks neater.
1. Eventually we'll need to pull signals from other parts of the whole process to notify OpMode_APRS of what's going on. For example, digital carrier detection is usually pulled from the state of clock recovery which right now is inside _APRS::Slicer_. If I move the _slice_ function (operator() in APRS::Slicer) into _APRS::Demodulator_ I can get at that signal more directly. The same holds true to signaling when a packet fails a CRC.

I realize the down side is that _APRS::Demodulator_ gets harder to look at, but if you take it member funtion-by-function it's not too bad. It does have to maintain variables for state across the demodulator, slicer, and decoder but all-in-all that's 9 variables or so.

silseva: Yes, good points.
I was thinking that it'd be good to keep the HDLC framing outside the demodulator. The rationale for this is that you can reuse it in case there are devices capable of handling the demodulating part in hardware outputting raw frames. Is similar to what we do for M17: the demodulator outputs raw frames and then there is the frame decoder that does the rest.
The demodulator can return a reference to a raw frame, and this is passed to the HDLC layer, avoiding unnecessary copies.
We can probably avoid the double buffering, and having the demodulator doing nothing until the frame is read and acknowledged by the HDLC: we risk losing some frames, but probably we can afford this.
What do you think?

rxt1077: I think having the _Demodulator_ spit out the data portion of an HDLC frame would be our best bet. Unfortunately HDLC requires bit unstuffing, HDLC flag removal, and LSB to MSB conversion just to check the Frame Check Sequence (CRC). You pretty much have to decode it to know if you've received a valid frame.

The good news is this gives you some flexibility. You could take that data and wrap it in a KISS frame to make a TNC or you could interpret it as an APRS/AX.25 packet.

We'll need double buffering if _Demodulator_ operates in chunks since you could close out one frame and start the next within the chunk you're processing.

Having tried the monolithic class technique I originally proposed, I'm now of the opinion that we should still have a _Decoder_ and _Slicer_ class within _Demodulator_ (like you proposed). It's just too hard to look at otherwise.

I've rewritten the _Decoder_ to make it easier to understand, based on some of your earlier comments regarding using an FSM. I now break things down into states acting on signals from the bit counter. I think it reads easier and addresses some of the obvious comments issues.

I should finish this refactor soon.

silseva: > I think having the _Demodulator_ spit out the data portion of an HDLC frame would be our best bet. Unfortunately HDLC requires bit unstuffing, HDLC flag removal, and LSB to MSB conversion just to check the Frame Check Sequence (CRC). You pretty much have to decode it to know if you've received a valid frame.

However this means that the demodulator has to do everything, comprised bit unstuffing, endianness conversion and CRC check. Or am I missing something?
What I proposed to do is to have the demodulator spitting out raw HDLC frames, that is everything between the two 0x7E markers, and then have a separate module that does the unstuffing and CRC validation. This would be similar to the M17 implementation, where we have the demodulator (which does not care about the content of the frames, except for the syncword) and then the frame decoder.

rxt1077: So to be specific with the terminology the modulation is AFSK (keying being a special form of modulation) on top of FM (also modulation). A chip on the radio already handles the FM demodulation for us and gives us base band audio. If we are demodulating AFSK you'd usually ask for discriminator output, or the difference between tones. This is what _APRS::Demodulator_ as originally written gives you.

To get up to HDLC you need to recover bit timing and make bit decisions this is what _APRS::Slicer_ does. It's still just bits, no bytes yet. Finally you can decode it, _APRS::Decoder_, to get HDLC flags, unstuff bits, check CRCs, etc.

Unfortunately the term demodulation is a little overloaded and sometimes it's used to mean recovering frames from something that is digitally modulated.

So we could put those steps in some overarching class: Demodulation -> Slicing -> Decoding, technically with a little signal conditioning before the Demodulation step (removing DC). I originally put those steps in OpMode_APRS (which made it a little busy).

If you really want an HDLC frame, it would come out of an intermediary step in _APRS::Decoder_ where you've detected a frame by counting bits, but you didn't unstuff bits, correct endianness, or check the CRC.

openrtx/src/protocols/APRS/HDLC.cpp
rxt1077: I ended up implementing this the same way crc_ccitt() is done. It should be faster and save a few bytes.

openrtx/src/protocols/APRS/HDLC.cpp
rxt1077: It now uses a core/crc.c function.

openrtx/src/protocols/APRS/HDLC.cpp
rxt1077: This Decoder was significantly refactored to make it easier to follow with fewer comments.

openrtx/src/protocols/APRS/Slicer.cpp
rxt1077: Simplified the comments and introduced DCD for eventual TX usage

openrtx/include/protocols/APRS/packet.h
rxt1077: An APRS packet will have a src, dst, and up to 8 additional digipeater addresses. So somewhere between 2-10.

openrtx/include/protocols/APRS/packet.h
rxt1077: The zero length array is not recommended by the documentation you provided, "The preferred mechanism to declare variable-length types like struct line above is the ISO C99 flexible array member, with slightly different syntax and semantics:", but it sounds like we actually have to use it because it supports sizeof and info[] does not?

I guess the point is to make it clear that it is an array of uint8_t's and not just a pointer to some random uint8_t AND to make it so we only need one malloc call, got it.

openrtx/src/rtx/OpMode_APRS.cpp
silseva: Sounds good!

openrtx/include/protocols/APRS/packet.h
silseva: Using the flexible array member syntax instead of `char info[0]` is totally fine, I had just forgot about its existence :)
I forgot to add a field for data size, so it'd be
```
struct packet {
    struct packet *prev;
    struct packet *next;
    struct address addresses[N];
    datetime_t ts;
    size_t payloadSize;
    uint8_t payload[];
}
```
In addition to the single malloc, you can access the payload with `packet->payload[i]`

openrtx/include/protocols/APRS/packet.h
silseva: Ok. Having a fixed array of 10 `struct address` is not the best, then, because it will take 120 bytes even when only two of them are used.
I'd have something like this, then:
```
struct packet {
    struct packet *prev;
    struct packet *next;
    struct address source;
    struct address destination;
    struct address *hops;
    uint8_t nbHops;
    datetime_t ts;
    size_t payloadSize;
    uint8_t payload[];
}
```
and allocate dynamically `nbHops * sizeof(struct address)` when necessary.
Where are the digipeater addresses stored, in the APRS frame?

rxt1077: Since we can calculate how many addresses a packet has and the size of the info field _before_ we allocate things, I ended up using two pointers so there's one malloc and both fields act like arrays. Best of both worlds if you can stomach the pointer math.

openrtx/include/protocols/APRS/packet.h
rxt1077: Great point. I implemented this and built a test to check my work.

openrtx/include/protocols/APRS/packet.h
silseva: Looks good!

openrtx/include/protocols/APRS/Decoder.hpp
```diff
@@ -0,0 +1,130 @@
...
+
+#ifndef APRS_DECODER_H
+#define APRS_DECODER_H
+
+#ifndef __cplusplus
+#error This header is C++ only!
+#endif
+
+#include <cstdint>
+//#include <cstddef>
```
silseva: This can be removed, I think :)

openrtx/include/protocols/APRS/Decoder.hpp
```diff
@@ -0,0 +1,130 @@
...
+     * Performs non-return-to-zero, inverted decoding and stores in input bit
+     * for the next comparison
+     *
+     * @param bit A uint8_t bit from the slicer
+     *
+     * @return A uint8_t NRZI decoded bit
+     */
+    uint8_t nrziDecode(const uint8_t bit);
+
+    enum class Signal { FLAG = 'F', ERROR = 'E', NONE = 'N', IGNORE = 'I' };
```
silseva: Do you mind moving this to line 117, please?
I tend to have a convention, for private class space ordering:
- functions
- constants and enums
- variables

openrtx/include/protocols/APRS/Decoder.hpp
```diff
@@ -0,0 +1,130 @@
...
+    void dumpFrame();
+
+    uint8_t onesCount = 0;                        ///< count of one bits recv
+    uint8_t prevBit = 0;                          ///< previous bit
+    uint8_t workingByte = 0;                      ///< byte being built
+    uint8_t bitPos = 0;                           ///< position in workingByte
+    std::unique_ptr<struct frameData> demodFrame; ///< frame being built
+    std::unique_ptr<struct frameData> readyFrame; ///< fully built frame
+    bool newFrame = false;                        ///< new frame available flag
+    enum class State { WAITING = 'W', INFRAME = 'I' };
```
silseva: Same of `enum class Signal`

openrtx/include/protocols/APRS/constants.h
```diff
@@ -0,0 +1,24 @@
...
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+#ifndef APRS_CONSTANTS_H
+#define APRS_CONSTANTS_H
+
+#define APRS_BUF_SIZE 64
+#define APRS_SAMPLE_RATE 9600
+#define APRS_SYMBOL_RATE 1200
+#define APRS_PACLEN 128
```
silseva: I'd move this to `frame.h`, for consistency

openrtx/include/protocols/APRS/frame.h
```diff
@@ -0,0 +1,25 @@
...
+#include "protocols/APRS/constants.h"
+
+#ifndef APRS_FRAME_H
+#define APRS_FRAME_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+struct frameData {
```
silseva, line 18: Is this a raw HDLC frame? If so, I'd call it `struct hdlcFrame` or something similar for better clarity

openrtx/include/protocols/APRS/packet.h
```diff
@@ -0,0 +1,48 @@
...
+#include "protocols/APRS/frame.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+struct aprsAddress {
+    char addr[7]; /* include null termination */
+    uint8_t ssid      : 4;
+    bool commandHeard : 1;
```
silseva:
```suggestion
    uint8_t commandHeard : 1;
```
Otherwise the compiler does not recognize the bitfield

openrtx/include/protocols/APRS/packet.h
```diff
@@ -0,0 +1,48 @@
...
+};
+
+struct aprsPacket {
+    struct aprsPacket *prev;
+    struct aprsPacket *next;
+    datetime_t ts;
+    uint8_t addressesLen;
+    uint8_t infoLen;
+    struct aprsAddress *addresses;
+    char *info;
```
silseva, line 33:
```suggestion
    uint8_t *payload;
```
maybe?

openrtx/include/protocols/APRS/packet.h
```diff
@@ -0,0 +1,48 @@
...
+    uint8_t __unused  : 3;
+};
+
+struct aprsPacket {
+    struct aprsPacket *prev;
+    struct aprsPacket *next;
+    datetime_t ts;
+    uint8_t addressesLen;
+    uint8_t infoLen;
+    struct aprsAddress *addresses;
```
silseva, line 32: Please move those two pointers up, immediately below `struct aprsPacket *next;`: this keeps the struct content aligned to 4 bytes as much as possible and saves space (otherwise there is extra, hidden, padding to align the fields to the machine word size)

openrtx/src/core/state.c
```diff
@@ -73,10 +74,16 @@ void state_init()
     {
         state.settings.brightness = 100;
     }
+
+    state.aprsStoredPkts = aprsPktList_init();
```
silseva:
```suggestion
#ifdef CONFIG_APRS
    state.aprsStoredPkts = aprsPktList_init();
#endif
```

openrtx/include/protocols/APRS/packet_list.h
```diff
@@ -0,0 +1,78 @@
...
+ */
+struct aprsPktList aprsPktList_delete(struct aprsPktList list,
+                                      struct aprsPacket *pkt);
+
+/**
+ * @brief Returns an empty, initialized APRS packet list
+ *
+ * @return An empty, initialized struct aprsPktList
+ */
+struct aprsPktList aprsPktList_init();
```
silseva:
```suggestion
void aprsPktList_init(struct aprsPktList *list);
```
Doing so we avoid the hidden copy due to the return-by-value.

openrtx/src/rtx/OpMode_APRS.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+void OpMode_APRS::offState(rtxStatus_t *const status)
+{
+    if (enterRx) {
+        radio_disableRtx();
+
+        radio_enableRx();
+        status->opStatus = RX;
+        rxAudioPath = audioPath_request(SOURCE_RTX, SINK_MCU, PRIO_RX);
+        //TODO: should we throw an error if rxAudioPath doesn't work?
```
silseva: Response to the TODO: no error, but I'd keel the `enterRx` flag to true and retry after 30ms.
Having 
```
if (basebandId > 0) {
   enterRx = false;
   return;
}
```
should suffice

openrtx/src/rtx/rtx.cpp
```diff
@@ -53,6 +58,8 @@ void rtx_init(pthread_mutex_t *m)
     rtxStatus.M17_link[0]   = '\0';
     rtxStatus.M17_refl[0]   = '\0';
     rtxStatus.M17_meta_text[0] = '\0';
+    rtxStatus.aprsRecv      = 0;
+    rtxStatus.aprsPkts      = aprsPktList_init();
```
silseva:
```suggestion
#ifdef CONFIG_APRS
    rtxStatus.aprsRecv  = 0;
    rtxStatus.aprsPkts = aprsPktList_init();
#endif
```

openrtx/include/protocols/APRS/Decoder.hpp
rxt1077: Can do, but it will require a forward declaration to compile. I'll put the forward declaration right before the countBits() prototype where it's used unless you want it somewhere else.

openrtx/include/protocols/APRS/frame.h
rxt1077: This isn't a raw HDLC frame. As far as I can tell there's a couple of issues with those that cause HDLC and AX.25 to be hopelessly intertwined in most implementations. This is just the data that is inside the HDLC frame which we already had to decode in the process.

In case you're interested, the HDLC/AX.25 issues that cause them to be linked are:

1. Frames are separated by one or more flags, but you don't know they're valid until you check the frame check sequence (CRC). You get a lot of spurious flags just listening to noise.
2. Checking the FCS (CRC) requires unstuffing the bits. Which requires working with the incoming signal bit by bit.
3. Watching for flags requires working with the incoming signal bit by bit.

It would be great if we could just have a decent indicator that we've received a flag and ship those bytes off to another layer for processing once one comes in. Unfortunately we have to do 90% of the work of the next layer just check to see if we've received a valid frame.

openrtx/include/protocols/APRS/packet.h
rxt1077: Believe, me I agree with you that info is a terrible name for this field and it would be better if the data type was uint8_t, but [per the current APRS spec (page 12)](https://raw.githubusercontent.com/wb2osz/aprsspec/main/APRS12c.pdf) this is the "INFORMATION FIELD" and it contains _characters_ (in fact it's limited to 91 of the printable ASCII characters).

openrtx/include/protocols/APRS/packet.h
rxt1077: `addresses` and `info` are at the end of the struct because they are each variable length arrays. Their size and pointers are determined dynamically and [set in packet.c](https://github.com/OpenRTX/OpenRTX/pull/396/changes#diff-ec6f0656a6b182a7eb98a9804dec9066ed5271281b6fbea33c1f42ffa8b4c074R59).

openrtx/include/protocols/APRS/packet.h
silseva: Gotcha, let’s leave it as it is. :smile:

openrtx/include/protocols/APRS/packet.h
silseva: True. Sorry, I forgot this detail.

openrtx/src/rtx/rtx.cpp
rxt1077: Just to be clear, rtxStatus_t (rtx/rtx.h) should be the same for all supported hardware but we use CONFIG flags to avoid initializing some attributes in rtxStatus_t for hardware that doesn't support the mode?

silseva: We should have `CONFIG`s also in the rtx status data structure, to follow the principle "you should not pay for what you do not use". Currently this is not the case, though. You can skip this review.

rxt1077: As long as there's nothing fundamentally wrong with having different hardware having different rtxStatus's I'm happy to implement it. I think the "you should not pay for what you do not use" mantra makes a lot of sense. My only concern was if rtxStatus was used to save the radio state in a file somewhere that multiple radios use, like a code plug or something. Barring that, I'd like to follow your original advice.

openrtx/include/protocols/APRS/constants.h
rxt1077: I agree, even though it _technically_ it's not just the size of `frameData.data` but a setting used by the `Decoder`.

It's also a terrible variable name because we're not really dealing with packets at this point, but it's historically an APRS _setting_ with that name. At some point we're going to need to decide exactly what APRS settings we want to support. PACLEN, TXDELAY, SLOTTIME, PERSIST, TXTAIL all come to mind with their byzantine meanings and names.

openrtx/src/rtx/rtx.cpp
rxt1077: Never mind, I see that what we really need is for the OpMode_APRS files to be pulled in by the build system IF the hardware supports it. That doesn't seem to be how it's currently set up. When I try cutting out the APRS parts of the rtx status structure OpMode_APRS can't compile.

openrtx/include/protocols/APRS/frame.h
silseva: I see what you mean, now.

silseva (approved):

PR #397, opened 2026-02-03
Implement an nvm driver for STM32 internal flash

mdiepart:
This PR implements an NVM driver for the internal STM32 flash storage.

The branch I work is the nvm-updates branch rebased on the master branch at the time of opening the draft PR.

### DIFF
--- a/openrtx/include/core/nvmem_access.h
+++ b/openrtx/include/core/nvmem_access.h
@@ -42,42 +42,43 @@ int nvm_getPart(const uint32_t idx, const uint32_t part,
  * This function allows to read data from an NVM device starting at a given
  * offset from the beginning of the device.
  *
- * @param dev: NVM device number.
- * @param part: partition number, -1 for direct device access.
+ * @param idx: index of the nonvolatile memory area.
+ * @param part: partition number.
  * @param offset: offset for the read operation.
  * @param data: pointer to a buffer where to store the data read.
  * @param len: number of bytes to read.
  * @return zero on success, a negative error code otherwise.
  */
-int nvm_read(const uint32_t dev, const int part, uint32_t offset, void *data,
-             size_t len);
+int nvm_read(const uint32_t idx, const uint32_t part, uint32_t offset,
+             void *data, size_t len);
 
 /**
  * Perform a write operation on a nonvolatile memory.
  * This function allows to write data to an NVM device starting at a given
  * offset from the beginning of the device.
  *
- * @param dev: NVM device number.
- * @param part: partition number, -1 for direct device access.
+ * @param idx: index of the nonvolatile memory area.
+ * @param part: partition number.
  * @param offset: offset for the write operation.
  * @param data: pointer to a buffer containing the data to write.
  * @param len: number of bytes to write.
  * @return zero on success, a negative error code otherwise.
  */
-int nvm_write(const uint32_t dev, const int part, uint32_t offset,
+int nvm_write(const uint32_t idx, const uint32_t part, uint32_t offset,
               const void *data, size_t len);
 
 /**
  * Perform an erase operation on a nonvolatile memory. Acceptable offset and
  * size depend on characteristics of the underlying device.
  *
- * @param dev: NVM device number.
- * @param part: partition number, -1 for direct device access.
+ * @param idx: index of the nonvolatile memory area.
+ * @param part: partition number.
  * @param offset: offset for the erase operation.
  * @param size: size of the area to be erased.
  * @return zero on success, a negative error code otherwise.
  */
-int nvm_erase(const uint32_t dev, const int part, uint32_t offset, size_t size);
+int nvm_erase(const uint32_t idx, const uint32_t part, uint32_t offset,
+              size_t size);
 
 #ifdef __cplusplus
 }
--- a/openrtx/src/core/nvmem_access.c
+++ b/openrtx/src/core/nvmem_access.c
@@ -6,49 +6,8 @@
 
 #include "core/nvmem_access.h"
 #include "core/nvmem_device.h"
-#include <stdbool.h>
 #include <stdint.h>
 #include <errno.h>
-
-/**
- * \internal
- * Compute the absolute offset from the beginning of an NVM device, given the
- * device descriptor and the partition number. If the partition number is set
- * to -1, the offset is considered from the beginning of the device.
- * This function performs also a bound check to guarantee that the following
- * operation stays within the limits of the partition (if any).
- *
- * @param nvm: pointer to NVM descriptor.
- * @param part: partition number.
- * @param offset: pointer to the offset.
- * @param len: lenght of the read/write/erase operation.
- * @return a negative error code or zero.
- */
-static inline int getAbsOffset(const struct nvmDescriptor *nvm, const int part,
-                               uint32_t *offset, size_t len)
-{
-    const struct nvmPartition *np;
-    uint32_t offs = *offset;
-
-    // Offset is relative to a partition
-    if (part >= 0) {
-        if ((size_t)part >= nvm->nbPart)
-            return -EINVAL;
-
-        np = &nvm->partitions[part];
-
-        if ((offs + len) > np->size)
-            return -EINVAL;
-
-        offs += np->offset;
-    }
-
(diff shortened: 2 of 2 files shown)


turnrye:
@silseva is there a follow-up needed to do similar work for the MK22FN512?

silseva:
Not in the immediate future, because is not really necessary. This one is needed because some devices, like the Module17, are using some sectors of the MCU flash to store settings and device persistence.

mdiepart:
Hey!
So just to be sure what is needed here. 
I'm implementing a driver to use the flash for the STM32 as a full-fledge NVM driver that can be used for raw storage but that could also be used in an EEEPROM driver.

I'm also looking at the various structures that are defined. The info struct has a erase_cycles field but what is the intended use for this? I am having a hard time figuring how any modification to this value will persist. Or is it just supposed to be the value within the datasheet and will not be modified?

Also regarding the erase_size field, this will prevent having a single driver with variable erase size depending on the flash pages. What we could do is 
1) Not use the value for flash and use the minimal page size
2) Have different info structs depending on the page size and constraint the driver to specific pages depending on their size (i.e. can only operate on 4k pages, OR can only operate on 16k pages)
3) Modify the way nvm drivers are implemented?

silseva:
> I'm also looking at the various structures that are defined. The info struct has a erase_cycles field but what is the intended use for this? I am having a hard time figuring how any modification to this value will persist. Or is it just supposed to be the value within the datasheet and will not be modified?

The `erase_cycles` field is meant to contain the maximum life of a flash sector, in terms of how many erases it can be subjected to. I introduced it with the last revision of the NVM API because LittleFS needs it for its wear leveling algorithm. In general, it can be useful whenever there is need to know which is the life limit of flash sectors.

> Also regarding the erase_size field, this will prevent having a single driver with variable erase size depending on the flash pages. What we could do is
> 
>     1. Not use the value for flash and use the minimal page size
> 
>     2. Have different info structs depending on the page size and constraint the driver to specific pages depending on their size (i.e. can only operate on 4k pages, OR can only operate on 16k pages)
> 
>     3. Modify the way nvm drivers are implemented?

As far as I know, STM32 are the only devices with variable flash page size. To add more fun, in newer devices (STM32G4, STM32U5, ...) the flash page size is fixed and (circa) unique.
I don't really know if it does make sense to add a `size_t (*page_size)(const struct nvmDevice *dev, uint32_t address);` function to the device driver API returning the size of the page a given address belongs to: in the majority of the cases it will just return a constant.
Another option is to have a `nvm_getPageSize` function in the high-level API, and tread the `erase_size` field either as a number or as a pointer to a list of tuples describing the flash configuration. To distinguish between one or the other, we add a `NVM_VARIABLE_PAGE` flag in the device attributes (`device_info` field).
To give an idea of what I have in mind, see Zephyr's a "flash page layout" [data structure and function](https://github.com/zephyrproject-rtos/zephyr/blob/72c43541ec906bc649145861fce21cbd2e90720a/drivers/flash/flash_stm32f2x.c#L169).
I think we need to take a step back and define a bit more what is the general plan, and from there derive the functionalities we need in the low level API. For example, at the moment I have in mind only two targets which will use the MCU flash for persistence and settings: Module17 and Micro17.

mdiepart:
I actually went the 2nd route. STM32H743xx has only 128k page size. And for the STM32F405, you have 4 pages of 16k and 1 page of 64k. The rest is 128k. So we will most probably only use 128k pages for flash storage. And probably will not need a driver that spreads across pages of different sizes.

I implemented a lot of sanity checks to make sure that everything is consistent, the driver must be given an address aligned on a page boundary and length that is an integer multiple of a page size.

Commits incoming with work in progress, implementing what I described.

mdiepart:
I think the majority of the driver is done. A few additional things need to be taken care of:
1) STM32H743 do not have a flash driver yet (flash.c/flash.h)
2) I need to make sure that I can read the flash size the way I did it (it seems that prefetch may need to be disabled). Reading flash size should also be moved to the flash driver.
4) I may switch to defining the flash layout the same way it is done in the zephyr flash driver.
3) Nothing is tested yet.

Regarding what you said about taking a step back, I don't think it is needed. Partly because if we keep refactoring stuff on and on, then nothing will get done. I think I have a good compromise here that will get us through what we need.

mdiepart:
Latest commits implemented a flash driver for STM32H743. This device is a bit more tricky as it has 2 memory banks. This device also has memory ECC which hinders write operations of less than 256 bits (32 bytes). As such the driver refuses to write less than 32 bytes at once (a block which must itself be 32 bytes aligned). 

I also fixed a few minor issues in the NVM driver. Points 2-4 of my previous reply remains unchanged.

mdiepart:
I started writing platform tests for the STM32Flash NVM driver. In the process, I already found and fixed quite a few issues.

silseva:
Since you're fixing things around, I suggest to use [fixup commits](https://moodledev.io/general/development/abc/fixups) so that the history remains clean (you'll see only what it has been done and not also the various fixes/reworks).

mdiepart:
Good point I intended to rebase the branch on itself at some point anyway to merge some commits

mdiepart:
I force pushed my commits (I re-organised them). I also changed the structure of the driver, so it made no sense keeping the whole history.
I added tests for the CS7000p but was not able to test them yet.
I'll do that later but I think this is ready for review already.

mdiepart:
The tests for the STM32H743 did not go as expected... 
I first found a few errors in the code that I fixed. Then I started chasing a programming sequence error that appeared during the second call to nvm_devRead(). This issue seemed fixed but power-cycling the radio shows that this error still appears every 4th test... I'm out of idea and don't really understand why this error appears randomly...

silseva:
I had only a general look at the code because all my commits on the `nvm-update` branch are adding "noise": I'll try to bring those changes into master as soon as possible, so that we can reduce the changes in the diff view of this PR.
This said, I have a couple of notes:
- I see that you implemented a lot of bound checking: some of the checks are already implemented by the NVM access function (see nvmem_access.h/.c) and others should actually be moved into the general NVM access API. I was thinking that checking for addresses being below the flash start address or above the flash end address is not strictly necessary, since any out-of-bounds access would anyways cause an hardware exception (busfault or whatever it is).
- for the `nvm_read` API I think we can just use `memcpy`: if you look at the [sources](https://github.com/eblot/newlib/blob/master/newlib/libc/string/memcpy.c) you can see that it is already implemented for efficient copy, more or less like you did.

We also changed the code style conventions quite recently: if you use `clang-format` you can get the files formatted according to it.

mdiepart:
I slightly reworked the nvmem_access entity to add some bounds checking. I removed the alignment check from the nvm device. I think however most of the bounds checking should stay in the nvm device because we have to prevent having request writing to existing memory but not covered by the page size of the current driver. 

I think the alignment check in the flash driver must stay as the operation should really fail if a non-aligned write/erase is attempted. 

Finally, I modified the STM32H743 flash driver to use memcpy. It seems to have made the programming sequence error worse and can't make it work anymore. I don't have any clue as to what is happening. This driver however is not in use in OpenRTX and will probably not be. So maybe you can check it and we can explicitely mark it as untested with a #warning directive?

silseva:
> I slightly reworked the nvmem_access entity to add some bounds checking. I removed the alignment check from the nvm device. I think however most of the bounds checking should stay in the nvm device because we have to prevent having request writing to existing memory but not covered by the page size of the current driver.
> 
> I think the alignment check in the flash driver must stay as the operation should really fail if a non-aligned write/erase is attempted.

In principle the application code should not call the low-level NVM driver APIs directly, and both these errors should (and must) be checked by the "userspace" NVM API made available in `nvmem_access.h`. But anyway, we can keep the additional checks.

> Finally, I modified the STM32H743 flash driver to use memcpy. It seems to have made the programming sequence error worse and can't make it work anymore. I don't have any clue as to what is happening. This driver however is not in use in OpenRTX and will probably not be. So maybe you can check it and we can explicitely mark it as untested with a #warning directive?

My idea was to use `memcpy` only for read operations, actually.

mdiepart:
I modified the read call of the NVM device to use memcpy, good catch.

mdiepart:
To be continued in #403

PR #398, opened 2026-02-07
Improve unit testing tooling

turnrye:
Generally, tests are some of the less exciting but necessary things that we deal with. This change hopes to make things a little better in how we use unit tests on OpenRTX. This should make writing and running unit tests simpler, give us a better understanding of what does or doesn't have unit tests, and addresses reviving a few unit tests that were previously abandoned or being skipped.

# Key Changes

* Completes the clang format adoption for unit tests
* Adopts catch2 as a testing framework
* Migrates all tests to C++
* Fixes the M17 demodulator and standby check tests
* Adds gcovr for collecting coverage
* Improves CI/CD integration
  * All tests are run now
  * Coverage is reported in the [build summary view](https://github.com/OpenRTX/OpenRTX/actions/runs/22123368693/attempts/1#summary-63948168329))
  * Test failures are reported with code annotations so that they're slightly easier to discover

# Usage

Run all unit tests and include verbose error logging if a test fails:
```sh
meson test -C build_linux --print-errorlogs
```

Here's the example output if a test is forced to fail, showing much more information about the failure compared to before:

```
―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― ✀  ――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
Randomness seeded to: 1034015665

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
m17_viterbi_test is a Catch2 v3.4.0 host application.
Run with -? for options

-------------------------------------------------------------------------------
Viterbi decode recovers punctured convolutional encoding
-------------------------------------------------------------------------------
../tests/unit/M17_viterbi.cpp:35
...............................................................................

../tests/unit/M17_viterbi.cpp:65: FAILED:
  REQUIRE( source[i] != result[i] )
with expansion:
  0 != 0
with message:
  Position 0: got 0x0 expected 0x0

===============================================================================
test cases: 1 | 1 failed
assertions: 1 | 1 failed

radio_linux: disableRtx() called
――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
```

You can also run only tests with a specific label, for instance:
```sh
 meson test -C build_linux --test-args '[m17] --allow-running-no-tests'
 ```

### CODE REVIEW

File: meson.build [hunk 1]
```diff
@@ -1074,10 +1079,12 @@ linux = alias_target('linux', linux_targets)
 ## ----------------------------------- Unit Tests ------------------------------
 ##
 
+if not meson.is_cross_build()
```
silseva: I think we can omit this; in the case where someone attempts to build and flash the unit tests for an embedded target, (s)he well surely get a compilation error.

tests/unit/ui_check_standby.cpp
```diff
@@ -0,0 +1,102 @@
...
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+#include <catch2/catch_test_macros.hpp>
+
+extern "C" {
+#include "core/state.h"
+#include "core/settings.h"
+
+bool _ui_checkStandby(long long);
```
silseva: I'd change this to `extern bool _ui_checkStandby(long long);` to further point out that this symbol belongs to another translation unit.

meson.build
turnrye: Fixed!

tests/unit/ui_check_standby.cpp
turnrye: Done! Btw, I looked to see if a tool like cppcheck or clangtidy could help with this rule, and apparently not.

silseva: Is not possible: a line like `bool _ui_checkStandby(long long);` can either be a forward declaration for a function defined in the same translation unit or for a function in another translation unit. To be able to determine whether it is `extern` or not, you would have to look at the compiled object file...

PR #401, opened 2026-02-13
ci: close stale issues and PRs

turnrye:
This change adds a github action in order to warn users after 30 days of inactivity that their PRs and issues are subject to being closed if they are not updated within the next 5 days. Labels are added to auto-closed issues so that analysis is easier, and issues with the labels bug or issue are skipped from this.

I chose not to do this for the enhancement type issues, as the longer term product backlog is being tracked in Tiaga, not GitHub Issues. Users are still welcome to log issues for enhancements, but this will encourage them to keep these discussions alive and work to a conclusion rather than having the longer-term roadmap split across two places.

### DIFF
--- a/.github/workflows/close-stale.yml
+++ b/.github/workflows/close-stale.yml
@@ -1,0 +1,20 @@
+name: 'Close stale issues and PRs'
+on:
+  schedule:
+    - cron: '30 1 * * *'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v10
+        with:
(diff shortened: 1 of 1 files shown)


turnrye:
Here is an example run of this automation on my fork, marking a very old PR as stale: https://github.com/turnrye/OpenRTX/pull/2#issuecomment-3900092441

Note that this is setup to run on a cron once, so it isn't expected that immediately following landing this change the automation will run.

silseva:
I'd elongate the `days-before-close` period to 15 days. Is quite a long time, but some people may not be able to react within five days.

turnrye:
> I'd elongate the `days-before-close` period to 15 days. Is quite a long time, but some people may not be able to react within five days.

Sounds good, I've changed it to 15d.

PR #403, opened 2026-02-24
Implement an nvmDevice for STM32 internal flash

mdiepart:
This is a continuation of #397 coming from a different source branch mainly.

The current state is:
The nvmDevice driver is implemented for both STM32F405  and STM32H743.
Tests are written for both STM32F405  and STM32H743.
The tests runs smoothly for STM32F405.
The tests often fails for STM32H743. There is a programming sequence error when writing to flash. I don't know why.

I think we can either discard the driver for the STM32H743 or (preferably) mark the STM32H743 driver as faulty with a compilation warning/error.


mdiepart:
These last commits solve the issue I've been having with the STM32H743. Driver is now fully functional.

silseva:
I merged the `nvm-updates` branch into master. Do you mind rebasing this branch on top of master and force-pushing it?

mdiepart:
@silseva it looks like I can't request a review. Would you mind reviewing this code so that we can get this merged?

silseva:
I didn't forget :)

silseva:
I had a look at the STM32H7 datasheet: the bitmap of `CR1` and `CR2` is the same, except for one bit which we're not using and `SR1` is identical to `SR2`. We can exploit this to factor out common code like the functions unlocking the banks and performing the erase.
You can have something like
```
static inline bool unlock_bank(struct flash_cr_typedef *cr)
{
    // Flash already unlocked
    if ((*cr & FLASH_CR_LOCK) == 0) {
        return true;
    }

    FLASH->KEYR1 = 0x45670123;
    FLASH->KEYR1 = 0xCDEF89AB;
    __ISB();

    // Succesful unlock
    if ((*cr & FLASH_CR_LOCK) == 0) {
        return true;
    }

    return false;
}
```
and then call it with `unlock_bank(&FLASH->CR1)` and `unlock_bank(&FLASH->CR2)`

mdiepart:
We would need to pass both the KEYR and CR registers as arguments. Also, the offset between KEYR1 and CR1 is not the same as the offset between KER2 and CR2. At this point, I don't think it helps in making the code any more readable...

Roughly the same goes for the erase function. I would then need to pass CR, SR and maybe KEYR if the unlock function is to be handled in that function...

silseva:
We can use C++ template functions:
```
template < CR, KEYR > unlock() ...
```

mdiepart:
I'll try it this afternoon but I don't think this is worth the trouble to switch to cpp for these files and use templates for what is, in the end, only two possibilities...
What I can do to improve code readability is reduce the compile time directives and move the flash layout and page size defined and so on to the device specific driver.

silseva:
>What I can do to improve code readability is reduce the compile time directives and move the flash layout and page size defined and so on to the device specific driver.

I think we were already discussing this: I'd move all the code in `NVM/flash_stm32.c` and remove the sources under `mcu/STM32F4xx`. To handle the differences between the STM32F4 and  STM32H7 we can have a single header file `flash_stm32.h` and two different source files, `flash_stm32f4.c` and `flash_stm32h7.c` for the two architectures. I did this for the ADC: https://github.com/OpenRTX/OpenRTX/tree/master/platform/drivers/ADC

platform/drivers/NVM/flash_stm32.h
```diff
@@ -0,0 +1,91 @@
...
+extern "C" {
+#endif
+
+// Currently, only STM32F405 and STM32H743 are used. STM32F405 can have a variable flash size.
+// STM32F743/753 is derived in STM32F743xI (2MB size) and STM32F743xG (1MB size).
+
+/**
+ * Flash page size enum
+ */
+enum STM32_Sector_Size_e {
```
silseva: Perhaps we can have a `CONFIG_SECTOR_SIZE`, only for the STM32F4, instead of the enum? It gives less flexibility, though, because it does not allow to declare two different STM32 flash devices with different sector size in the same program. But... do we actually need to declare more than one of them?
Opinions welcome

platform/drivers/NVM/flash_stm32.c
```diff
@@ -0,0 +1,174 @@
...
+        .address_low = 0x8000000,
+        .address_high = 0x81FFFFF,
+        .first_sector = 0,
+    },
+#else
+#error "Missing flash memory layout for this device."
+#endif
+};
+
+int STM32Flash_init(struct nvmDevice *dev)
```
silseva: We can omit these, actually.

platform/drivers/NVM/flash_stm32.h
mdiepart: Code is there and very unlikely to cause problems, my opinion is we should keep it :)

platform/drivers/NVM/flash_stm32.c
mdiepart: Not sure what you are referring to?

mdiepart:
> I think we were already discussing this: I'd move all the code in `NVM/flash_stm32.c` and remove the sources under `mcu/STM32F4xx`. To handle the differences between the STM32F4 and STM32H7 we can have a single header file `flash_stm32.h` and two different source files, `flash_stm32f4.c` and `flash_stm32h7.c` for the two architectures. I did this for the ADC: https://github.com/OpenRTX/OpenRTX/tree/master/platform/drivers/ADC

There is a lot of common code for the logic of the internal flash itself, should it stay in the header file?

platform/drivers/NVM/flash_stm32.c
silseva: I'm speaking of the init and terminate functions, is not mandatory for a driver to have them: in this case they are empty and we can directly omit them.

platform/drivers/NVM/flash_stm32.h
mdiepart: Also, regarding the sector size, I don't think a define would cut it. For Module17, we may want to use 16K sectors for the (tiny) vfo and 128k pages for the settings (we could not use 16K pages for both because I don't think the first one can be used for this?).

mdiepart:
I added a bunch of fixups (I am voluntarily not rebasing yet so that you can review changes incrementally).
I mainly renamed STM32Flash to internalFlash. 
I then moved the two copies of flash.h to peripherals/flash.h. I also added two variables to stm32yyyxxx/flash.c instead of internal_flash.c. I haven't moved the enum for sector sizes yet, I still have to find a way to declare this somewhere else, c.f. my answer to your comment above :stuck_out_tongue: .

To be continued tomorrow :)

Edit: I see I messed up and code does not compile, I'll fix it tomorrow.

platform/drivers/NVM/flash_stm32.h
silseva: Ok, we have a usecase to keep the sector size as an enum: let's keep it as it is.

silseva:
> I mainly renamed STM32Flash to internalFlash.

I'd have kept it named `stm32Flash`, so that there is a reference to the target device. Actually, I'd change the prefix of the function names to `flashStm32`: we already have `spiStm32` and `adcStm32` in the drivers, so I'd go on with this convention for consistency.

> I then moved the two copies of flash.h to peripherals/flash.h. I also added two variables to stm32yyyxxx/flash.c instead of internal_flash.c.

> There is a lot of common code for the logic of the internal flash itself, should it stay in the header file?

From the partial reorganization you did, I can see that in the end the common code is around 70 lines: I think we can just have it duplicated in the two implementation files and call it a day.

So, I'd go for a common header file and two different implementations for each MCU family. This allows also for some optimizations: for example in the case of the STM32H7 the function performing bounds checking on the address can be reduced to checking only if the address is above or below the address range.

mdiepart:
> I'd have kept it named `stm32Flash`, so that there is a reference to the target device. Actually, I'd change the prefix of the function names to `flashStm32`: we already have `spiStm32` and `adcStm32` in the drivers, so I'd go on with this convention for consistency.

Well, the way it have been re-organized makes the base code completely generic, so keeping a specific name does not make much sense to me...

There is also an issue with the enums. I can't use "opaque enums" so I have to get rid of them. The best way I found is to write the erase_size in the init function but that also means that I have to get rid of most of the const specifiers everywhere... . I did try this and to also remove the priv structure which forces me to remove the boundary checks function... This grew up in a messy direction and stopped it there for tonight. I hadn't read your latest reply yet, but this may help in implementing device specific code in another way making code neater...

TL;DR: I tried things, didn't like it, will try again later ^^'

In the meantime, the latest fixups should make the code compile again.

mdiepart:
The massive amount of fixups got too confusing and I rebased the branch to clean-up the history.

I think this version is becoming quite good, though heavy rework was involved and there may be small issues here and there, mainly regarding coding style. Also, I haven't had time to execute the platform tests I wrote since then.

I had to remove the const specifier to some nvm structs and deciced to remove them all. I don't think it makes that much sense to keep them and it makes it impossible to have reasonably flexible code without massive code deduplication. 

Let me know what you think.

Also, maybe @turnrye and @marcoSchr could give their opinion?

silseva:
> I had to remove the const specifier to some nvm structs and deciced to remove them all. I don't think it makes that much sense to keep them and it makes it impossible to have reasonably flexible code without massive code deduplication.

The NVM structs have to be `const` actually: the reason is that, by doing so, their instances end up in the `.rodata` area which resides in flash memory. Given that in the 99% of the cases the data contained in those structs will never change, is better to not "waste" RAM space for them.

You can have const pointers to structs and still be able to change their fields. I'm already doing this in the zephyr flash driver: https://github.com/OpenRTX/OpenRTX/blob/b262fd5fe1e1959c53f24d39cc08b8b8b9920941/platform/drivers/NVM/flash_zephyr.h#L29 https://github.com/OpenRTX/OpenRTX/blob/b262fd5fe1e1959c53f24d39cc08b8b8b9920941/platform/drivers/NVM/flash_zephyr.c#L32

To handle the additional information for the STM32F4 flash, I'd do the following:
```
#define INTERNAL_FLASH_DEVICE_DEFINE(name)        \
    static struct nvmInfo name##_nvmInfo;         \
    static struct internalFlashArea name##_priv;  \
    static struct nvmDevice name = {              \
        .priv = &name##_priv,                     \
        .ops = &internalFlash_ops,                \
        .info = &name##_nvmInfo,                  \
     };
```
and then fill the fields of `priv` in the initialization function.

Another option is to go back to the `enum` for sector sizes and have something like this:
```
enum sectorSize {
    SECTOR_SIZE_16K
    SECTOR_SIZE_64K
    SECTOR_SIZE_128K
};

#define INTERNAL_FLASH_DEVICE_DEFINE(name, sector_size)        \
    static struct nvmInfo name##_nvmInfo;         \
    static struct nvmDevice name = {              \
        .priv = (const void *) sector_size,                     \
        .ops = &internalFlash_ops,                \
        .info = &name##_nvmInfo,                  \
     };
```
and then use `dev->priv` as the index into the array of `struct internalFlashArea`. For STM32H7 this will be unused, but this detail is hidden into the implementation file.

mdiepart:
I don't think going back to enums is possible as the enum would need to be in the header file which is device agnostic. Or we would need to define every possible page size in the enum but then we need a way to map the sector sizes from the enum to an index in the array... I don't like the idea.

Also, wile it is possible to have a const pointer to a struct, this means that then, there is both space used in flash for the pointer and then space used in ram for the struct content? We are talking about a few dozen bytes.

silseva:
I'd say, let's make the header file device-specific for STM32. We also have to take into account that other MCUs may have different ways to access and manage the internal flash memory, so we'll probably end up having a set of device-specific drivers in any case.
Another thing that I forgot to mention, is that I'd like to move all the drivers that can be directly instantiated from OpenRTX main codebase into the `platform/drivers` folder.

mdiepart:
This is starting to be change after change after change... I'm wondering if I'm up to the task...

mdiepart:
Ok, let's step back and make sure of what to do.
I have two PR waiting to be reviewed and I can't afford to rewrite each PR three times to change the angle of attack, change the features set, or handle hypothetical use cases.

This PR was initially mainly for Module17 to store settings using the current nvm driver. I thus created a driver for the internal flash of that MCU (that covers more than half the OpenRTX devices). I also added support for STM32H743 because this was a low-hanging fruit. Then I had to make the header device generic which is a major PITA because of all the const values everywhere in the storage driver. Now I need to make the header device specific and we are talking about supporting MCUs that, for now, don't need this driver.

I say, we fix minor issues about this driver and move on. Later down the road, if we need to support additional devices, we will see if it requires making the driver device specific. It will probably not because the header file contains basically nothing except for the type of driver and the sector size (which is generic enough).

mdiepart:
I removed the commit removing the const specifiers, and I cast the nvmInfo struct to a non-const struct in the init function.

silseva:
> Ok, let's step back and make sure of what to do. I have two PR waiting to be reviewed and I can't afford to rewrite each PR three times to change the angle of attack, change the features set, or handle hypothetical use cases.
> 
> This PR was initially mainly for Module17 to store settings using the current nvm driver. I thus created a driver for the internal flash of that MCU (that covers more than half the OpenRTX devices). I also added support for STM32H743 because this was a low-hanging fruit. Then I had to make the header device generic which is a major PITA because of all the const values everywhere in the storage driver. Now I need to make the header device specific and we are talking about supporting MCUs that, for now, don't need this driver.
> 
> I say, we fix minor issues about this driver and move on. Later down the road, if we need to support additional devices, we will see if it requires making the driver device specific. It will probably not because the header file contains basically nothing except for the type of driver and the sector size (which is generic enough).

Acyually in my mind the plan was to go just for an STM32F4-specific driver to minimize the effort, but I wasn't clear in telling that.
Sorry for the mess and thanks for the patience. ❤️

mdiepart:
It's all good. However I tried running the platform tests and something is wrong. I'm glad I wrote them but now I'm trying to debug whatever is happening.

mdiepart:
Ok, only thing left: 

Should we rewrite the MDx/Mod17 nvmem api to use this one? Those two devices are the only reason I left the original flash.h/flash.c files.

silseva:
> Ok, only thing left:
> 
> Should we rewrite the MDx/Mod17 nvmem api to use this one? Those two devices are the only reason I left the original flash.h/flash.c files.

I'd say no. When the new setting system is deployed, we move everything to the new API and then cleanup all the old files

silseva:
> It's all good. However I tried running the platform tests and something is wrong. I'm glad I wrote them but now I'm trying to debug whatever is happening.

I can take care of it, if you want

mdiepart:
> > It's all good. However I tried running the platform tests and something is wrong. I'm glad I wrote them but now I'm trying to debug whatever is happening.
> 
> I can take care of it, if you want

Fix is in the fixups but please do run platform tests on your side too :)

PR #406, opened 2026-02-26
Refactoring and update of nonvolatile memory subsystem

silseva:


mdiepart:
This does look good!
I'll compile and run it on my devices tomorrow :)

mdiepart:
Do you want to merge this into master now? Or do you want to make a bigger PR with all the NVM features currently in development?

mdiepart:
I think posix_files should use an nvmDevice rather than nvmFileDevice. The *priv element of the structure should be used to store the file descriptor.

We should either store the fd itself within priv (cast the int to a pointer to void), or malloc the size of an integer. Those may need to remove the const specifier from priv (which would make sense anyway I think, so that the init function can manage *priv)

openrtx/src/core/nvmem_access.c
```diff
@@ -4,108 +4,98 @@
...
 
-    return nvm_devRead(nvm->dev, offset, data, len);
-}
-
-int nvm_write(const uint32_t dev, const int part, uint32_t offset, const void *data,
-              size_t len)
-{
-    const struct nvmDescriptor *nvm = nvm_getDesc(dev);
-    if(nvm == NULL)
+    // Accesses exceeding the nvm->size boundary are caugth by this check,
```
turnrye:
```suggestion
    // Accesses exceeding the nvm->size boundary are caught by this check,
```

openrtx/include/interfaces/nvmem.h
```diff
@@ -103,37 +100,46 @@ struct nvmOps
...
-{
-    const char                *name;        ///< Name
-    const struct nvmDevice    *dev;         ///< Associated device driver
-    const size_t               partNum;     ///< Number of partitions
-    const struct nvmPartition *partitions;  ///< Partion table
+struct nvmDescriptor {
+    const char *name;                      ///< Name
+    const struct nvmDevice *dev;           ///< Associated device driver
+    const size_t baseAddr;                 ///< Base address of the area
+    const size_t size;                     ///< Area size, in byte
```
turnrye:
```suggestion
    const size_t size;                     ///< Area size, in bytes
```

openrtx/include/interfaces/nvmem.h
```diff
@@ -103,37 +100,46 @@ struct nvmOps
...
+    const size_t nbPart;                   ///< Number of partitions
+    const struct nvmPartition *partitions; ///< Partion table
```
turnrye:
```suggestion
    const struct nvmPartition *partitions; ///< Partition table
```

platform/drivers/NVM/nvmem_CS7000.c
```diff
@@ -63,17 +58,30 @@ static const struct nvmDescriptor extMem[] =
...
+#endif
+        .nbPart     = sizeof(memPartitions)/sizeof(struct nvmPartition),
         .partitions = memPartitions
     },
     {
         .name       = "Virtual EEPROM",
         .dev        = &eeep,
-        .partNum    = 0,
+        .baseAddr   = 0x00000000,
+        .size       = 0xFFFFFFFF,
```
turnrye: The other size definitions have comments that explain their values; is this a special value? Should we document this for posterity's sake?

silseva:
> I think posix_files should use an nvmDevice rather than nvmFileDevice. The *priv element of the structure should be used to store the file descriptor.
> 
> We should either store the fd itself within priv (cast the int to a pointer to void), or malloc the size of an integer. Those may need to remove the const specifier from priv (which would make sense anyway I think, so that the init function can manage *priv)

I'll create a Taiga task for this

PR #408, opened 2026-03-02
build: linux: enable stdio by default

mdiepart:
This PR enables STDIO by default for linux builds. Given that their main use is to allow developers to debug OpeNRTX more easily, I think enabling it makes sense.


silseva:
There is no need to define the `ENABLE_STDIO` macro in linux, actually. It is used only in the libc_integration files for the embedded devices: https://github.com/search?q=repo%3AOpenRTX%2FOpenRTX%20ENABLE_STDIO&type=code

On linux all the printf-like functions go directly to stdout.

mdiepart:
Ah, I could have bet that I had an issue with printf printing nothing to the output on linux... But indeed, ENABLE_STDIO has nothing to do with it ^^'

PR #409, opened 2026-03-02
NVM: Reorganize VFO and settings

mdiepart:
This PR re-organizes the VFO and settings structure such that parameters that change frequently are stored in the VFO and parameters that change rarely are stored in the settings structure.

This is work in progress as this currently breaks the VFO for existing devices. The VFO must be updated such that when the radio detects that the VFO is outdated, it uses the default VFO. This was, for now, done for MDx only.


mdiepart:
The last commit fixes the VFO versioning for CS7000. Module17 does not have a VFO and the other devices do not have persistence yet.

The next step regarding VFO is to implement a storage scheme, common to all devices if possible. This was already worked on by @turnrye with the slice storage iirc.

silseva:
Do you mind splitting this in two PRs, one for settings and one for VFO state?

mdiepart:
The settings part is really in #410.
Is mainly adresses the settings moving from settings to vfo

mdiepart:
@silseva I removed the implementation commits from this branch and changed the target branch to persistence. 
This PR basically re-organizes the fields between VFO and settings. I'll get started on the new VFO storage (sliced).

mdiepart:
Last commit bumps the cps version as the structure actually changed.

openrtx/include/core/settings.h
```diff
@@ -51,28 +51,24 @@ display_timer_t;
...
-    char    m17_dest[10];         // M17 destination
-    bool    showBatteryIcon;      // Battery display true: icon, false: percentage
-    bool    gpsSetTime;           // Use GPS to ajust RTC time
-    char    M17_meta_text[53];    // M17 Meta Text to send
+    uint8_t  brightness;           ///< Display brightness
+    uint8_t  contrast;             ///< Display contrast
+    uint8_t  voxLevel;             ///< Vox level
+    int8_t   utc_timezone;         ///< Timezone, in units of half hours
+    char     callsign[10];         ///< Plaintext callsign
+    uint16_t display_timer   : 4,  ///< Standby timer
```
silseva: I'd use `uint8_t`s and group the fields together by logical separation. For example:

```
uint8_t display_timer   : 4,  ///< Standby timer
        gps_enabled     : 1,  ///< GPS active
        gpsSetTime      : 1,  ///< Use GPS to ajust RTC time
        showBatteryIcon : 1,  ///< Battery display true: icon, false: percentage
        m17_can_rx      : 1;  ///< Check M17 CAN on RX

uint8_t vpLevel         : 3,  ///< Voice prompt level
        vpPhoneticSpell : 1,  ///< Phonetic spell enabled
        vpPhoneticSpell : 1,  ///< Phonetic spell enabled
        _reserved : 3;
```

openrtx/include/core/cps.h
```diff
@@ -231,6 +229,7 @@ typedef struct
             rx_only        : 1,    //< 1 means RX-only channel
             _unused        : 5;    //< Padding to 8 bits
 
+    uint8_t  sqlLevel;             //< Squelch level
```
silseva, line 237: Even though RTX-related, I'd consider the squelch level part of the settings because is something you set independently of which channel the radio is tuned to.

openrtx/include/core/cps.h
```diff
@@ -243,14 +242,21 @@ typedef struct
     char    descr[CPS_STR_SIZE];   //< Description of the channel
     geo_t   ch_location;           //< Transmitter geolocation
 
+    // These three values should really be in m17Info_t below but
+    // because these can be changed in the settings at any time,
+    // they can not be dependant on the selected opMode.
+    uint8_t rx_can : 4,             ///< Channel Access Number for RX
```
silseva: AFAIK the channel access number is one, for both TX and RX. Is like the DMR color code: you cannot transmit with one CC and receive on another.
Better to ask in the M17 discord, though...

silseva:
About the VFO channel: as I mentioned a couple of times in the last days, I think that recycling a CPS channel for it is not the best way to go. Even more if you take into account that the current CPS implementation has major design flaws and needs to be redone (I wouldn't have pushed for PTSCD, otherwise).

What I had in mind for the VFO is something similar:
```
struct fmCfg {
    uint8_t rxTone;          //< RX CTC/DCS tone index
    uint8_t txTone;          //< TX CTC/DCS tone index
    uint8_t rxToneEn  : 1,   //< RX CTC/DCS tone enable
            txToneEn  : 1,   //< TX CTC/DCS tone enable
            bandwidth : 2,   //< 12.5 / 25kHz
            _reserved : 4;   //< TODO: tone type (CTCSS/DCS/...)
};

struct m17Cfg {
    char    destination[10];    //< M17 destination
    uint8_t can       : 4,      //< Channel access number
            _reserved : 4;      //< TODO: encryption, ...
};

struct vfo {
    uint32_t power;         //< Tx power, in mW
    uint32_t rx_frequency;  //< RX Frequency, in Hz
    uint32_t tx_frequency;  //< TX Frequency, in Hz
    uint8_t mode;           //< Operating mode

    struct fmCfg fm;
#ifdef CONFIG_M17
    struct m17Cfg m17;
#endif
};
```
Given that the VFO channel is the "one channel to rule 'em all" because it includes all the operating modes at the same time, I think it is worth spending a bit of time to design all the data structures properly: then we'll use them as-is also for the codeplug channels.

Conceptually (and with more other fields I just omitted), a codeplug channel becomes:
```
struct channel {
    uint32_t power;         //< Tx power, in mW
    uint32_t rx_frequency;  //< RX Frequency, in Hz
    uint32_t tx_frequency;  //< TX Frequency, in Hz
    uint8_t mode;           //< Operating mode

    union {
        struct fmCfg fm;
        struct m17Cfg m17;
    };
};
```

openrtx/include/core/cps.h
lemielek: Aren't squelch levels separate - or could be - for each vfo/band?
Cannot check on TRXes because I am at hospital, but I have that thing in mind.
Or, I don't understand that thread question.

openrtx/include/core/cps.h
mdiepart: I will merge both to a single CAN value.

openrtx/include/core/cps.h
mdiepart: I do think that squelch is better located in the VFO. You may want different squelch values in different bands, or on different channels (close-by repeater,...). Squelch is also something that may be modified at every boot and would thus cause the full settings structure to be written to flash at every boot, which is something we want to avoid.

openrtx/include/core/settings.h
mdiepart: I'm not sure there is a real added value to this and we risk fragmenting the flags. For example, the 3 reserved bits for vp would either never get used (and be lost, even though it is only 3 bits) or will later be used for something different anyway. In fact, I even though about making a single uint32_t where we would put all the flags and keep some empty space for potential future flags and have them all in the same place in the settings...

mdiepart:
I'll get started on a separate vfo structure then :)
Should the channel_t struct be put aside for the time being?

Also, regarding the GNSS struct, do you agree that it should be removed from the VFO and be placed somewhere else? (some runtime-only struct)

openrtx/include/core/cps.h
silseva: Fair enough. Then I'd put it among the FM-only settings, because is something you use only for analog modes

silseva:
> I'll get started on a separate vfo structure then :) 
> Should the channel_t struct be put aside for the time being?

Absolutely. The `channel_t` data structure is obsolete and it will be removed when switching to PTSCD.

> Also, regarding the GNSS struct, do you agree that it should be removed from the VFO and be placed somewhere else? (some runtime-only struct)

Absolutely again. :)
For the VFO data structure we should start from a blank file and see what does _need_ to be put there. As I said, the VFO is a special channel and should be treated as such.

mdiepart:
Here is what I have in mind. I'll try to get started on this this evening or (more probably) tomorrow evening.

-  "VFO" should actually be named "Tuner". The tuner can be in VFO mode or Memory mode.
- Tuner contains only the data that needs to be written to disk:
  -  A version field
  - A VFO/Memory flag
  - A channel number for the memory mode
  - Rx and Tx frequencies, power, mode, mode-related parameters, ...
Access to main tuner fields would be implemented by functions. These would return the correct freq/power/mode/... depending on if the tuner is in VFO or memory mode.

A copy of the current channel (for memory mode) can be somehow added to the tuner for caching and fast access but must not be written to NVM.

In terms of structure layout, this could be solved with the tuner_t structure containing a "kernel" with the fields needing to be written to NVM (version, mode, VFO/Mem flag) and a channel_t structure containing the memory channel data for quick access. Saving the tuner do disk would only save the kernel. Loading the tuner from nvm would read the kernel and populate the channel. This implementation is independent enough (imo) from the actual implementation of the codeplug in the background. Changes would be needed when the CPS overhaul will land, but these would probably stay limited, partly because the changes would be limited to the functions that return data from the codeplug. 

What do you think of this approach?

silseva:
I totally forgot about the need to have persistent VFO/memory mode and channel number. Now the question is: is there something else, in general, that we have to keep persistent across reboots? It seems like that at this point we are shifting towards more towards the concept of persistent state than just the tuner.

Apart from this, the approach is definitely good. I'm just wondering whether is good to have something like this
```
struct vfo {
    uint32_t txFrequency;
    [...]
};

struct tuner {
    uint32_t chIndex;
    uint8_t mode;    // VFO/MEM
    struct vfo vfo;
    struct channel channel;
};
```
or having the VFO data merged into the tuner structure
```
struct tuner {
    uint32_t chIndex;
    uint8_t vfo_mem;
    uint32_t txFrequency;
    [...]
    struct channel channel;
};
```

silseva:
On a side note: I'm progressively removing all the `typedef struct` expressions in favor of just having `struct` in order to make clear that is a composite type. Using plain `struct` allows also to use forward declaration, allowing to not include the header with the structure definition if not necessary and, consequently, reducing the `#include` dependencies.

mdiepart:
I started implementing the tuner, let me know what you think. 

Quick comment: I know the latest commit breaks things, it will be fixed but I wanted to get the draft out asap for early review.

mdiepart:
So I took a stab at reworking the channel_t into a tuner struct and it touches basically every part of the code (ui, voice prompts, the radio drivers, ...). At this stage it will either heavily delay the new NVM implementation or, if we want NVM soon, the job needs to be botched which I don't think is good. 
The NVM APIs that will follow (vfo slicer #456 ) or were implemented (#410) do support versioning of the structures, meaning that we can move forward with this PR without doing the full rework, and implementing it later, taking our time to do it right. 
I will remove the latest WIP commit and open a draft PR.

The various changes that still need to be done for the original topic of this PR will be performed in the coming days :)

meson.build
```diff
@@ -1117,7 +1117,7 @@ m17_rrc_test = executable('m17_rrc_test',
                           kwargs: unit_test_opts)
 
 cps_test = executable('cps_test',
-                      sources : unit_test_src + ['tests/unit/cps.cpp'],
+                      sources : unit_test_src + ['tests/unit/cps.cpp', 'openrtx/src/core/cps.c'],
```
silseva, line 1120: `openrtx/src/core/cps.c` should be already part of `unit_test_src`, isn't it?

mdiepart: Not it is not, `unit_test_src` only uses `linux_src` and `ui_src_default`, neither of which uses `openrtx_src` in which `openrtx/scr/core/cps.c` is included.

PR #410, opened 2026-03-02
Implementation of a common settings storage API

mdiepart:
This PR implements a new storage scheme, common to all devices, for storing settings in NVM.
The implementation uses two NVM partitions (called A and B). Latest settings are stored alternatively in partition A and B. The settings are stored as a single chunk along with a magic word, the length, a counter and a CRC. Thanks to the content stored along with the settings, the radio can gracefully handle previous versions of the settings by comparing the lengths of the stored settings with the current size of the structure. This implies that from now on, any additional setting gets appended to the current structure. 

While the implementation have already been tested on MD-UV380 and linux, tests did not go much further.

- [x] Test the implementation with corrupted data
- [x] Test the implementation with old settings

I have no idea how things are done in T-TWR Plus as I haven't looked into it yet.

Most importantly, this PR must wait for #409 as that PR reduces the size of the settings structure which is not allowed once this API is merged.


mdiepart:
I have tested the implementation against corrupted and outdated settings, squashing a few bugs. 
I also started a larger discussion in TG-780 regarding the implementation for other devices.
Also, the zephyr build seems to have an error with my part_status enum and I have absolutely no idea why...

mdiepart:
Seems like EMPTY is a reserved keyword of zephyr? 
Renaming the fields seems to have done it :)

mdiepart:
I decided to remove the implementation commits from the PR so this one can get merged now. A separate PR will implement the new NVM APIs for as many devices as possible.

openrtx/include/core/settings.h
```diff
@@ -81,4 +99,76 @@ static const settings_t default_settings =
...
+{
+    PART_CLEAN,      ///< Partition contains entries
+    PART_EMPTY,      ///< Partition contains no entry
+    PART_CORRUPTED,  ///< Partition contains invalid data
+} part_status;
+
+/**
+ * Structure defining the binary layout of the settings in NVM memory
+ */
+typedef struct {
```
silseva: Proposal: if we have something like
```
struct settingsHeader {
    uint32_t MAGIC;
    uint16_t length;
    uint16_t counter;
    uint16_t crc;
};

struct settingsBlock {
    struct settingsHeader header;
    settings_t settings;
};
```
does it make your life easier while handling data in the nonvolatile memory? Instead of having to read six bytes into a temporary buffer and do pointer math, you can just read `sizeof(struct settingsHeader))` and access the fields.
Then, to compute the CRC, we set the rule that it has to be computed by setting the `header.crc` field to 0x0000. Code becomes:
```
struct settingsBlock block;

// Compute and update
block.crc = 0;
uint16_t crc = crc_ccitt(&block, sizeof(struct settingsBlock));
block.crc = crc;


// Verify
uint16_t crc = block.crc;
block.crc = 0;
if (crc_ccitt(&block, sizeof(struct settingsBlock) != crc)
    // cry
```

silseva:
General comment: I usually add `static` to all the helper functions and variables that are being used only inside a given .c file to reduce symbol visibility. I already had surprises due to two variables in different files having the same name being silently merged at link time...

openrtx/include/core/settings.h
mdiepart: Well, strictly speaking, what makes my life easier is not rewriting the code :p
But if you think it makes the code easier *to maintain* though, I can make the change.

mdiepart: I started doing the changes to see what it would change, and it does actually remove quite a few hardcoded values (header length, offsets, ...) which, imo, helps making the code more legible, so I will keep going and put that in a fixup.

silseva: Yes, the main advantage is that it makes the code more easier to read and maintain in the future. By using `sizeof` and, eventually `offsetof`, in the future we can add new fields to the data structures without having to change many lines of code.

mdiepart:
Is this good to go?

openrtx/src/core/settings.c
```diff
@@ -0,0 +1,479 @@
...
+ */
+static int find_last_store(const int dev, const int part_nb, size_t *offset,
+                    size_t limit)
+{
+    settings_header_t header_buffer;
+    *offset = 0;
+    size_t prev_offset = 0;
+
+    // Go through the partition to find settings struct
+    while (*offset < (limit - 6)) {
```
silseva: Shouldn't be
```suggestion
    while (*offset < (limit - sizeof(settings_header_t)) {
```
?

openrtx/src/core/settings.c
```diff
@@ -0,0 +1,479 @@
...
+        int ret = nvm_read(dev, part_nb, *offset, &header_buffer, 6);
```
silseva:
```suggestion
        int ret = nvm_read(dev, part_nb, *offset, &header_buffer, sizeof(settings_header_t));
```

openrtx/src/core/settings.c
```diff
@@ -0,0 +1,479 @@
...
+        prev_offset = *offset;
+        *offset += header_buffer.length;
+    }
+
+    /*
+     * If offset == prev_offset == 0 and magic == 0xFFFFFFFF, the partition is empty
+     * If offset > prev_offset and magic == 0xFFFFFFFF there is free space
+     * If magic != 0xFFFFFFFF then the partition contains invalid data and needs to be cleaned
+     */
+    if ((header_buffer.MAGIC != 0xFFFFFFFF) && (header_buffer.MAGIC != SETTINGS_MAGIC))
```
silseva: We need to check also for `header_buffer.MAGIC != 0`, otherwise the function will always return -EILSEQ when the storage backend is a file or an eMMC

openrtx/src/core/settings.c
```diff
@@ -0,0 +1,479 @@
...
+        return ret;
+
+    if (store->header.length <= sizeof(settings_store_t)) {
+        if(store->header.length < sizeof(settings_store_t)) {
+            // Pre-init with default settings
+            store->settings = default_settings;
+        }
+
+        // read settings
+        ret = nvm_read(dev, part, offset + sizeof(settings_header_t), &(store->settings),
```
silseva: Shouldn't this be
```suggestion
        ret = nvm_read(dev, part, offset + sizeof(settings_header_t), &(store->settings), store->header.length);
```
?

openrtx/src/core/settings.c
```diff
@@ -0,0 +1,479 @@
...
+        return ret;
+
+    // Check if we have enough space to write the store
+    if ((*offset + sizeof(settings_store_t)) > pInfo.size)
+        erase = true;
+
+    if (erase) {
+        ret = nvm_erase(dev, part, 0, pInfo.size);
+        if (ret == -ENOTSUP) {
+            // If device does not implement erase function (such as posix files)
```
silseva: If at line 137 you handle also the case of magic == 0, this probably is not anymore necessary (?)

openrtx/src/core/settings.c
mdiepart: Absolutely, thank you

openrtx/src/core/settings.c
mdiepart: The way I currently handle files is that I consider them corrupted if it is all zeroes. Because there is no "erase" function, I know this is a specific case and then I write 0xFF to the area so that the logic stays the same for all implementations. I am not sure how this would apply to eMMC. Do we have any example of a radio using raw eMMC device?

openrtx/src/core/settings.c
mdiepart: Yes

openrtx/src/core/settings.c
mdiepart: I would still have to handle cases where the device does not have any erase function...

openrtx/src/core/settings.c
mdiepart: The length in the header is the length of the structure including the header itself, so I have to subtract it when reading the settings structure.

mdiepart:
Last commits address: 
- hardcoded values replaced with sizeof
- applying clang-format

openrtx/src/core/settings.c
silseva: Aha! The first time I read the code I didn't notice it.

openrtx/src/core/settings.c
silseva: From what I remember an erased eMMC reads all zeroes instead of 0xFFFF. Actually the intent of my comment was to make the implementation future-proof rather than covering an existing case.

About the fill with 0xFF: I re-read all the code and now I understand why you do so. It came to my mind a case where this is not very efficient, though: when the NVM storage is an I2C eeprom (like in GD-77 and DM-1801) the settings system is filling an entire block with 0xFF and immediately after it writes a valid entry at the top. Apart from taking some time, this spends extra write life of the eeprom memory itself every time a block needs to be erased.

If I understood all the logic correctly, we can probably change it to consider a partition either "valid" or "invalid", with the latter case being either a partition with corrupted data or a totally empty one. When writing to an invalid partition, then, it will be enough to directly write the new data block (preceded by an erase, if necessary) without the 0xFF fill.
The only downside of this approach that comes to my mind is that, when an already-erased flash memory is used, this will be considered invalid and erased again even if not necessary. This, however, should happen only the first time the code is run on the device. 

What do you think? I could have missed some important detail...

mdiepart: I had a quick look. It looks like eMMC reads 0xFF partly because it uses flash memory for the physical support. This however made me think about how to handle "corrupted" memory. The only reason we *need* to erase memory before writing is in the cases where we can only erase a full page. Maybe the solution is to not erase memory when there is no erase function and just overwrite what is considered "corrupted". 
I'll implement that hopefully tomorrow. Maybe later down the road if we have an eMMC use case or rather a use case with memory that erases to 0x00, it will be time to see if something needs to be changed in either the settings API or the NVM API.
Does this seems good to you?

silseva: > I had a quick look. It looks like eMMC reads 0xFF partly because it uses flash memory for the physical support.

My bad, I thought eMMC had a different behavior. I got confused by the experience I had with one of those in the past.

> This however made me think about how to handle "corrupted" memory. The only reason we _need_ to erase memory before writing is in the cases where we can only erase a full page. Maybe the solution is to not erase memory when there is no erase function and just overwrite what is considered "corrupted". I'll implement that hopefully tomorrow.

Yes, that's my thought. When we do not need to erase memory, writing a new valid entry at the beginning of the page should be enough to make it again valid. All the old data below this new entry will be anyways overwritten over time.

> Maybe later down the road if we have an eMMC use case or rather a use case with memory that erases to 0x00, it will be time to see if something needs to be changed in either the settings API or the NVM API.

Yes, let's leave it for later. As Italian people say: "Don't bandage your head before breaking it". :smile:

mdiepart:
I started modifying the API to better handle devices that do not require erasing before writing. This is not working yet... The last commit actually introduced some bugs.

mdiepart:
Well, I talked too soon, and squashed the bugs. I checked the implementation on linux (on my side, the current commits don't currently propose any implementation). Implementations will come later in other PRs.

I rebased on master but kept the fixups so that you can keep your review incremental. The last fixup of the lot contains the actual changes to support memories that do not require erase before write. 

Basically, if I detect an incorrect magic word in the header, if I don't need to erase before write, then I need to check if the incorrect magic word is at the beginning of the partition (empty partition) or further down the partition (then the partition actually contains some valid settings.

Let me know what you think of this, so that we can soon be done with this part of the NVM rework and move on with the rest.

Also, I think we should not merge to master but to a dedicated nvm rework branch and merge that branch to master once we are done. It will be easier given what I have planned for the rest of the rework (we will avoid the risk of having radios in the wild that are between two steps of the rework).

silseva:
> Let me know what you think of this, so that we can soon be done with this part of the NVM rework and move on with the rest.

C'est parfait!

> Also, I think we should not merge to master but to a dedicated nvm rework branch and merge that branch to master once we are done. It will be easier given what I have planned for the rest of the rework (we will avoid the risk of having radios in the wild that are between two steps of the rework).

Oui, absolument! J'ai eu la meme idée, je vais a créer un nouveau branch et je ferai le merge la.

silseva (approved):

PR #411, opened 2026-03-03
Cache devcontainer for builds

turnrye:
This change updates the CI to publish the devcontainer. This is to simplify the build process so that the `build` stage doesnt have to configure the entire build environment each time. This also addresses some build instability that @mdiepart was running into where network flakiness was leading to build failures.

[Here's an example build](https://github.com/turnrye/OpenRTX/actions/runs/22604554084/job/65493673952) on my fork where the build step took 3min 21sec.

Note that this change will result in a package being published with the name `openrtx-devcontainer`, with the only tag being `latest` always.

The first build after merge will need to be fired off manually, and it will only be updated in the future when a python requirement, the dockerfile, or the devcontainer config file change (as opposed to today where it's re-built each time).

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,5 +1,7 @@
 FROM ubuntu:24.04
 
+LABEL org.opencontainers.image.source=https://github.com/OpenRTX/OpenRTX
+
 # Install dependencies
 RUN apt-get update -y && apt-get install -y --no-install-recommends \
   git \
--- a/.github/workflows/build-devcontainer.yml
+++ b/.github/workflows/build-devcontainer.yml
@@ -1,0 +1,43 @@
+name: "Build devcontainer"
+on:
+  workflow_dispatch:
+  push:
+    paths:
+      - 'requirements.txt'
+      - '.github/workflows/build-devcontainer.yml'
+      - '.devcontainer/Dockerfile'
+      - '.devcontainer/devcontainer.json'
+  pull_request:
+    paths:
+      - 'requirements.txt'
+      - '.github/workflows/build-devcontainer.yml'
+      - '.devcontainer/Dockerfile'
+      - '.devcontainer/devcontainer.json'
+
+env:
+  RADIO_TOOL_VERSION: 0.2.2
+  BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
+  DEVCONTAINER_NAME: ghcr.io/openrtx/openrtx-devcontainer
+jobs:
+  build:
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v2 
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Pre-build dev container image
+        uses: devcontainers/ci@v0.3
+        with:
+            imageName: ${{ env.DEVCONTAINER_NAME }}
+            cacheFrom: ${{ env.DEVCONTAINER_NAME }}
+            push: always
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,6 +7,7 @@ on:
 env:
   RADIO_TOOL_VERSION: 0.2.2
   BRANCH_NAME: ${{ github.head_ref || github.ref_name }} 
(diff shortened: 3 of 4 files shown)


silseva:
I just had a very quick look, but is this somehow related to #348 and #350?

turnrye:
Yes, related. Those had a different goal though. Peter was trying to prebuild the image so that devs wouldn't have to build it locally when using the devcontainer. My goal here is different, focusing instead on prebuildimg the image so that CI wouldn't have to rebuild the image each time the build job ran. As a result, I don't have to worry about multi architecture builds like he did, which simplifies things significantly.

It does remind me that of course this new build should trigger on changes to the dockerfile or the devcontainer config, so I'll add those.

turnrye:
Ok, I've added those additional files to trigger the build too. This is ready for a review and merge.

turnrye:
For this to work, @silseva you'll need to give permissions to github ci to create packages:

> denied: installation not allowed to Create organization package

mdiepart:
Nice! Thank you for your help :D

silseva:
> For this to work, @silseva you'll need to give permissions to github ci to create packages:
> 
> > denied: installation not allowed to Create organization package

It seems that you have to add something to the yaml file. At least according to this discussion: https://github.com/orgs/community/discussions/57724
But I may have understood nothing because I'm quite tired...

silseva:
> Yes, related. Those had a different goal though. Peter was trying to prebuild the image so that devs wouldn't have to build it locally when using the devcontainer. My goal here is different, focusing instead on prebuildimg the image so that CI wouldn't have to rebuild the image each time the build job ran. As a result, I don't have to worry about multi architecture builds like he did, which simplifies things significantly.

Do you think that it does make sense to have an image that works both for developers and the CI? It does not need to support all the architectures like Peter did, at least for now.

turnrye:
> > Yes, related. Those had a different goal though. Peter was trying to prebuild the image so that devs wouldn't have to build it locally when using the devcontainer. My goal here is different, focusing instead on prebuildimg the image so that CI wouldn't have to rebuild the image each time the build job ran. As a result, I don't have to worry about multi architecture builds like he did, which simplifies things significantly.
> 
> Do you think that it does make sense to have an image that works both for developers and the CI? It does not need to support all the architectures like Peter did, at least for now.
> 

Yes, and I expect this does that. I plan to test that after this lands and raise changes to make that work out of the box for amd64 users. I don't want to include changes to the devcontainer configuration besides the CI job as part of this PR, because I want to quickly address the more critical issue: flaky CI.

.github/workflows/build-devcontainer.yml
```diff
@@ -0,0 +1,40 @@
...
+      - '.devcontainer/Dockerfile'
+      - '.devcontainer/devcontainer.json'
+
+env:
+  RADIO_TOOL_VERSION: 0.2.2
+  BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
+  DEVCONTAINER_NAME: ghcr.io/openrtx/openrtx-devcontainer
+jobs:
+  build:
+    runs-on: ubuntu-24.04
```
silseva, line 23: It seems that adding `permissions: write-all` at the same level of the `runs-on` keyword is enough to fix the "installation not allowed" issue

turnrye: Done -- let's see how this goes. I avoided write-all though and gave it the lesser scoped permission. I also made it such that `main.yml`, which isn't authenticated, should never try to push its container. It'll always pull the cache if its there though and fall back to building if the cache is missing (like the current state).

.devcontainer/Dockerfile
```diff
@@ -1,5 +1,7 @@
 FROM ubuntu:24.04
 
+LABEL org.opencontainers.image.source=https://github.com/OpenRTX/OpenRTX
```
turnrye, line 3: Added this based on https://docs.github.com/en/packages/learn-github-packages/connecting-a-repository-to-a-package#connecting-a-repository-to-a-container-image-using-the-command-line pointing to this as a key label to help GHCR understand where to put the package in an org-owned repo.

mdiepart:
Isn't it possible that because this is only a PR, the image can't be pushed to the org repo while if it was merged it could ?
I have no idea how this whole thing works which is why I asked if @turnrye was able to help with this to begin with 😅

turnrye:
Yes, but what I can do now is push my branch to the org repo rather than my fork and reopen the PR. That should have more luck. This is failing right now because its my fork with my forks keys trying to push to openrtx's registry.

silseva:
I'll merge this PR into a branch to see if pushing to openrtx registry works: if so, I'll merge it into master.

turnrye:
@silseva looks like this worked, the devcontainer is being published to the proper location.

Issue #413, opened 2026-03-05
CRC16 implementations could be faster and in one place


rxt1077:
While refactoring the APRS Rx code I noticed that there is a `core/crc.c` and it seems to be used for NVM stuff. CRC CCITT _is_ what HDLC (framing format for APRS) uses, but it starts with 0xFFFF instead of zero. M17 seems to use a CRC in its link setup frames as well, but has a different polynomial. We can improve the performance of all three of these CRCs by using a nibble sized (to save space) lookup table for each. We could also put them all in one place.

Testing it out on my machine, I found about a 50% speedup in CRC calculation for M17:

```C
#include <stdio.h>
#include <stdint.h>
#include <time.h>

/**
 * Static table used for by crc16M17
 */
static const uint16_t CRC16_M17_TABLE[] = {
    0x0000, 0x5935, 0xb26a, 0xeb5f, 0x3de1, 0x64d4, 0x8f8b, 0xd6be,
    0x7bc2, 0x22f7, 0xc9a8, 0x909d, 0x4623, 0x1f16, 0xf449, 0xad7c
};

/**
 * @brief Calculate the CRC used by M17 link setup frames
 *
 * Created using pycrc https://pycrc.org with the configuration:
 *  - Width         = 16
 *  - Poly          = 0x5935
 *  - XorIn         = 0xffff
 *  - ReflectIn     = False
 *  - XorOut        = 0x0000
 *  - ReflectOut    = False
 *  - Algorithm     = table-driven
 *
 *  @return A uint16_t CRC value
 */
uint16_t lookup_crc16M17(const void *data, const size_t len)
{
    uint16_t crc = 0xFFFF;
    const uint8_t *d = (const uint8_t *)data;
    uint8_t tbl_idx;

    for (size_t i = 0; i < len; i++) {
        tbl_idx = (crc >> 12) ^ (d[i] >> 4);
        crc = CRC16_M17_TABLE[tbl_idx & 0x0f] ^ (crc << 4);
        tbl_idx = (crc >> 12) ^ (d[i] >> 0);
        crc = CRC16_M17_TABLE[tbl_idx & 0x0f] ^ (crc << 4);
    }
    return crc & 0xffff;
}

uint16_t current_crc16M17(const uint8_t *data, const size_t len) {
    uint16_t crc = 0xFFFF;

    for (size_t i = 0; i < len; i++) {
        crc ^= data[i] << 8;
        for (uint8_t j = 0; j < 8; j++) {
[...]

PR #414, opened 2026-03-06
core: crc: improve M17 CRC and move to core

rxt1077:
This PR updates the CRC function used by M17 to use a 4-bit lookup table and moves it out of the `protocols/m17` to `core/crc.c`.

This addresses #413 and should allow an M17 LSF CRC to be calculated about 50% faster. It doesn't touch `crc_ccitt()` as that function with a lower-density polynomial is [already optimized beyond what a 4-bit lookup table can achieve](https://en.wikipedia.org/wiki/Computation_of_cyclic_redundancy_checks#Multi-bit_computation_using_lookup_tables).

### DIFF
--- a/openrtx/src/core/crc.c
+++ b/openrtx/src/core/crc.c
@@ -6,16 +6,33 @@
 
 #include "core/crc.h"
 
+uint16_t crc_m17(const void *data, const size_t len)
+{
+    static const uint16_t CRC_M17_TABLE[] = { 0x0000, 0x5935, 0xb26a, 0xeb5f,
+                                              0x3de1, 0x64d4, 0x8f8b, 0xd6be,
+                                              0x7bc2, 0x22f7, 0xc9a8, 0x909d,
+                                              0x4623, 0x1f16, 0xf449, 0xad7c };
+    uint16_t crc = 0xFFFF;
+    const uint8_t *d = (const uint8_t *)data;
+
+    for (size_t i = 0; i < len; i++) {
+        uint8_t tbl_idx = (crc >> 12) ^ (d[i] >> 4);
+        crc = CRC_M17_TABLE[tbl_idx & 0x0f] ^ (crc << 4);
+        tbl_idx = (crc >> 12) ^ d[i];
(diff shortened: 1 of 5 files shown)


openrtx/src/core/crc.c
```diff
@@ -6,16 +6,34 @@
...
 
+uint16_t crc_m17(const void *data, const size_t len)
+{
+    static const uint16_t CRC_M17_TABLE[] = { 0x0000, 0x5935, 0xb26a, 0xeb5f,
+                                              0x3de1, 0x64d4, 0x8f8b, 0xd6be,
+                                              0x7bc2, 0x22f7, 0xc9a8, 0x909d,
+                                              0x4623, 0x1f16, 0xf449, 0xad7c };
+    uint16_t crc = 0xFFFF;
+    const uint8_t *d = (const uint8_t *)data;
+    uint8_t tbl_idx;
```
silseva: This variable can go inside the body of the `for` loop, to minimize its scope

silseva:
Looks good to me!
Just a curiosity: why do you think is better to move this function to `core/crc.c`?

openrtx/src/core/crc.c
```diff
@@ -6,16 +6,34 @@
...
+
+    for (size_t i = 0; i < len; i++) {
+        tbl_idx = (crc >> 12) ^ (d[i] >> 4);
+        crc = CRC_M17_TABLE[tbl_idx & 0x0f] ^ (crc << 4);
+        tbl_idx = (crc >> 12) ^ (d[i] >> 0);
```
silseva: I think we can avoid the `>> 0`

rxt1077:
> Looks good to me! Just a curiosity: why do you think is better to move this function to `core/crc.c`?

I think if we make a core set of error detection and correction routines we'll see some reuse across the code base. I was surprised to find an optimized version of the CRC I needed for APRS already in the OpenRTX code. I realize that the M17 CRC uses a non-standard polynomial, but it's not out of the realm of possibility that some other radio protocol will come along and want to use the same thing.

rxt1077:
Changes fixedup and squashed!

PR #415, opened 2026-03-12
Retevis C62 platform support (based on v0.5.0)

Tunas1337:
Will add description later. Questions, reviews and contributions welcome.


platform/drivers/baseband/radio_C62.cpp
```diff
@@ -0,0 +1,267 @@
...
+        bk4819_gpio_pin_set(GPIO_VHF_TX_PA, true); // VHF TX PA
+    } else {
+        bk4819_gpio_pin_set(GPIO_UHF_TX_PA, true); // UHF TX PA
+    }
+
+    bk4819_gpio_pin_set(GPIO_ALC_TX_LED, true); // ALC / TX LED
+
+    // depending on power level set PWM duty cycle for APC voltage control
+    // Maybe need table for this instead of crude linear mapping, and also consider frequency dependence of PA efficiency
+    platform_set_tx_power(std::min(config->txPower * 100 / 5000, 100U)); // crude linear mapping of power to duty cycle, max at 5W
```
Marko13-dev, line 196: I think until we measure the exact values better use quadratic curve. It matches better the default 1W setting of OpenRTX. Roughly checked with NanoVNA.
   platform_set_tx_power(std::min(config->txPower * config->txPower * 100 / (5000 * 5000), 100U)); // crude quadratic mapping of power to duty cycle, max at 5W

platform/drivers/keyboard/keyboard_c62.c
```diff
@@ -0,0 +1,138 @@
...
+
+    // Map raw keys to KEY_ enum values
+    keyboard_t result = 0;
+
+    static const uint32_t keyMap[] = {
+        KEY_ENTER, KEY_UP, KEY_DOWN, KEY_ESC,
+        KEY_3, KEY_4, KEY_5, KEY_6,
+        KEY_7, KEY_8, KEY_9, KEY_STAR,
+        KEY_0, KEY_HASH, 0, 0,  // 14 and 15 are unused
+        KEY_MONI, KEY_F2, KEY_1, KEY_2
```
Marko13-dev, line 101: KEY_MONI should be KEY_F1 here in the keyboard driver?
If you want the KEY_F1 behave the same as KEY_MONI, better change it in ui.c?

platform/targets/c62/platform.c
```diff
@@ -0,0 +1,350 @@
...
+
+int8_t platform_getChSelector()
+{
+    return 0;
+}
+
+bool platform_getPttStatus()
+{
+    //return gpio_pin_get_dt(&button_ptt); // This may brick your radio! Verify what i did in radio_C62.cpp before running this!
+    return false;
```
Marko13-dev, line 281: TX disabled by default, maybe ok for now.

Marko13-dev (commented):
Some comments

silseva:
There are three things that need to be done, in general:

1. Fix the license headers are not correct, as reported by the REUSE CI
2. Format all the new sources according to the new coding style (there is a `clang_format.sh` script in the scripts folder)
3. Remove all the commented-out code

silseva:
Another thing: can you remove the `modules/lib` subfolder? It adds more than 200 files and makes it very difficult to review the code. In general is preferable to not commit libraries and manage them with submodules or through the zephyr manifest

Tunas1337:
> Another thing: can you remove the `modules/lib` subfolder? It adds more than 200 files and makes it very difficult to review the code. In general is preferable to not commit libraries and manage them with submodules or through the zephyr manifest

That's a fantastic question :) I will see if I can figure out a way to do that. The issue is that the additions to the modules/lib subfolder, as far as I can remember, contain a hand-picked selection of everything that was needed from LSF.

silseva:
Actually stale, but not to be closed.

marcoSchr:
The baseband driver for the bk4819 should not be a zephyr specific driver. There are more devices that use this chip and having a generic driver that can be used by all targets would be good. There is also the [Talkpod A36Plus merge request](https://github.com/OpenRTX/OpenRTX/pull/278) where a similar driver is implemented.

We should make sure to have a single implementation for the hardware driver and then do platform specific configuration for the hardware in the platform drivers, similar to how it is already done in the HR C6000 driver.

turnrye:
Hi folks! @OE3ANC has continued working on this, addressing some of the previous feedbacks and preparing the code for review. As I understand it, due to the nature of the reworks he's taken a fresh branch and will be therefore raising a separate PR. Thank you to @Tunas1337 for his work on this PR, for the commits that are the basis for this new platform, and to @OE3ANC for dedicating his time to get this ready for review. The new PR will be linked to this PR to help keep the attribution and provenance chain clear.

PR #417, opened 2026-03-13
M17 packet mode

turnrye:
This is a continuation of my effort to land https://github.com/OpenRTX/OpenRTX/pull/314 in a piecemeal fashion. This PR focuses on adding support for the M17 spec's packet mode. A future PR will finish integrating with the opmode and then finally implement the UIs.

# What

Implement M17 protocol packet mode per the [m17 spec](https://raw.githubusercontent.com/M17-Project/M17_spec/refs/heads/main/M17_spec.tex). These six commits add M17 packet mode transmit and receive to OpenRTX, mirroring the architecture already used for stream mode.

# How

packet mode reuses stream mode's entire FEC/PHY pipeline, but with three spec-mandated differences: the P3 puncture matrix, the 0x75FF sync word, and the 25+1 byte frame structure with EOF/counter metadata replacing stream's LICH+FN structure.

# Why

So that we can implement the SMS feature, and perhaps in the future other M17 packet mode use cases (e.g. APRS-over-M17).

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1129,6 +1129,10 @@ ui_check_standby_test = executable('ui_check_standby_test',
                                     sources : unit_test_src + ['tests/unit/ui_check_standby.cpp'],
                                     kwargs  : unit_test_opts)
 
+m17_packet_test = executable('m17_packet_test',
+                              sources : unit_test_src + ['tests/unit/M17_packet.cpp'],
+                              kwargs  : unit_test_opts)
+
 test('M17 Golay Unit Test',   m17_golay_test)
 test('M17 Viterbi Unit Test', m17_viterbi_test)
 test('M17 Demodulator Test',  m17_demodulator_test)
@@ -1139,4 +1143,4 @@ test('Codeplug Test',         cps_test)
 ## test('Linux InputStream Test', linux_inputStream_test)
 test('minmea conversion Test', minmea_conversion_test)
 test('UI Check Standby Test', ui_check_standby_test)
-
+test('M17 Packet Frame Test', m17_packet_test)
--- a/scripts/clang_format.sh
+++ b/scripts/clang_format.sh
@@ -100,6 +100,7 @@ tests/unit/M17_rrc.cpp
 tests/unit/M17_viterbi.cpp
 tests/unit/ui_check_standby.cpp
 tests/unit/M17_metatext.cpp
+tests/unit/M17_packet.cpp
 EOF
 )
 
--- a/tests/unit/M17_packet.cpp
+++ b/tests/unit/M17_packet.cpp
@@ -1,0 +1,518 @@
+/*
+ * SPDX-FileCopyrightText: Copyright 2020-2026 OpenRTX Contributors
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+#include <catch2/catch_test_macros.hpp>
+#include <cstdint>
+#include <cstring>
+#include <array>
+#include <algorithm>
+#include "protocols/M17/FrameEncoder.hpp"
+#include "protocols/M17/FrameDecoder.hpp"
+#include "protocols/M17/PacketFrame.hpp"
+#include "protocols/M17/Datatypes.hpp"
+#include "protocols/M17/Constants.hpp"
+#include "protocols/M17/LinkSetupFrame.hpp"
+#include "core/crc.h"
+
+using namespace M17;
+
+TEST_CASE("Freshly constructed PacketFrame is zeroed", "[m17][packet]")
+{
+    PacketFrame frame;
+    const uint8_t *raw = frame.data();
+
+    for (size_t i = 0; i < PacketFrame::FRAME_SIZE; i++) {
+        REQUIRE(raw[i] == 0);
+    }
+}
+
+TEST_CASE("Payload provides read/write access", "[m17][packet]")
+{
+    PacketFrame frame;
+
+    // Write a known pattern
+    for (size_t i = 0; i < PacketFrame::DATA_SIZE; i++)
+        frame[i] = static_cast<uint8_t>(i + 1);
+
+    // Read back via data()
+    const uint8_t *raw = frame.data();
+    for (size_t i = 0; i < PacketFrame::DATA_SIZE; i++) {
+        REQUIRE(raw[i] == static_cast<uint8_t>(i + 1));
+    }
+}
+
+TEST_CASE("clear() resets every byte to zero", "[m17][packet]")
+{
+    PacketFrame frame;
+
+    for (size_t i = 0; i < PacketFrame::DATA_SIZE; i++)
+        frame[i] = 0xFF;
+    frame.setEof(true);
+    frame.setCounter(31);
+
+    frame.clear();
+    const uint8_t *raw = frame.data();
+    for (size_t i = 0; i < PacketFrame::FRAME_SIZE; i++) {
+        REQUIRE(raw[i] == 0);
+    }
+}
+
+TEST_CASE("Encode then decode round-trip", "[m17][packet]")
+{
+    FrameEncoder encoder;
+    FrameDecoder decoder;
+
+    // Build a test payload: "HELLO M17 PACKET!!" padded with zeros.
+    PacketFrame frame;
+    const char msg[] = "HELLO M17 PACKET!!";
+    std::copy_n(msg, sizeof(msg) - 1, &frame[0]);
+
+    // Encode
+    frame_t encoded;
+    encoder.encodePacketFrame(frame, encoded);
+
+    // The first two bytes should be the PACKET_SYNC_WORD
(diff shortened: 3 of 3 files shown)


openrtx/include/protocols/M17/M17Datatypes.hpp
```diff
@@ -17,9 +17,10 @@
 namespace M17
 {
 
-using call_t    = std::array< uint8_t, 6 >;    // Data type for encoded callsign
-using payload_t = std::array< uint8_t, 16 >;   // Data type for frame payload field
-using lich_t    = std::array< uint8_t, 12 >;   // Data type for Golay(24,12) encoded LICH data
+using call_t       = std::array< uint8_t, 6 >;    // Data type for encoded callsign
+using payload_t    = std::array< uint8_t, 16 >;   // Data type for frame payload field
+using pktFrame_t   = std::array< uint8_t, 26 >;   // Data type for a full packet frame (25 data + 1 metadata)
```
silseva: Probably we do not really need to declare the `pktFrame_t` type, what do you think?
I was already thinking about cleaning up some of the datatypes listed here, as they are not really needed.

openrtx/include/protocols/M17/M17PacketFrame.hpp
```diff
@@ -0,0 +1,147 @@
...
+ * This class describes and handles a generic M17 packet frame.
+ *
+ * The frame is 26 bytes on the wire: 25 bytes of packet data followed by a
+ * 1-byte metadata field (EOF flag, 5-bit counter, 2 reserved bits).
+ * The payload() accessor exposes only the 25-byte data portion; metadata
+ * is manipulated exclusively through setEof()/isEof()/setCounter()/getCounter().
+ */
+class M17PacketFrame
+{
+    static constexpr uint8_t EOF_BIT = 0x80;
```
silseva: Please move these to the bottom of the class, at line 137.

silseva:
For the packet frame class:
 - let's call it `PacketFrame` without the "M17" prefix: we already did this for the metadata and, in the long term, we decided to remove the "M17" prefix from all the M17-related files and data structures (we already have the namespace for this);
 - instead of having the `payload` method returning a reference to the internal container, I'd implement the `operator []` like in [std::array](https://en.cppreference.com/w/cpp/container/array/operator_at.html). This allows to do `packet[0] = 3;` instead of `packet.payload()[0] = 3;`
 - I'd replace the `getData` method with a `data` one like, again, in [std::array](https://en.cppreference.com/w/cpp/container/array/data.html)
 
I'll do the same with the other M17 data structures to have everything uniform. We have to keep in mind that the current M17 implementation still contains some effects of the quick-and-dirty implementation I did in two weeks during the summer of 2022.

turnrye:
Let's land #420 first and then I can rebase this. But in the meantime feel free to continue to give PR feedback.

turnrye:
I've been doing some manual verification of these changes, and I'd identified a regression in demodulation after the approach that I was taking from some refactoring. To demonstrate this I created a unit test that provided coverage, then updated this implementation to have the test pass. This resulted in a few minor changes to make testing simpler. This again is ready for review.

openrtx/include/protocols/M17/M17Datatypes.hpp
silseva: There is still the indent left over :)

openrtx/include/protocols/M17/PacketFrame.hpp
```diff
@@ -0,0 +1,149 @@
...
+    {
+        return reinterpret_cast<const uint8_t *>(&frameData);
+    }
+
+private:
+    static constexpr uint8_t EOF_BIT = 0x80;
+    static constexpr uint8_t COUNTER_MASK = 0x1F;
+    static constexpr uint8_t COUNTER_SHIFT = 2;
+    static constexpr uint8_t COUNTER_CLEAR =
+        static_cast<uint8_t>(~(COUNTER_MASK << COUNTER_SHIFT));
```
silseva: Is the `static_cast` actually needed?

openrtx/src/protocols/M17/Demodulator.cpp
```diff
@@ -328,36 +331,78 @@ void M17Demodulator::reset()
...
 {
+    // Three synchronizers are checked per sample (LSF, stream, packet).
+    // Each convolve() is O(SYNCW_SIZE) = 8 MACs; the total overhead is ~50
+    // cycles per sample on the slowest supported target.
     int32_t syncThresh = static_cast< int32_t >(corrThreshold * 33.0f);
-    int8_t  syncStatus = streamSync.update(correlator, syncThresh, -syncThresh);
+
+    // Try LSF sync first
+    int8_t syncStatus = lsfSync.update(correlator, syncThresh, -syncThresh);
+
```
silseva: We can remove these extra newlines before the `if` clause. I don't remember why I added them in the beginning, but they do not really make sense.

openrtx/include/protocols/M17/Demodulator.hpp
```diff
@@ -81,6 +81,17 @@ class M17Demodulator
...
+    /**
+     * Process a single raw baseband sample through the full demodulation
+     * chain (DC removal, RRC filtering, clock recovery, correlation and
+     * state machine).  This is the inner loop of update(), exposed so that
+     * unit tests can feed synthetic baseband without the audio subsystem.
+     *
+     * @param rawSample: signed 16-bit baseband sample.
+     * @param invertPhase: invert the phase of the sample before decoding.
+     */
+    void processSample(int16_t rawSample, bool invertPhase = false);
```
silseva: Two things:

1. splitting the code creating a function processing a single sample is an important change and it has to go in a dedicated commit.
2. I'd simply call it `sample`.

openrtx/include/protocols/M17/M17Demodulator.hpp
```diff
@@ -135,6 +146,17 @@ class M17Demodulator
...
 
+    /**
+     * Quantize the correlator syncword region using the given synchronizer
+     * and check the hamming distance against the expected sync word.
+     *
+     * @param sync: synchronizer that detected the candidate.
+     * @param syncWord: expected sync word to compare against.
+     * @return hamming distance between quantized syncword and expected one.
+     */
+    uint8_t trySync(Synchronizer< M17_SYNCWORD_SYMBOLS, SAMPLES_PER_SYMBOL >& sync,
```
silseva: This function has to go above the block of `static constexpr` declarations, for consistency

openrtx/src/protocols/M17/Demodulator.cpp
```diff
@@ -328,36 +331,78 @@ void M17Demodulator::reset()
...
-void M17Demodulator::syncedState()
+uint8_t M17Demodulator::trySync(
+    Synchronizer< M17_SYNCWORD_SYMBOLS, SAMPLES_PER_SYMBOL >& sync,
+    const syncw_t& syncWord)
 {
-    // Set sampling point and deviation, zero frame symbol count
-    samplingPoint  = streamSync.samplingIndex();
-    auto deviation = correlator.maxDeviation(samplingPoint);
-    frameIndex     = 0;
+    uint32_t sp    = sync.samplingIndex();
```
silseva: Why there is `uint32_t sp = sync.samplingIndex();` and then `samplingPoint = sp;`? Can it be just `samplingPoint = sync.samplingIndex();` as it was before?

openrtx/src/protocols/M17/M17Demodulator.cpp
```diff
@@ -328,36 +331,78 @@ void M17Demodulator::reset()
...
+    // If no stream, try packet sync
+    syncStatus = packetSync.update(correlator, syncThresh, -syncThresh);
 
     if(syncStatus != 0)
         demodState = DemodState::SYNCED;
 }
 
-void M17Demodulator::syncedState()
+uint8_t M17Demodulator::trySync(
+    Synchronizer< M17_SYNCWORD_SYMBOLS, SAMPLES_PER_SYMBOL >& sync,
```
silseva: There is no real need to pass a reference to the entire synchronizer when the function only needs it to get the current sampling point: just have `uint8_t M17Demodulator::trySync(const uint32_t samplePoint, const syncw_t& syncWord)`.

Also, I'd call the function `testSyncword` or something like this: we are synchronized and we have to detect if the syncword is valid. `trySync` may be interpreted as "trying to synchronize".

openrtx/src/protocols/M17/Demodulator.cpp
silseva: Actually we can logically or together the three returned values and have a single `if` statement.
I mean
```
int8_t syncStatus = lsfSync.update(correlator, syncThresh, -syncThresh);
syncStatus |= streamSync.update(correlator, syncThresh, -syncThresh);
syncStatus |= packetSync.update(correlator, syncThresh, -syncThresh);
```

turnrye:
Thanks for your review feedback @silseva ! This is ready for review again.

PR #419, opened 2026-03-18
add subproject/catch2 to .gitignore

rxt1077:
This follows the convention used for other subprojects in this repository and makes it a little easier to track while files have changed when committing.


silseva:
Thanks!

PR #421, opened 2026-03-22
voice-prompts: add replacement voice prompt tooling

turnrye:
This change introduces an alternative approach to generating voice prompts. Compared to the current solution, this one has the following benefits:

1. It is a single script colocated with the rest of the project
2. It is entirely FOSS, leveraging kokoro with open source voices (to maximize voice quality, I'm actually blending 3 voices together)
3. It checks to see if the voice prompt file needs to be regenerated and fails the build if so

Bonus, people can use their own voice model with ease and produce fun results that way.

## Verification

Here are a sampling of the before/after sounds:

* Current master with the old tooling: [master.wav](https://github.com/user-attachments/files/32464293/master.wav)
* This PR: [newpr.wav](https://github.com/user-attachments/files/32464311/newpr.wav)

To proxy intelligibility, I took the generated codec2 encoded sounds and ran them through Whisper to try to answer whether the sounds can be understood properly, and I compared the produced transcript vs the actual text. The old approach, which was based on proprietary commercial AWS Polly Brian voice, still scored slightly better, but this approach I've taken is very close.


turnrye:
I've done some _very basic_ testing of this, and I left nearly all of the actual python coding to the LLM agent. Raising this in case there's any interest in a feature focused code review, validating that some of the feature decisions I made are reasonable. The substance of the code itself is prone to change as I review this myself.

turnrye:
I've done some additional testing here and so far, I've been generally pleased with this. It does make the VPC generation tightly coupled to the structures on `voicePrompts.h`, `ui_strings.h`, and `EnglishStrings.h`, but breakages to this will be obvious by unexplained build failures demanding that voice prompts be regenerated. So, tooling would alert us to these sorts of issues, and in the future localization would see this solution get pretty significantly reworked anyway, so we're best to wait until the needs there are a bit clearer and rework it once. I think this still makes a significant step forward and simplifies string management with some manageable risks in the meantime.

This is ready for review.

silseva:
Looks good! Just one note: instead of hardcoding the paths for the string table and VP enum in the python script, I'd pass them by argument.

turnrye:
> Looks good! Just one note: instead of hardcoding the paths for the string table and VP enum in the python script, I'd pass them by argument.

Sounds good -- change made!

```
--voice-prompts-header: path to voicePrompts.h
--ui-strings-header: path to ui_strings.h
--english-strings-header: path to EnglishStrings.h
```

silseva:
> ```
> --voice-prompts-header: path to voicePrompts.h
> --ui-strings-header: path to ui_strings.h
> --english-strings-header: path to EnglishStrings.h
> ```

I think we can entirely skip the `ui_strings.h` file: I see that is being used to retrieve the string "labels" like `const char* languageName`, but these can be retrieved also from the file containing the strings for a given language.

silseva:
@turnrye I know you're still working on this, can I change the status of the PR back to draft until you're done?

turnrye:
> @turnrye I know you're still working on this, can I change the status of the PR back to draft until you're done?

Mmm actually Im not working on this currently! I made the change to consolidate strings a bit (15c977ea4c380468788e1eb6d647ae52044d4ca7) and then it's been ready for review. I didnt ping you again because I was hoping to help steer your attention towards some of the older PRs first.

turnrye:
@silseva this is ready for review. I've merged EnglishStrings.h into ui_strings.h, and so now the voice prompt script is only looking to two places: `voicePrompts.h` and `ui_strings.c` (for the englishStrings initializer).

I still intend in a future localization focused commit to further tidy up strings vs voiceprompts once we're beginning to integrate gettext patterns. But that's a lot more refactoring than we need to do for now, so I'm suggesting we stop here as good enough to unblock voice generatino.

silseva:
Looks good! Since you touched the `ui_strings` files, I'd take the occasion to do a clang-format pass on them. :)

turnrye:
> Looks good! Since you touched the `ui_strings` files, I'd take the occasion to do a clang-format pass on them. :)

Good thinking -- I just added a commit for this and rebased. This is ready for review again.

silseva:
During the Friedrichshafen ham radio fair we asked some people from the german blind hams community to quickly test the voice prompts generated using piper: they said that quality of the speech generated by piper makes them quite difficult to understand. If they have to choose between piper and the speech generator we're currently using, they will go for the second one.

They also suggested some open-source tools for speech generation, @marcoSchr may have a list of them.

turnrye:
> During the Friedrichshafen ham radio fair we asked some people from the german blind hams community to quickly test the voice prompts generated using piper: they said that quality of the speech generated by piper makes them quite difficult to understand. If they have to choose between piper and the speech generator we're currently using, they will go for the second one.
> 
> They also suggested some open-source tools for speech generation, @marcoSchr may have a list of them.

Great to hear that there were some feedbacks, thanks for taking this to the fair! I found https://5uck1ess.github.io/tts-bench/scores.html as a benchmark/eval site. Seems like there are a number of models that could be used that on paper have higher scores. 

Switching it out is not so complex. Unless there are concerns, I can take the top performing voice `Maya1` and jump over to that.

@marcoSchr : Ill wait to hear from you before I take any steps though in case you have any more specific feedback here.

marcoSchr:
@turnrye The main tool that was suggested for voice generation is [Vocello](https://github.com/PowerBeef/QwenVoice). The problem I see with this, is that it is MacOS only. The model underneath is [Qwen3-TTS](https://github.com/QwenLM/Qwen3-TTS).

The other model suggested to us is [Thorsten-Voice](https://www.thorsten-voice.de/en/), which is German only.
This might only be interesting once we have separate voices for different languages.

Alternatively using older tools like espeak-ng was suggested.

We can test espeak-ng as an alternative to piper which you have implemented here.
It might sound better, when being compressed with Codec2.

turnrye:
Hey @marcoSchr and @silseva ! I've reworked this a bit now. Here's where it stands:

- Voice prompts are generated now using kokoro tts, which [benchmarks](https://tts.ai/tts-arena/) right now put at the top in terms of quality (ahead of qwen3-tts that was suggested)
- I've cleaned up a few nits that I found myself from the original PR

I think this is ready again for testing and for review.

turnrye:
Since the previous code reviews, this uses the same pipeline as before for generating the files. But, the model was updated to be kokoro and the voice was set to a 3-way style blend  of bm_lewis:0.34 + bm_fable:0.33 + am_michael:0.33. This is ready for review.

Issue #423, opened 2026-03-24
core/dsp.h uses memset but string.h is not included


rxt1077:
[dsp_resetState()](https://github.com/OpenRTX/OpenRTX/blob/3139f71e26cf0c00ad31973024aac2a63b0df970/openrtx/include/core/dsp.h#L28) uses memset, but string.h is not included. If you use `dsp.h` in another file that doesn't somehow include `string.h/cstring` you'll get an error.

silseva:
PR merged!

PR #425, opened 2026-03-24
core/dsp.h: add string.h

rxt1077:
dsp_resetState() uses memset, but string.h was not included. This adds an include for string.h at the top of core/dsp.h fixing issue #423.


silseva:
Thank you!

PR #426, opened 2026-03-24
Add basic vibrate support for CS7000P

ranguli:
This PR adds vibrate/rumble support for the CS7000P (the only OpenRTX target which has such a feature AFAIK). It exposes `platform_vibrateOn()` and `platform_vibrateOff()` functions which enable and disable the vibration motor respectively. A basic platform test is included as well.

Useful applications could include device vibration when receiving SMS over M17, APRS packets, misc. alerts, or haptic feedback which may also be beneficial for accessibility.

This is my first PR that isn't documentation-related, I _think_ this is implemented correctly from a conventions/architecture standpoint, however any feedback is very welcome!

### DIFF
--- a/openrtx/include/interfaces/platform.h
+++ b/openrtx/include/interfaces/platform.h
@@ -164,6 +164,18 @@ const hwInfo_t *platform_getHwInfo();
 const struct gpsDevice *platform_initGps();
 #endif
 
+#ifdef CONFIG_VIBRATION
+/**
+ * This function turns on the vibration motor.
+ */
+void platform_vibrateOn();
+
+/**
+ * This function turns off the vibration motor.
+ */
+void platform_vibrateOff();
+#endif
+
 #ifdef __cplusplus
 }
 #endif
--- a/platform/targets/CS7000-PLUS/hwconfig.h
+++ b/platform/targets/CS7000-PLUS/hwconfig.h
@@ -66,6 +66,9 @@ extern const struct gpsDevice gps;
 /* Use extended addressing for external flash memory */
 #define CONFIG_W25Qx_EXT_ADDR
 
+/* Device has a vibration motor */
+#define CONFIG_VIBRATION
+
 #ifdef __cplusplus
 }
 #endif
--- a/platform/targets/CS7000-PLUS/pinmap.h
+++ b/platform/targets/CS7000-PLUS/pinmap.h
@@ -159,4 +159,7 @@
 #define ALPU_SDA      GPIOB,9
 #define ALPU_SCL      GPIOB,8
 
+/* Vibration motor */
+#define VIBR_MOTOR    &extGpio,2
+
 #endif /* PINMAP_H */
--- a/platform/targets/CS7000-PLUS/platform.c
+++ b/platform/targets/CS7000-PLUS/platform.c
@@ -206,3 +206,13 @@ const struct gpsDevice *platform_initGps()
 
     return &gps;
 }
+
+void platform_vibrateOn()
+{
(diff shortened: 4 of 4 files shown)


silseva:
The PR is good, there are just a couple of minimal things that need improvement.
Speaking about the feature itself, I'm unsure whether to add it or not to the codebase: at the moment only the CS7000 Plus will have it, and not all the CS7000P devices in the wild actually have the vibration motor. To me, it seems to be more like a "corporate" feature like man-down detection and not something really ham-radio related, but opinions are welcome.
@mdiepart @turnrye what do you think?

turnrye:
> The PR is good, there are just a couple of minimal things that need improvement. Speaking about the feature itself, I'm unsure whether to add it or not to the codebase: at the moment only the CS7000 Plus will have it, and not all the CS7000P devices in the wild actually have the vibration motor. To me, it seems to be more like a "corporate" feature like man-down detection and not something really ham-radio related, but opinions are welcome. @mdiepart @turnrye what do you think?

I welcome the contribution, though it is niche. I hold a special place for the CS7000P as the first commercial radio that ships with OpenRTX. I think it makes sense to give a few extra features where they exist for these special platforms, the same way that we have brought in the entire zephyr stack for ttwr.

Where this may potentially snowball, but its probably in the direction that makes sense to invest: I think this highlights that today we also don't really have a notification or alert tone concept, but this isn't surprising because we don't have things to notify on (e.g. we haven't implemented things like selcall or dug much into specific dst callsign usecases for m17). But I could imagine with SMS RX and APRS RX coming, we may have the first use cases soon that could benefit.

mdiepart:
I second the argument of @turnrye. We should not add all the features the commercial vendors support (most of those makes absolutely no sense for amateur radio operation). However, some of those make sense. 

Also, this is a small but potentially useful contribution, we should merge it because it is not worth debating for days :p

mdiepart:
By the way, is there any features flag in the CS7000p so that we can know at runtime if it has vibrator/bluetooth/... ?

silseva:
> By the way, is there any features flag in the CS7000p so that we can know at runtime if it has vibrator/bluetooth/... ?

It is known at compile time through the `CONFIG_VIBRATE` macro.
I was thinking that, probably, at some point it would be good to implement a subsystem to manage all the extra features, but I'd leave it for the future.

mdiepart:
So nothing in the secure registers of the external flash?

silseva:
Good question. I need to check what does the original firmware do, I never dug that deep into it to see whether the hardware configuration is stored in the external flash or not.

openrtx/include/interfaces/platform.h
```diff
@@ -164,6 +164,19 @@ const hwInfo_t *platform_getHwInfo();
...
+/**
+ * This function turns on the vibration motor.
+ */
+void platform_vibrateOn();
+
+/**
+ * This function turns off the vibration motor.
+ */
+void platform_vibrateOff();
+
```
silseva: This extra line can be trimmed

openrtx/include/interfaces/platform.h
```diff
@@ -164,6 +164,19 @@ const hwInfo_t *platform_getHwInfo();
 const struct gpsDevice *platform_initGps();
 #endif
 
+#ifdef CONFIG_VIBRATE
```
silseva: I'd call it `CONFIG_VIBRATION` to keep it consistent with the other macros defining hardware features (`CONFIG_RTC`, ...).

silseva (changes requested):

PR #427, opened 2026-03-25
M17 packet datalink and rtx support

turnrye:
Implements M17 packet mode's Data Link Layer for receive and transmit, adding the infrastructure needed by higher-level applications (e.g. SMS) to send and receive M17 packet transmissions.

## Provenance

This change is based on the packet / sms opmode work done in https://github.com/OpenRTX/OpenRTX/pull/314 with substantial refactorings to better align with the architecture and the previous feedback that I've received.

### DIFF
--- a/openrtx/include/core/settings.h
+++ b/openrtx/include/core/settings.h
@@ -48,6 +48,7 @@ typedef struct {
     bool showBatteryIcon;      // Battery display true: icon, false: percentage
     bool gpsSetTime;           // Use GPS to ajust RTC time
     char M17_meta_text[53];    // M17 Meta Text to send
+    bool m17_sms_match_call;   // SMS match callsign to receive
 } __attribute__((packed)) settings_t;
 
 static const settings_t default_settings = {
@@ -73,6 +74,7 @@ static const settings_t default_settings = {
     false,     // Display battery icon
     false,     // Update RTC with GPS
     "OpenRTX", // Default M17 meta text
+    true,      // SMS match callsign
 };
 
 #endif /* SETTINGS_H */
--- a/openrtx/include/core/settings.h
+++ b/openrtx/include/core/settings.h
@@ -10,75 +10,69 @@
 #include "hwconfig.h"
 #include <stdbool.h>
 
-typedef enum
-{
-    TIMER_OFF =  0,
-    TIMER_5S  =  1,
-    TIMER_10S =  2,
-    TIMER_15S =  3,
-    TIMER_20S =  4,
-    TIMER_25S =  5,
-    TIMER_30S =  6,
-    TIMER_1M  =  7,
-    TIMER_2M  =  8,
-    TIMER_3M  =  9,
-    TIMER_4M  = 10,
-    TIMER_5M  = 11,
+typedef enum {
+    TIMER_OFF = 0,
+    TIMER_5S = 1,
+    TIMER_10S = 2,
+    TIMER_15S = 3,
+    TIMER_20S = 4,
+    TIMER_25S = 5,
+    TIMER_30S = 6,
+    TIMER_1M = 7,
+    TIMER_2M = 8,
+    TIMER_3M = 9,
+    TIMER_4M = 10,
+    TIMER_5M = 11,
     TIMER_15M = 12,
     TIMER_30M = 13,
     TIMER_45M = 14,
-    TIMER_1H  = 15
-}
-display_timer_t;
+    TIMER_1H = 15
+} display_timer_t;
 
-typedef struct
-{
-    uint8_t brightness;           // Display brightness
-    uint8_t contrast;             // Display contrast
-    uint8_t sqlLevel;             // Squelch level
-    uint8_t voxLevel;             // Vox level
-    int8_t  utc_timezone;         // Timezone, in units of half hours
-    bool    gps_enabled;          // GPS active
-    char    callsign[10];         // Plaintext callsign
-    uint8_t display_timer   : 4,  // Standby timer
-            m17_can         : 4;  // M17 CAN
-    uint8_t vpLevel         : 3,  // Voice prompt level
-            vpPhoneticSpell : 1,  // Phonetic spell enabled
-            macroMenuLatch  : 1,  // Automatic latch of macro menu
-            _reserved       : 3;
-    bool    m17_can_rx;           // Check M17 CAN on RX
-    char    m17_dest[10];         // M17 destination
-    bool    showBatteryIcon;      // Battery display true: icon, false: percentage
-    bool    gpsSetTime;           // Use GPS to ajust RTC time
-    char    M17_meta_text[53];    // M17 Meta Text to send
-}
-__attribute__((packed)) settings_t;
+typedef struct {
+    uint8_t brightness;        // Display brightness
+    uint8_t contrast;          // Display contrast
+    uint8_t sqlLevel;          // Squelch level
+    uint8_t voxLevel;          // Vox level
+    int8_t utc_timezone;       // Timezone, in units of half hours
+    bool gps_enabled;          // GPS active
+    char callsign[10];         // Plaintext callsign
+    uint8_t display_timer : 4, // Standby timer
+        m17_can           : 4; // M17 CAN
+    uint8_t vpLevel       : 3, // Voice prompt level
+        vpPhoneticSpell   : 1, // Phonetic spell enabled
+        macroMenuLatch    : 1, // Automatic latch of macro menu
+        _reserved         : 3;
+    bool m17_can_rx;           // Check M17 CAN on RX
+    char m17_dest[10];         // M17 destination
+    bool showBatteryIcon;      // Battery display true: icon, false: percentage
+    bool gpsSetTime;           // Use GPS to ajust RTC time
+    char M17_meta_text[53];    // M17 Meta Text to send
+} __attribute__((packed)) settings_t;
 
-
-static const settings_t default_settings =
-{
-    100,                          // Brightness
+static const settings_t default_settings = {
+    100,                     // Brightness
 #ifdef CONFIG_SCREEN_CONTRAST
-    CONFIG_DEFAULT_CONTRAST,      // Contrast
+    CONFIG_DEFAULT_CONTRAST, // Contrast
 #else
-    255,                          // Contrast
+    255, // Contrast
 #endif
-    4,                            // Squelch level, 4 = S3
-    0,                            // Vox level
-    0,                            // UTC Timezone
-    false,                        // GPS enabled
-    "",                           // Empty callsign
-    TIMER_30S,                    // 30 seconds
-    0,                            // M17 CAN
-    0,                            // Voice prompts off
-    0,                            // Phonetic spell off
-    1,                            // Automatic latch of macro menu enabled
-    0,                            // not used
-    false,                        // Check M17 CAN on RX
-    "",                           // Empty M17 destination
-    false,                        // Display battery icon
-    false,                        // Update RTC with GPS
-    "OpenRTX",                    // Default M17 meta text
+    4,         // Squelch level, 4 = S3
+    0,         // Vox level
+    0,         // UTC Timezone
+    false,     // GPS enabled
+    "",        // Empty callsign
+    TIMER_30S, // 30 seconds
+    0,         // M17 CAN
+    0,         // Voice prompts off
+    0,         // Phonetic spell off
+    1,         // Automatic latch of macro menu enabled
+    0,         // not used
+    false,     // Check M17 CAN on RX
+    "",        // Empty M17 destination
+    false,     // Display battery icon
+    false,     // Update RTC with GPS
+    "OpenRTX", // Default M17 meta text
 };
 
 #endif /* SETTINGS_H */
--- a/openrtx/include/protocols/M17/PacketDeframer.hpp
+++ b/openrtx/include/protocols/M17/PacketDeframer.hpp
@@ -1,0 +1,90 @@
+/*
+ * SPDX-FileCopyrightText: Copyright 2020-2026 OpenRTX Contributors
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+#ifndef M17_PACKETDEFRAMER_H
+#define M17_PACKETDEFRAMER_H
+
+#ifndef __cplusplus
+#error This header is C++ only!
+#endif
+
+#include "PacketFrame.hpp"
+#include "PacketFramer.hpp"
+#include <cstddef>
+#include <cstdint>
+
+namespace M17
+{
+
+/**
+ * Result codes returned by PacketDeframer::pushFrame().
+ */
+enum class DeframerResult {
+    IN_PROGRESS,  ///< Frame accepted, waiting for more
+    COMPLETE,     ///< Final frame accepted, CRC valid — packet ready
+    ERR_CRC,      ///< Final frame received but CRC check failed
+    ERR_OVERFLOW, ///< Frame would exceed buffer capacity
+    ERR_SEQUENCE  ///< Frame counter does not match expected sequence
+};
+
+/**
+ * Stateful deframer for M17 Data Link Layer packet mode.
+ *
+ * Receives PacketFrames one at a time (as they arrive from the decoder),
+ * validates sequential counters, accumulates payload data, and verifies
+ * the trailing CRC on the final frame.  Protocol-ID-agnostic — works for
+ * any M17 packet type.
+ */
+class PacketDeframer
+{
+public:
+    PacketDeframer();
+
+    /**
+     * Reset the deframer to accept a new packet from frame 0.
+     */
+    void reset();
+
+    /**
+     * Feed the next received PacketFrame into the reassembly buffer.
+     *
+     * @param frame: the PacketFrame from the decoder.
+     * @return result code indicating reassembly status.
+     */
+    DeframerResult pushFrame(const PacketFrame &frame);
+
+    /**
+     * Get a pointer to the reassembled packet data (valid after COMPLETE).
+     * The returned data does NOT include the trailing 2-byte CRC.
+     *
+     * @return pointer to the internal buffer.
+     */
+    const uint8_t *data() const
+    {
+        return buffer;
+    }
+
+    /**
+     * Get the length of the reassembled packet data (excluding CRC).
+     * Valid after pushFrame() returns COMPLETE.
+     *
+     * @return number of application-data bytes.
+     */
+    size_t length() const
+    {
+        return dataLen;
+    }
+
+private:
+    uint8_t buffer[MAX_PACKET_DATA];
+    size_t totalLen; ///< Bytes accumulated so far (including CRC area)
+    size_t dataLen;  ///< Application data length (totalLen - 2) after CRC check
+    uint8_t nextCounter; ///< Expected frame sequence number
+};
+
+} // namespace M17
+
+#endif /* M17_PACKETDEFRAMER_H */
--- a/openrtx/include/protocols/M17/PacketDeframer.hpp
+++ b/openrtx/include/protocols/M17/PacketDeframer.hpp
@@ -1,0 +1,110 @@
+/*
+ * SPDX-FileCopyrightText: Copyright 2020-2026 OpenRTX Contributors
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+#ifndef M17_PACKETDEFRAMER_H
+#define M17_PACKETDEFRAMER_H
+
+#ifndef __cplusplus
+#error This header is C++ only!
+#endif
+
+#include "PacketFrame.hpp"
+#include <cstddef>
+#include <cstdint>
+
+namespace M17
+{
+
+/**
+ * Result codes returned by PacketDeframer::pushFrame().
+ */
+enum class DeframerResult {
+    IN_PROGRESS,      ///< Frame accepted, waiting for more
+    COMPLETE,         ///< Final frame accepted, CRC valid — packet ready
+    ERR_CRC,          ///< Final frame received but CRC check failed
+    ERR_OVERFLOW,     ///< Frame would exceed buffer capacity
+    ERR_SEQUENCE,     ///< Frame counter does not match expected sequence
+    ERR_INVALID_FRAME ///< EOF counter is 0 or exceeds DATA_SIZE
+};
+
+/**
+ * Stateful deframer for M17 Data Link Layer packet mode.
+ *
+ * Receives PacketFrames one at a time (as they arrive from the decoder),
+ * validates sequential counters, accumulates payload data, and verifies
+ * the trailing CRC on the final frame.  Protocol-ID-agnostic — works for
+ * any M17 packet type.
+ *
+ * Before pushing frames, the caller must supply an external receive buffer
+ * via init().  The deframer writes directly into that buffer (zero-copy).
+ */
+class PacketDeframer
+{
+public:
+    PacketDeframer();
+
+    /**
+     * Bind the deframer to a caller-supplied receive buffer.
+     *
+     * Must be called before the first pushFrame() for each new packet.
+     * reset() preserves the bound buffer so that init() need only be called
+     * once per descriptor.
+     *
+     * @param buffer:   Writable buffer to assemble the packet into.
+     * @param capacity: Size of buffer in bytes.  Must be large enough for
+      *                  the full packet payload (up to M17_MAX_PACKET_DATA bytes).
+     * @return true on success; false if buffer is null or capacity is zero.
+     */
+    bool init(uint8_t *buffer, size_t capacity);
+
+    /**
+     * Reset the deframer to accept a new packet from frame 0.
+     * The bound buffer pointer and capacity are preserved.
+     */
+    void reset();
+
+    /**
+     * Feed the next received PacketFrame into the reassembly buffer.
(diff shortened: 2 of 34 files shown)


openrtx/include/core/settings.h
```diff
@@ -47,6 +47,7 @@ typedef struct
             macroMenuLatch  : 1,  // Automatic latch of macro menu
             _reserved       : 3;
     bool    m17_can_rx;           // Check M17 CAN on RX
+    bool    m17_sms_match_call;   // SMS match callsign to receive
```
silseva: This has to go to the bottom of the data structure, otherwise it breaks all the existing settings

turnrye: Fixed!

openrtx/include/rtx/OpMode_M17.hpp
```diff
@@ -109,17 +133,14 @@ class OpMode_M17 : public OpMode
...
+     * Function handling packet TX (SMS send).
      *
-     * \param localCs plain text callsign from the user
-     * \param incomingCs plain text destination callsign
-     * \return true if local an incoming callsigns match.
+     * @param status: pointer to the rtxStatus_t structure containing the
+     * current RTX status.
      */
-    bool compareCallsigns(const std::string& localCs, const std::string& incomingCs);
+#ifdef CONFIG_M17
```
silseva, line 141:
```suggestion
```

openrtx/include/rtx/OpMode_M17.hpp
```diff
@@ -109,17 +133,14 @@ class OpMode_M17 : public OpMode
...
+    void txPacketState(rtxStatus_t *const status);
+#endif
```
silseva, line 143:
```suggestion
```

openrtx/src/protocols/M17/PacketAssembler.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+    totalLen = dataLen + 2;
+    if (totalLen > MAX_PACKET_DATA)
+        return false;
+
+    src = buffer;
+
+    // Compute CRC over application data, append big-endian.
+    uint16_t crc = crc_m17(buffer, dataLen);
+    buffer[dataLen] = static_cast<uint8_t>(crc >> 8);
```
silseva, line 32: This is at risk of causing a buffer overflow: the buffer passed as input parameter can be exactly `dataLen` bytes long. It'd be better to write the CRC in a private variable and append it when composing the last block.

openrtx/src/protocols/M17/PacketAssembler.cpp
```diff
@@ -0,0 +1,66 @@
...
+    buffer[dataLen + 1] = static_cast<uint8_t>(crc & 0xFF);
+
+    numFrames = (totalLen + PacketFrame::DATA_SIZE - 1)
```
silseva, line 35: You need to take into account also `sizeof(uint16_t)` for the two bytes of CRC, otherwise the message can get truncated.

openrtx/src/rtx/OpMode_M17.cpp
```diff
@@ -154,6 +186,16 @@ void OpMode_M17::offState(rtxStatus_t *const status)
         return;
     }
 
+#ifdef CONFIG_M17
+    if(state.havePacketData)
+    {
+        startTx = true;
+        status->opStatus = TX;
+        status->txDisable = 0;
```
silseva, line 194:
```suggestion
```

Never override `txDisable`, is a safety feature :)

openrtx/src/rtx/OpMode_M17.cpp
```diff
@@ -269,13 +311,72 @@ void OpMode_M17::rxState(rtxStatus_t *const status)
...
+#ifdef CONFIG_M17
+                // Check if packet frame and SMS receive enabled
+                else if(type == FrameType::PACKET &&
+                        (canMatch == true) &&
+                        ((callMatch == true) || !state.settings.m17_sms_match_call))
+                {
+                    PacketFrame pf = decoder.getPacketFrame();
+
+                    // Start reassembly on first frame with SMS protocol ID
+                    if(!pktStarted && pf[0] == 0x05)
```
silseva, line 323: Actually this code should be generic and reassemble any packet. Check on content type will be done later.

openrtx/src/rtx/OpMode_M17.cpp
```diff
@@ -394,25 +496,91 @@ void OpMode_M17::txState(rtxStatus_t *const status)
     }
 }
 
-bool OpMode_M17::compareCallsigns(const std::string& localCs,
-                                  const std::string& incomingCs)
+#ifdef CONFIG_M17
```
silseva, line 499:
```suggestion
```

openrtx/src/rtx/OpMode_M17.cpp
```diff
@@ -394,25 +496,91 @@ void OpMode_M17::txState(rtxStatus_t *const status)
...
+    modulator.stop();
+
+    startRx = true;
+    state.havePacketData = false;
+    memset(state.sms_message, 0, sizeof(state.sms_message));
+    lastCRC = 0;
+    status->txDisable = 1;
+    status->opStatus = OFF;
 }
+#endif
```
silseva, line 586:
```suggestion
```

silseva:
Probably we can rename `PacketAssembler` and `PacketDisassembler` classes into `PacketFramer` and `PacketDeframer` to be more adherent to the OSI model.

openrtx/src/protocols/M17/PacketAssembler.cpp
turnrye: L25 already handles this with the +2

turnrye:
Hey @silseva this is ready for review again! Thanks!

silseva:
When applying clang-format, please do it in dedicated commits. Otherwise the changes become too much and is very difficult to review the code... :sweat_smile:

turnrye:
> When applying clang-format, please do it in dedicated commits. Otherwise the changes become too much and is very difficult to review the code... 😅

Sure thing.

I also think in general we may be to the point where we should just clang format everything, I think we're past most of the risks for conflicts at this point. I can do this in a separate MR next week if you agree.

silseva:
> Sure thing.

Thank you!
 
> I also think in general we may be to the point where we should just clang format everything, I think we're past most of the risks for conflicts at this point. I can do this in a separate MR next week if you agree.

I think is better to still do the code style update progressively: I noticed that clang-format tend to mess up things badly if, for example, there is a missing comma or minimal things like this. We still need to fine-tune the formatting rules as well.

silseva:
You didn't push the commit :)

turnrye:
> You didn't push the commit :)

Oops! It's here now.

openrtx/include/rtx/rtx.h
```diff
@@ -87,7 +87,10 @@ enum opstatus {
  */
 enum pktStatus {
     PKT_STATUS_IDLE,      /**< No operation in progress */
-    PKT_STATUS_SUBMITTED, /**< Submitted to RX/TX queue */
+    PKT_STATUS_SUBMITTED, /**< Reserved; not currently set by any OpMode.
```
silseva: The status is changed to "submitted" here: https://github.com/OpenRTX/OpenRTX/blob/80ecc7d1bafcd48d44a31e4157e2798b1aab6c0f/openrtx/src/rtx/rtx.cpp#L211

When a packet descriptor is set to this state it means that it has been acquired by the rtx subsystem and its fields should not be touched by the application layer.

turnrye: Oops, fixed!

openrtx/src/protocols/M17/PacketDeframer.cpp
```diff
@@ -0,0 +1,75 @@
...
+        memcpy(&buffer[totalLen], frame.data(), PacketFrame::DATA_SIZE);
+        totalLen += PacketFrame::DATA_SIZE;
+        nextCounter++;
+
+        return DeframerResult::IN_PROGRESS;
+    }
+
+    // EOF frame: counter holds the number of valid bytes in this chunk
+    uint8_t lastBytes = counter;
+    if (lastBytes == 0 || lastBytes > PacketFrame::DATA_SIZE)
```
silseva: Aren't those error conditions? Here the cases are either a packet with zero payload (kinda acceptable, but not correct) or a packet with more than 25 bytes of payload, which is definitely an error.

openrtx/src/protocols/M17/PacketDeframer.cpp
```diff
@@ -0,0 +1,75 @@
...
+    totalLen += lastBytes;
+
+    // Need at least 3 bytes: 1 byte of data + 2 bytes CRC
+    if (totalLen < 3)
+        return DeframerResult::ERR_CRC;
+
+    // CRC is over all bytes except the trailing 2-byte CRC itself,
+    // stored big-endian at the end of the packet.
+    uint16_t computed = crc_m17(buffer, totalLen - 2);
+    uint16_t stored = (static_cast<uint16_t>(buffer[totalLen - 2]) << 8)
```
silseva: I think you can safely omit type casts:

```suggestion
    uint16_t stored = (buffer[totalLen - 2] << 8) | buffer[totalLen - 1];
```

openrtx/include/protocols/M17/PacketDeframer.hpp
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * Stateful deframer for M17 Data Link Layer packet mode.
+ *
+ * Receives PacketFrames one at a time (as they arrive from the decoder),
+ * validates sequential counters, accumulates payload data, and verifies
+ * the trailing CRC on the final frame.  Protocol-ID-agnostic — works for
+ * any M17 packet type.
+ */
+class PacketDeframer
```
silseva, line 44: What about adding an `init` method like the in the `PacketFramer` class providing an external buffer? This can be a nice way to have zero-copy packet deframing.

openrtx/include/rtx/OpMode_M17.hpp
```diff
@@ -119,27 +145,37 @@ class OpMode_M17 : public OpMode
      * \param incomingCs plain text destination callsign
      * \return true if local an incoming callsigns match.
      */
-    bool compareCallsigns(const std::string& localCs, const std::string& incomingCs);
+    bool compareCallsigns(const std::string &localCs,
```
silseva: I think this function is now unused: if so, it can be removed

openrtx/src/protocols/M17/PacketFramer.cpp
```diff
@@ -0,0 +1,89 @@
...
+    // Compute CRC over application data and store internally.
+    crc = crc_m17(buffer, len);
+
+    offset = 0;
+    counter = 0;
+
+    return true;
+}
+
+bool PacketFramer::nextFrame(PacketFrame &frame)
```
silseva, line 40: I'd make this function return true on when the last frame has been prepared, so that caller code knows immediately that at the next round there will be need for a new init.

openrtx/src/protocols/M17/SmsPacket.cpp
```diff
@@ -0,0 +1,56 @@
...
+                         size_t bufferSize)
+{
+    if (msgLen == 0 || message == nullptr || buffer == nullptr)
+        return 0;
+
+    // SMS application format: [0x05][UTF-8 text][0x00]
+    const size_t appDataLen = 1 + msgLen + 1;
+
+    // Buffer needs room for app data; CRC is stored internally by PacketFramer
+    if (appDataLen > bufferSize || appDataLen + 2 > M17::MAX_PACKET_DATA)
```
silseva: From what I see in the `PacketFrame` class the "CRC is stored internally by PacketFramer" is not the case anymore

openrtx/src/rtx/OpMode_M17.cpp
```diff
@@ -141,28 +156,31 @@ void OpMode_M17::offState(rtxStatus_t *const status)
...
 
-    if(platform_getPttStatus() && (status->txDisable == 0))
-    {
+    if (platform_getPttStatus() && (status->txDisable == 0)) {
         startTx = true;
         status->opStatus = TX;
         return;
     }
 
+    if (currTxPkt != nullptr) {
```
silseva, line 163: This condition overrides the `txDisable` flag, not good

openrtx/src/rtx/OpMode_M17.cpp
```diff
@@ -242,52 +253,93 @@ void OpMode_M17::rxState(rtxStatus_t *const status)
...
                     StreamFrame sf = decoder.getStreamFrame();
-                    codec_pushFrame(sf.data(),     false);
+                    codec_pushFrame(sf.data(), false);
                     codec_pushFrame(sf.data() + 8, false);
                 }
+                // Handle incoming packet frame — content-agnostic
+                else if (type == FrameType::PACKET && canMatch) {
+                    PacketFrame pf = decoder.getPacketFrame();
+
+                    if (!pktStarted) {
```
silseva: I think you can use `currRxPkt == nullptr` as a condition in place of  `pktStarted`. If the RX queue is empy, there is no point in decoding a full incoming packed to then just drop it because there is no descriptor ready to receive it.

openrtx/src/rtx/OpMode_M17.cpp
```diff
@@ -386,32 +435,120 @@ void OpMode_M17::txState(rtxStatus_t *const status)
...
+    currTxPkt = packet;
+    pthread_mutex_unlock(&txMutex);
+    return 0;
+}
+
+void OpMode_M17::txPacketState(rtxStatus_t *const status)
+{
+    frame_t m17Frame;
+
+    demodulator.stopBasebandSampling();
```
silseva: No need for this, the demodulator is already stopped when transitioning to OFF state from RX

openrtx/src/rtx/OpMode_M17.cpp
```diff
@@ -386,32 +435,120 @@ void OpMode_M17::txState(rtxStatus_t *const status)
...
+
+    // Initialise framer with caller-supplied data
+    if (!pktFramer.init(static_cast<uint8_t *>(currTxPkt->buffer),
+                        currTxPkt->size)) {
+        pthread_mutex_lock(&txMutex);
```
silseva: Probably you don't need a critical section here: at this point you're owning a packet descriptor and you can flip the pointer back to `nullptr` without anyone else stomping you.

openrtx/src/rtx/OpMode_M17.cpp
```diff
@@ -386,32 +435,120 @@ void OpMode_M17::txState(rtxStatus_t *const status)
...
+
+    radio_enableTx();
+    modulator.invertPhase(invertTxPhase);
+    modulator.start();
+    modulator.sendPreamble();
+    modulator.sendFrame(m17Frame);
+
+    // Send packet frames
+    PacketFrame pf;
+    while (pktFramer.nextFrame(pf)) {
```
silseva: IGH! Doing so you're blocking all the rest of the RTX thread and state machine for up to 25*40ms = 1s!!
This should be an enter-send one-exit function.

silseva (dismissed):

silseva:
@turnrye can you rebase on top of the current `release-v0.5.0` branch and force-push, please?

silseva:
Also, can you please address the two review comments that are still open? :smile:

openrtx/src/rtx/OpMode_M17.cpp
turnrye: Fixed!

openrtx/src/rtx/OpMode_M17.cpp
turnrye: When you originally flagged the concern about txDisable a few months ago, this is what prompted me to raise `47a6ada7` to split txDisable into txDisable and pttDisable. The packet path must skip pttDisable by design; it's set whenever the user is in a menu, which is exactly when SMS is composed. Requiring pttDisable == 0 would make the feature permanently non-functional, make us weaken the pttDisable definition, or force the packet tx to take place on the vfo screen. 

I've added an explicit early-failure check here for txDisable so the caller gets a real error instead of a silent completion.

Hopefully this explanation + these other guards helps address your concerns! If not please help me see what some other alternatives would be.

turnrye:
I've rebased this, plus addressed both of your feedbacks. I've also done some basic manual testing to make sure the full feature still works (i.e. this PR + future commits not yet raised) and things are still good. So, this is ready for review again @silseva !

openrtx/include/rtx/rtx.h
```diff
@@ -101,6 +104,8 @@ struct pktDesc {
     void *buffer;
     size_t size;
     ssize_t res;
+    char destination
```
silseva: I'm not sure about this: `pktDesc` is meant to be a generic packet descriptor, if you add the `destination` field (with a fixed length of 10 elements) you're forcing an M17-specific part in it.
In my mind all the opmode-specific parts should be set using `rtx_configure`, leaving the descriptor only to packet payload. I agree that using `rtx_configure` is a bit cumbersome, I plan to add soon another function allowing to set/update only the opmode-specific configuration.

openrtx/include/rtx/rtx.h
```diff
@@ -130,6 +135,15 @@ void rtx_configure(const rtxStatus_t *cfg);
...
 rtxStatus_t rtx_getCurrentStatus();
 
+/**
+ * Obtain a read-only pointer to the RTX driver's internal status.
+ * The pointer is valid for the lifetime of the process.  Callers must
+ * not write through it.  No copy is made; suitable for use on threads
+ * with small stacks (e.g. the 512-byte embedded RTX thread).
+ * @return pointer to the RTX driver's internal status data structure.
+ */
+const rtxStatus_t *rtx_getStatus();
```
silseva, line 138: I'd replace `rtx_getCurrentStatus` with this function rather than just adding it and having a mixed behaviour around the codebase. Also, this should go in a dedicated commit.

openrtx/src/rtx/rtx.cpp
```diff
@@ -88,9 +88,15 @@ void rtx_configure(const rtxStatus_t *cfg)
...
 }
 
-rtxStatus_t rtx_getCurrentStatus()
+const rtxStatus_t *rtx_getStatus()
 {
-    return rtxStatus;
+    return &rtxStatus;
+}
+
+void rtx_setPacketDestination(const char *dst)
```
silseva: I'm sorry, but this is not good. This function allows any thread to write directly into `rtxStatus` without any concurrency lock: very bad things can happen.
Can't application code set the packet destination address via `rtx_configure`, at least for the time being?

silseva: As discussed on discord, the best solution would be to embed the source and destination callsigns alongside the payload:
```
struct m17Packet __attribute__((packed)) {
  char src[10];
  char dst[10];
  uint8_t payload[MAX_PACKET_DATA];
};
```
Sending a frame will become:
```
struct m17Packet packet;
struct pktDesc txDesc;

packet.src = "N0CALL";
packet.dst = "TE5ST";
memcpy(packet.payload, "something", strlen("something"));

txDesc.status = PKT_STATUS_IDLE;
txDesc.buffer = &packet;
txDesc.size = 20 + strlen("something")
```

silseva (changes requested):

openrtx/include/rtx/rtx.h
```diff
@@ -169,4 +176,26 @@ int rtx_addPacketTx(struct pktDesc *packet);
...
+ *
+ * On RX, the deframer writes directly into the payload region.  The
+ * src/dst fields are populated from the M17 LSF by the OpMode layer.
+ */
+struct m17Packet {
+    char src[10]; /**< Source callsign (10 bytes, null-padded) */
+    char dst[10]; /**< Destination callsign (10 bytes, null-padded) */
+    uint8_t payload[M17_MAX_PACKET_DATA]; /**< Packet payload data */
+} __attribute__((packed));
+
```
silseva: I'd move this to an `m17.h` header file in `protocols/M17`. This header file will be useful for me as well, when I'll refactor the rtx state data structure moving all the M17-related fields to dedicated data structures. The idea is to have the application code, whether C or C++, include `m17.h` and get access to all the data structures needed to interface with the rtx layer.
With this move, I'd remove the `M17_MAX_PACKET_DATA` field from `M17PacketFrame.hpp` and include the header: this avoids the risk of updating one field and not the other.

openrtx/src/rtx/OpMode_M17.cpp
```diff
@@ -256,13 +282,52 @@ void OpMode_M17::rxState(rtxStatus_t *const status)
...
+                        } else {
+                            pktDeframer.init(nullptr, 0);
+                        }
+                    }
+
+                    auto result = pktDeframer.pushFrame(pf);
+
+                    if (result == DeframerResult::COMPLETE) {
+                        if (currRxPkt != nullptr) {
+                            // Zero-copy: data already in currRxPkt->buffer
```
silseva: Here we can set the `src` and `dst` fields of the last received packet.

openrtx/src/rtx/rtx.cpp
```diff
@@ -238,7 +238,7 @@ int rtx_addPacketTx(struct pktDesc *packet)
     if (packet->status != PKT_STATUS_IDLE)
         return -EPERM;
 
-    if ((packet->buffer == NULL) || (packet->size == 0))
+    if ((packet->buffer == NULL) || (packet->size < 20))
```
silseva: Please revert this change and eventually move the check to `OpMode_M17::addPacketTx`. Minimum packet size depends on the specific opmode, while this function is generic.
Also: you are setting a minimum limit of 20 bytes on a TX packet descriptor, but not in an RX one: it should be the same for both, as we'd like to set the `src` and `dst` fields of a received packet.

openrtx/include/protocols/M17/PacketFrame.hpp
```diff
@@ -154,6 +155,13 @@ class PacketFrame
     friend class FrameDecoder;
 };
 
+/**
+ * Maximum number of bytes in an M17 Single Packet (application data + CRC).
+ * Per the spec, up to 823 bytes of application data plus 2 bytes of CRC,
+ * yielding at most 33 frames of 25 bytes each.
+ */
+static constexpr size_t MAX_PACKET_DATA = M17_MAX_PACKET_DATA;
```
silseva: I guess we can just use `M17_MAX_PACKET_DATA` everywhere in the class and get rid of this declaration. :)

turnrye: Sounds good, I'd wondered what you'd say after the new header was introduced :)

openrtx/include/rtx/rtx.h
```diff
@@ -169,4 +176,6 @@ int rtx_addPacketTx(struct pktDesc *packet);
 }
 #endif
 
+#include "../protocols/M17/m17.h"
+
```
silseva: Please, no! 😿 
Seriously, I think that this is not needed anymore: application code can `#include "protocols/M17/m17.h" when needed.

openrtx/src/protocols/M17/PacketDeframer.cpp
```diff
@@ -0,0 +1,94 @@
...
+bool PacketDeframer::init(uint8_t *buf, size_t cap)
+{
+    buffer = nullptr;
+    capacity = 0;
+    totalLen = 0;
+    dataLen = 0;
+    nextCounter = 0;
+
+    if (buf == nullptr || cap == 0)
+        return false;
```
silseva: These two lines should go at the beginning of the function, to avoid having an half-baked `PaketDeframer` in case of errors.

openrtx/src/protocols/M17/PacketDeframer.cpp
```diff
@@ -0,0 +1,94 @@
...
+        memcpy(&buffer[totalLen], frame.data(), PacketFrame::DATA_SIZE);
+        totalLen += PacketFrame::DATA_SIZE;
+        nextCounter++;
+
+        return DeframerResult::IN_PROGRESS;
+    }
+
+    // EOF frame: counter holds the number of valid bytes in this chunk.
+    // A counter of 0 or greater than DATA_SIZE is malformed.
+    uint8_t lastBytes = counter;
```
silseva: I'd ust use `counter` everywhere

openrtx/src/protocols/M17/PacketDeframer.cpp
turnrye: So, making this change results in the `PacketDeframer: init(nullptr,0) after COMPLETE unbinds buffer (M17_packet_disassembly.cpp:426-451)` unit test failing.

I think what's happening there is: by moving the early return to the top before the buffer is cleared leaves a stale pointer, which allows pushFrame to write into the previous packet's buffer. The original version here avoids that because it clears all fields before validating. But if you'd prefer the guard at the top, I can restructure it so that the validation is first but then the unbinding is also maintained. That would look like this:
```
bool PacketDeframer::init(uint8_t *buf, size_t cap)
{
    if (buf == nullptr || cap == 0)
    {
        buffer    = nullptr;
        capacity  = 0;
        totalLen  = 0;
        dataLen   = 0;
        nextCounter = 0;
        return false;
    }

    buffer    = buf;
    capacity  = cap;
    totalLen  = 0;
    dataLen   = 0;
    nextCounter = 0;
    return true;
}
```

I'm not making this change for the moment, as I wanted to know what your thoughts where first.

silseva: Let's leave it as it is now.

PR #428, opened 2026-03-25
Swap KEY_MONI and KEY_F1 behaviours

ranguli:
This PR is a prerequisite for addressing issue [#797](https://tasks.openrtx.org/project/openrtx/issue/797) / user story [#804](https://tasks.openrtx.org/project/openrtx/us/804). In order to add functionality that opens the squelch when the MONI key is held, the MONI key first needs to be unbound as it is currently used for the macro menu.

## Current behaviour

- `KEY_MONI` activates the macro menu for as long as it is held. If the accessibility option **Macro Latch** is selected, then the macro menu will stay open after `KEY_MONI` has been held for a certain amount of time. Pressing `KEY_MONI` again will hide the macro menu. 
- `KEY_F1` when momentary pressed will replay the last voice prompt, if voice prompts are enabled.
  - Short-pressing `KEY_F1` again while a voice prompt is playing will stop the voice prompt playback.

## Proposed behaviour:
(Based on discussion from OpenRTX office hours April 2026)

- `KEY_F1` activates the macro menu for as long as it is held. If the accessibility option **Macro Latch** is selected, then the macro menu will stay open aftey `KEY_F1` has been held for a certain amount of time. Pressing `KEY_F1` again will hide the macro menu.
- A short press of `KEY_MONI` will replay the last voice prompt, if voice prompts are enabled.
  - Short-pressing `KEY_MONI` again while a voice prompt is playing will stop the voice prompt playback. 

## To-Do

- [x] Remap the macro menu to `KEY_F1`
- [x] Remap the "replay last voice prompt" feature to `KEY_MONI` (only for short presses and only when voice prompts are enabled).
- [ ] Device testing

### DIFF
--- a/openrtx/src/ui/default/ui.c
+++ b/openrtx/src/ui/default/ui.c
@@ -1411,24 +1411,24 @@ void ui_updateFSM(bool *sync_rtx)
         bool f1Handled = false;
         vpQueueFlags_t queueFlags = vp_getVoiceLevelQueueFlags();
         // If we get out of standby, we ignore the kdb event
-        // unless is the MONI key for the MACRO functions
-        if (_ui_exitStandby(now) && !(msg.keys & KEY_MONI))
+        // unless is the F1 key for the MACRO functions
+        if (_ui_exitStandby(now) && !(msg.keys & KEY_F1))
             return;
-        // If MONI is pressed, activate MACRO functions
-        bool moniPressed = msg.keys & KEY_MONI;
-        if(moniPressed || macro_latched)
+        // If the F1 key is pressed, activate MACRO functions
+        bool f1Pressed = msg.keys & KEY_F1;
+        if(f1Pressed || macro_latched)
         {
             macro_menu = true;
 
             if(state.settings.macroMenuLatch == 1)
             {
-                // long press moni on its own latches function.
-                if (moniPressed && msg.long_press && !macro_latched)
+                // long press KEY_F1 on its own latches function.
+                if (f1Pressed && msg.long_press && !macro_latched)
                 {
                     macro_latched = true;
                     vp_beep(BEEP_FUNCTION_LATCH_ON, LONG_BEEP);
                 }
-                else if (moniPressed && macro_latched)
+                else if (f1Pressed && macro_latched)
                 {
                     macro_latched = false;
                     vp_beep(BEEP_FUNCTION_LATCH_OFF, LONG_BEEP);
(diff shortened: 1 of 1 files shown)


silseva:
Did you test it with voice prompts as well? Currently pressing F1 replays the latest voice prompt.

ranguli:
@silseva Ah I was unaware of that feature! I will test it and see :+1:

ranguli:
@silseva I can confirm that my changes do currently break that feature and pressing F1 no longer plays the voice prompt. We could have F2 replay the last voice prompt instead, but based on the [research](https://gist.github.com/ranguli/d152d15e32ddff684a5b02d7393312af) I did into the number of buttons on OpenRTX targets, not every radio has F1, MONI and F2 keys. 

What would be the best way to proceed? I think MONI is a pretty important basic function but we of course can't break any accessibility features :slightly_smiling_face:

silseva:
This needs a proper discussion with the rest of the developers. We also have to standardize among all the radios which of the side keys is assigned to which functionality.

ranguli:
Understandable, that sounds good! :+1:

lz2sat:
One possible solution is for the moni button to be a monitor key, and the F1 (F2 when present) buttons to be software-defined as to what function they should have from the menu, depending on the user's needs (as is the case in some models). Another possible solution, also found in some manufacturers, is the possibility for one button to have 2 functions (short press, function 1, for example, voice prompt, hold for 2 seconds, long press, function 2, function button or menu access), of course, models that have F1 and F2 will be able to have up to 4 functions in this way. The moni button can also be a short press that unblocks the squelch momentarily, holding it for more than 2 seconds the squelch remains unblocked until the next press of the button.

lemielek:
Is it very hard to allow Moni button to have changeable function via menu?
Do we need to hardcode it?
Long press for second function is also nice option if possible.

ranguli:
> One possible solution is for the moni button to be a monitor key, and the F1 (F2 when present) buttons to be software-defined as to what function they should have from the menu, depending on the user's needs (as is the case in some models).

> Is it very hard to allow Moni button to have changeable function via menu?

It seems like almost every OpenRTX target I looked for the [comparison table](https://gist.github.com/ranguli/d152d15e32ddff684a5b02d7393312af) allows the user to map side-keys/function keys to custom functions in the stock firmware. So there is a precedent for allowing the user to customize key behaviour. Aside from it being an expected/basic feature it could also be helpful with regards to solving this issue.

PR #429, opened 2026-03-27
power: Add a 2.5W output power option

ranguli:
This PR adds an option for 2.5W power output. It incorporates the work of contributor [romanatrn](https://github.com/romanatrn/openrtx) who had begun this work in #358. 

Unlike <1W power output options, this should not require additional calibration points for devices as the calibration data from 1W to 5W is presumed to be linear.  This PR also opens the door for several power options to be added cleanly without `if` conditionals using the `tx_power_steps` array, similar to what is done for `freq_steps` directly above it.

image

Based on the work by romatrn, the font size of the power output is changed dynamically so that a longer strings containing a decimal point (like "2.5W") will still fit on-screen. 

(Down the road, in another PR a solution should be devised for handling larger power outputs like 10W, 50W etc. on high-power HTs and mobile radios) 

_Edit: This has been tested on the Linux emulator and a CS-7000-M17 Plus. I don't have the appropriate equipment to test how close the output is to 2.5W, so testers would be much appreciated._

### DIFF
--- a/openrtx/src/ui/default/ui_menu.c
+++ b/openrtx/src/ui/default/ui_menu.c
@@ -1284,8 +1284,19 @@ bool _ui_drawMacroMenu(ui_state_t* ui_state)
     // Remember that power is expressed in mW!
     unsigned int power_int = (last_state.channel.power / 1000);
     unsigned int power_dec = (last_state.channel.power % 1000) / 100;
-    gfx_print(pos_2, layout.top_font, TEXT_ALIGN_RIGHT,
-              color_white, "%d.%dW", power_int, power_dec);
+
+    if (power_dec == 0)
+    {
+        // Print power output as an integer without a decimal (i.e "5W")
+        gfx_print(pos_2, layout.top_font, TEXT_ALIGN_RIGHT,
+                  color_white, "%dW", power_int);
+    }
+    else
+    {
+        // Smaller font to fit larger strings on screen (i.e "2.5W")
+        gfx_print(pos_2, layout.top_font - 1, TEXT_ALIGN_RIGHT,
+                  color_white, "%d.%dW", power_int, power_dec);
+    }
 
     // Third row
 #if defined(CONFIG_UI_NO_KEYBOARD)


openrtx/src/core/state.c
```diff
@@ -67,6 +72,7 @@ void state_init()
     state.emergency     = false;
     state.txDisable     = false;
     state.step_index    = 4; // Default frequency step 12.5kHz
+    state.txPowerIndex  = 3; // Default power output of 1W
```
Marko13-dev: Why not setting txPowerIndex to 0 ? I don't understand why setting it to 3 (even the result should be same because of the modulo code below)

openrtx/src/ui/default/ui_menu.c
```diff
@@ -1282,10 +1282,9 @@ bool _ui_drawMacroMenu(ui_state_t* ui_state)
 
     // Compute x.y format for TX power avoiding to pull in floating point math.
     // Remember that power is expressed in mW!
-    unsigned int power_int = (last_state.channel.power / 1000);
-    unsigned int power_dec = (last_state.channel.power % 1000) / 100;
+    unsigned int power = (last_state.channel.power / 1000);
     gfx_print(pos_2, layout.top_font, TEXT_ALIGN_RIGHT,
-              color_white, "%d.%dW", power_int, power_dec);
+              color_white, "%dW", power);
```
Marko13-dev: Maybe we can use the short version when power_dec == 0 and otherwise the long version with the dot?

openrtx/src/core/state.c
ranguli: Good catch, thank you! 👍 I was initially experimenting with <1W options, and so an index of 3 ended up being 1W. I missed this and didn't change it over after removing the other options.

ranguli: I've pushed a fix now to set `state.txPowerIndex = 0;`

openrtx/src/ui/default/ui_menu.c
ranguli: Yeah good idea! I experimented with this and got it to make only 2.5W smaller. I'm not sure if `FONT_SIZE_6PT` is the correct value or whether there's a more idiomatic variable to use here.

Image

```c
    unsigned int power_int = (last_state.channel.power / 1000);
    unsigned int power_dec = (last_state.channel.power % 1000) / 100;

    if (power_dec == 0)
    {
        gfx_print(pos_2, layout.top_font, TEXT_ALIGN_RIGHT,
               color_white, "%dW", power_int);
    }
    else {
        gfx_print(pos_2, FONT_SIZE_6PT, TEXT_ALIGN_RIGHT,
              color_white, "%d.%dW", power_int, power_dec);
    }
```

If the use of `FONT_SIZE_6PT` here is appropriate I can commit this change

Marko13-dev: Or if you want it more generic maybe just one font size smaller than top_font? But at the end I think it is personal preference.
(layout.top_font > 0) ? layout.top_font - 1 : 0

ranguli: Done :+1: I agree, `layout.top_font - 1` seems more readable and has a clearer intent than switching from `layout.top_font` to `FONT_SIZE_6PT`.

openrtx/include/core/state.h
```diff
@@ -48,12 +48,16 @@ typedef struct
     bool       restore_eflash;
     bool       txDisable;
     uint8_t    step_index;
+    uint8_t    txPowerIndex;
```
silseva: This field is being used only by the UI, is better to have it inside the UI state data structure: https://github.com/OpenRTX/OpenRTX/blob/75bf2fee175443f136c8e63381d5256511fe42ec/openrtx/include/ui/ui_default.h#L204

openrtx/src/core/state.c
```diff
@@ -25,6 +25,11 @@ const uint32_t freq_steps[] = { 1000, 5000, 6250, 10000, 12500, 15000,
                                 20000, 25000, 50000, 100000 };
 const size_t n_freq_steps   = sizeof(freq_steps) / sizeof(freq_steps[0]);
 
+// TX power output steps, in mW
+// NOTE: Values below 1W likely require additional calibration data
+const uint32_t tx_power_steps[] = { 1000, 2500, 5000 };
```
silseva: This list can go in ui.c, and it can be made `static const`.
https://github.com/OpenRTX/OpenRTX/blob/75bf2fee175443f136c8e63381d5256511fe42ec/openrtx/src/ui/default/ui.c#L255

openrtx/src/core/state.c
```diff
@@ -25,6 +25,11 @@ const uint32_t freq_steps[] = { 1000, 5000, 6250, 10000, 12500, 15000,
...
+const size_t n_tx_power_steps = sizeof(tx_power_steps) / sizeof(tx_power_steps[0]);
```
silseva: Some time ago we added the `ARRAY_SIZE` macro in `core/utils.h`: you can directly use it in the code instead of defining a variable. The advantage is that it will evaluate to a numeric constant at compile time and allows the compiler to better optimize the code.

openrtx/src/ui/default/ui.c
```diff
@@ -1005,10 +1005,12 @@ static void _ui_fsm_menuMacro(kbd_msg_t msg, bool *sync_rtx)
...
             break;
         case 6:
-            if (state.channel.power == 1000)
-                state.channel.power = 5000;
-            else
-                state.channel.power = 1000;
+
+            // Cycle over the available power output steps
+            state.txPowerIndex++;
+            state.txPowerIndex %= n_tx_power_steps;
```
silseva:
```suggestion
            state.txPowerIndex = (state.txPoweIndex + 1) % ARRAY_SIZE(tx_power_steps);
```

silseva:
@ranguli when applying the changes, please use [fixup commits](https://moodledev.io/general/development/abc/fixups) instead of adding the changes on top with the "Fix ..." commits. This allows to do a git rebase with autosquash and keep the history clean.

Another thing: with the current implementation, the current position of the TX power selection array is reset to zero each time the radio is power-cycled. This means that, instead of having the progression 1W > 2.5W > 5W, the selection will always "jump" to 1W and move one step forward each time the key 6 is pressed.
Probably this sequence can be handled better using a `switch` statement based on the value of `state.channel.power`. See my comment here: https://github.com/OpenRTX/OpenRTX/pull/358/changes#r2366272402

ranguli:
@silseva Thanks for the review :+1: I've addressed the changes (moving `tx_power_steps` to `ui.c` as `static const`, using the `ARRAY_SIZE()` macro instead of `n_tx_power_steps`, etc).

For the power-reset issue, instead of a `switch` I used a lookup on `state.channel.power`, which gets rid of `txPowerIndex`, and should keep things in sync with `tx_power_steps[]` if new power options (i.e <1W or high power for mobile radios) are ever added:

```diff
         case 6:
+        {
+            // Find current power level in the table
+            uint8_t current_index = 0;
+            for (uint8_t i = 0; i < ARRAY_SIZE(tx_power_steps); i++)
+            {
+                if (tx_power_steps[i] == state.channel.power)
+                {
+                    current_index = i;
+                    break;
+                }
+            }
 
             // Cycle over the available power output steps
-            state.txPowerIndex++;
-            state.txPowerIndex %= n_tx_power_steps;
-            state.channel.power = tx_power_steps[state.txPowerIndex];
+            current_index = (current_index + 1) % ARRAY_SIZE(tx_power_steps);
+            state.channel.power = tx_power_steps[current_index];
 
             *sync_rtx = true;
             vp_announcePower(state.channel.power, queueFlags);
             break;
+        }
```

Let me know if you'd prefer the `switch` approach that's more like [this](https://github.com/OpenRTX/OpenRTX/pull/358/changes#diff-a29a963807e0deeb8f60f210e335cb23879ad2963e7ba962b8ba7452ba47a5b1R993) part from #358 instead, then I will submit the changes :+1:

silseva:
Given that the UI allows to cycle through the power levels only along one direction, I think that we can adopt a KISS approach and use a `switch` statement, eventually wrapped in a dedicated function.

Something like this:
```
static uint32_t _ui_getNextPowerLevel(uint32_t curr)
{
    switch(curr) {
        case 1000:
            return 2500;
            break;

        case 2500:
            return 5000;
            break;

        default:
            return 1000;
    }
}
```
This has the extra advantage that, in case the power level assumes some borked values, it will always be assigned a sane default.

ranguli:
Nice, ok I will use the `switch` statement approach (with the dedicated function to be added later on as mentioned) and then push changes :+1:

lz2sat:
Better decision was 1W and 5W because that`s 7 Db difference , so ~ 1 S point by S meter , the best is to have 0.2W , 1W and full 5W (Low , Medium and High Power )

silseva:
We can have as many TX power levels as we wish. :)
Good point about the 7dB difference, to be kept in mind when adding fractional power levels!

PR #439, opened 2026-04-09
Add USB serial support for the CS7000P/STM32H7 targets

ranguli:
## Summary

This PR adds support for USB serial for STM32H7 targets using tinyUSB. While this obviously only applies to the CS7000P, the interface and the H7 implementation should make adding support for F4 targets easier.

Cool future use cases of this besides debugging include using OpenRTX devices as a TNC for APRS/AX.25/M17 packet, using a HT as a serial USB GPS device (accurate time source for portable POTA ops) and as transport for CAT control.

## What this does

- Bumps `tinyUSB` from 0.12 to 0.20. Needed this for bugfixes/STM32H7 support, and it's added as a dependency in the Meson build for the CS7000P target.
- Configure `tusb_config.h` to add support for the H7
- Adds a platform-agnostic `usb_serial` interface that targets can implement.
- Adds an STM32H7 implementation of the platform-agnostic `usb_serial` interface.
- Updates the CS7000P linker script (USB OTG on the H7 supposedly only works with specific memory regions)
- Adds `usb_serial_init()` at startup and integrates USB serial into the main thread
- `stdout` and `stderr` are redirected to the serial output if the `ENABLE_STDIO` option is added to `meson.build`, just like the [docs suggest](https://openrtx.org/#/software?id=debugging-with-com-port) for STM32F4 targets.
- Communication is unidirectional (device -> host only, i.e support for `stdin`).
- The device shows up on a host as a single CDC virtual serial port.

I'm aware that people have tried implementing this before so there are a few things that cropped up which caused issues:

1. This [YouTube video](https://www.youtube.com/watch?v=XgsnJdY8LsY) goes over TinyUSB setup with the STM32H7 and explained a couple things that I was unaware of.
2. A big issue is the requirement of a clock source, which sounds obvious in hindsight. So the on-board 48MHz crystal oscillator is used. I don't know why but this never came up in any research besides the video.
3. There was a weird power-related issue (that I still don't fully understand) which ended up resulting in not being able to turn the CS7000P off, apparently due to the internal USB voltage regulator causing a voltage to appear that's above the power-off threshold used by `platform_pwrButtonStatus()`. **I think it's a good idea for someone with more hardware experience than me to look at this and make sure it's fine.**

## Testing

- Device enumerates correctly (verified on Linux)
    ```
    ~ journalctl -k -f
    kernel: usb 1-1: new full-speed USB device number 112 using xhci_hcd
    kernel: usb 1-1: New USB device found, idVendor=0483, idProduct=4001, bcdDevice= 1.00
    kernel: usb 1-1: New USB device strings: Mfr=1, Product=2, SerialNumber=3
    kernel: usb 1-1: Manufacturer: OpenRTX
    kernel: cdc_acm 1-1:1.0: ttyACM10: USB ACM device
    ````
- Tested with `minicom`, `HTerm`, `picocom`, and `screen`
  - The latter two don't really like `\r\n` line endings by default but that's more of a host/tty issue than a code issue.
- Temporarily added some `printf()` statements in the UI thread to mock debugging and confirm nothing breaks
- Unplugging and then reconnecting the device to the host does not cause any crashes or freezes on the device.
- Serial connection appears to be properly de-initialized when the device is powered-off.
- Added a platform test (`usb_serial_test.c`) which prints to the serial output in a loop

### DIFF
--- a/platform/drivers/USB/usb_serial_stm32h7.c
+++ b/platform/drivers/USB/usb_serial_stm32h7.c
@@ -1,0 +1,243 @@
+/*
+ * SPDX-FileCopyrightText: Copyright 2020-2026 OpenRTX Contributors
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+#include "interfaces/usb_serial.h"
+#include "peripherals/gpio.h"
+#include "pinmap.h"
+#include "hwconfig.h"
+#include "tusb.h"
+#include <stm32h743xx.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <stddef.h>
+#include <stdio.h>
+#include <stdatomic.h>
+
+/*
+ * USB device mount state, maintained by tud_mount_cb / tud_umount_cb.
+ * Written only from the main thread (tinyUSB callbacks run inside tud_task()).
+ * Read from any thread via usb_serial_write().  _Atomic ensures the read is
+ * safe without a mutex.
+ *
+ * Gated on tud_mounted() (device enumerated) rather than tud_cdc_connected()
+ * (DTR asserted by host).  DTR-gating is too strict: many host tools - loggers,
+ * scripted readers, cat - never assert DTR, which would make output appear
+ * silently dropped.  If the host has not opened a terminal the ring buffer just
+ * fills up and writes return short; no data is sent to the CDC FIFO until a
+ * consumer connects.
+ */
+static _Atomic bool usb_mounted = false;
+
+/*
+ * Initialisation complete flag.  Set at the end of usb_serial_init().
+ * Guards usb_serial_available() and usb_serial_read() against calls before
+ * tinyUSB state is valid.  usb_serial_write() is guarded implicitly by
+ * usb_mounted (false until tud_mount_cb fires after init).
+ */
+static _Atomic bool usb_ready = false;
+
+void usb_serial_init(void)
+{
+    /* Enable D2 SRAM1 clock so that the .usb_ram section at 0x30000000 is
+     * accessible. RCC_AHB2ENR resets to 0; without this, any access to
+     * tinyUSB buffers placed in D2 SRAM causes a bus fault.
+     */
+    RCC->AHB2ENR |= RCC_AHB2ENR_D2SRAM1EN;
+    __DSB();
+
+    /*
+     * Enable the USB voltage detector (PWR_CR3_USB33DEN) and wait for
+     * VDDUSB to stabilise.  The FS PHY requires a valid 3.3 V supply on
+     * VDDUSB; without it the PHY clock never starts and tinyUSB's
+     * reset_core() hangs forever polling GRSTCTL_AHBIDL.
+     *
+     * Note: PWR_CR3_USBREGEN (the internal USB voltage regulator) is
+     * intentionally NOT set.  The regulator sources current from the
+     * battery and creates a leakage path that keeps the battery voltage
+     * reading above the power-off threshold in platform_pwrButtonStatus(),
+     * preventing the device from turning off.  The external 3.3 V rail
+     * is sufficient; only the detector is needed.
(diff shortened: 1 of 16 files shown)


platform/drivers/USB/usb_serial_stm32h7.c
```diff
@@ -0,0 +1,295 @@
...
+        tx_buf[head] = p[written++];
+        head = next;
+    }
+
+    atomic_store_explicit(&tx_head, head, memory_order_release);
+
+    pthread_mutex_unlock(&tx_mutex);
+
+    return (ssize_t) written;
+}
```
ranguli, line 348: To-Do: This function _should_ be blocking and not best-effort as this could allow USB CDC serial to be the mechanism for CAT control.

ranguli: @silseva Can you clarify how you want the blocking behaviour implemented?

silseva: The execution flow should stay inside the `write` function until all characters are sent out.
To avoid endlessly blocking the execution flow when the USB is not connected, the function should return an error before attempting to write data.

mdiepart:
Hi!
I have been trying to dump the content of the external flash of my CS7000 M17 Plus and have been trying to port tinyusb to it, before being told about this PR, which is great as I never managed to have anything working but this PR does :D
I am not sure I understand all the changes as they seem to differ significantly from the way USB is implemented in the other devices. Maybe it would be worth a shot migrating all the devices to the same architecture for USB?

Also, I then tried to use this PR to dump the flash but this does not work out of the box as the usb init is done in the state task which is not started in `nvm_dump.c` and also tinyUSB relies on the `tud_task()` being called repetitively, which is not the case in the platform tests. I think USB subsystem would be better initialized in `bsp_init2()` function with an `#ifdef` to only init the USB subsystem if STDIO is actually enabled. I don't really know how the `tud_task()` calls should be handled, but the state task should definitely not be started in the platform tests such as `nvm_dump()` as this creates the risk of making the tests reliant on the working state of other bits of code (in my case it seems the calibration coefficients disappeared, so OpenRTX does not start but `nvm_dump()` does).

ranguli:
Hi @mdiepart,

> I am not sure I understand all the changes

Honestly there are some parts of it I don't fully understand either, other than it's what TinyUSB does/recommends. USB is insanely complicated :sweat_smile:  

> as they seem to differ significantly from the way USB is implemented in the other devices.

Yeah it's very different structurally from the USB on say the MD3x0, which I think was mostly a side-effect of switching to use TinyUSB. I believe @silseva had suggested that some kind of TinyUSB-based approach should replace the existing one, so hopefully the structural differences won't be an issue.

> Also, I then tried to use this PR to dump the flash but this does not work out of the box as the usb init is done in the state task which is not started in nvm_dump.c and also tinyUSB relies on the tud_task() being called repetitively, which is not the case in the platform tests.

Yeah this PR was intended to be more of a 'basic USB support' that provided console debugging and lay the groundwork for console debugging and maybe packet/TNC or CAT control. That being said, the USB system should definitely be designed to accommodate the use case of dumping NAND as that would be extremely useful  :+1: 

> I think USB subsystem would be better initialized in bsp_init2() function with an #ifdef to only init the USB subsystem if STDIO is actually enabled. I don't really know how the tud_task() calls should be handled, but the state task should definitely not be started in the platform tests such as nvm_dump() as this creates the risk of making the tests reliant on the working state of other bits of code (in my case it seems the calibration coefficients disappeared, so OpenRTX does not start but nvm_dump() does)

I don't have any strong opinions either way on the details of how we make it more flexible, initializing it in BSP init sounds fine to me! I'll need to take a closer look to see if that requires re-working the USB code. @silseva do you have any suggestions about how this should be done?

silseva:
Not stale, it will be merged when migrating to miosix 3.0

PR #440, opened 2026-04-10
CS7000P: Use hardware VOX circuit for detecting microphone input level

ranguli:
This PR updates the source used for determining the microphone input level on the CS7000P for metering. Previously, the microphone input ADC was being used, which has a 1.65V bias and is not suitable for this purpose, which caused a bug where the microphone input was always 50%.

The `platform_getMicLevel()` function now samples an ADC channel for the radios hardware VOX circuit instead. previously unused by OpenRTX on the CS7000P. The VOX circuit is a peak detector and isn't a perfect indicator for mic level, but seems to be the best available. 

This change also makes the CS7000P ready for VOX functionality being added to OpenRTX in the future, which would require `platform_getMicLevel()`.


silseva:
Looks good!
You can squash together the first and second commit, actually. :)

silseva (approved):

ranguli:
@silseva squashed :)

PR #442, opened 2026-04-12
Mic oversampling

turnrye:
This PR takes @JetseVerschuren's work on oversampling and makes a few tweaks based on real world testing (cs7000p and uv390) plus anticipated code review feedback. I'm helping with getting the changes landed here as well as trying to quantify the impact of the work.

## Summary of problem and changes

OpenRTX samples the microphone ADC at 8 kHz: the same rate codec2 consumes samples. Quantization noise from the 12-bit ADC is encoded alongside speech, degrading transmitted audio.
This change oversamples the ADC at N×8 kHz, then averages every N samples back to 8 kHz before feeding codec2. Averaging acts as a low-pass decimation filter that discards out-of-band quantization noise, yielding a theoretical noise reduction of 10·log10(N) dB.

The oversampling factor is configured per-platform via CONFIG_MIC_OVERSAMPLE in each target's hwconfig.h. Most platforms use factor 8 (64 kHz). The CS7000-PLUS uses factor 6 (48 kHz) because its LPTIM3 timer cannot reliably trigger ADC conversions at 64 kHz. Platforms without oversampling support use factor 1, which compiles out the decimation path.

### Speech SNR Improved >4dB

A simple power-ratio SNR of real-world voice recordings was measured (only considering the times when there was no speaking, following ITU-T P.56 and using WebRTC VAD). **This resulted in a measurable, consistent noise floor drop (consistent across all pairs) of over 4 dB with oversampling enabled.**

This aligns with the theoretical expectation from oversampling. Decimating by a factor of N reduces uncorrelated noise by 10log10(N). For the CS7000-PLUS with oversampling=6: 10log10(6) = 7.8 dB theoretical max. The measured 4.9 dB is in the right ballpark — the gap is expected since not all noise in the system is uncorrelated quantization noise (some is analog, some is RF)

### Raw spoken recordings

Reading some of the [harvard speech quality measurement sentences](https://www.cs.columbia.edu/~hgs/audio/harvard.html) from https://doi.org/10.1109/IEEESTD.1969.7405210. Using radios held about 3 inches away from the mouth. In a household environment with some background noises. Recordings produced using m17-fme and the codec2 to ogg mechanism, placed in webm containers to make github happy.

[catherine-cs7000p-without-oversampling.webm](https://github.com/user-attachments/assets/89070694-0e9f-40a1-b762-ef2a009eeac4)
[catherine-cs7000p-with-oversampling.webm](https://github.com/user-attachments/assets/85a6ede7-002c-4d11-bb98-47826271d290)
[ryan-cs7000p-without-oversampling.webm](https://github.com/user-attachments/assets/abc820ef-8089-4e88-8aad-c659e2d47a26)
[ryan-cs7000p-with-oversampling.webm](https://github.com/user-attachments/assets/472ae178-917a-41f9-bd0b-53e2efb19812)
[catherine-uv390-without-oversampling.webm](https://github.com/user-attachments/assets/e6dfaf2f-ea6b-43f1-b15f-e045d327347a)
[catherine-uv390-with-oversampling.webm](https://github.com/user-attachments/assets/e88531c3-253b-4a46-b92b-629f2522f0c8)
[ryan-uv390-without-oversampling.webm](https://github.com/user-attachments/assets/9175b772-cee4-429e-a5e8-4c86b1e387bc)
[ryan-uv390-with-oversampling.webm](https://github.com/user-attachments/assets/e792f62a-bd26-4de4-ae17-e4e2dba962bf)

Note: ryan's recordings without oversampling appear to be mislabeled, and one is missing (tbd which one)

### DIFF
--- a/openrtx/include/core/dsp.h
+++ b/openrtx/include/core/dsp.h
@@ -11,6 +11,7 @@
 #include <stddef.h>
 #include <stdint.h>
 #include <string.h>
+#include "interfaces/audio.h"
 
 #ifdef __cplusplus
 extern "C" {
@@ -59,6 +60,42 @@ static inline void dsp_removeDcOffset(struct dcBlock *dcb, int16_t *buffer,
     for (size_t i = 0; i < length; i++)
         buffer[i] = dsp_dcBlockFilter(dcb, buffer[i]);
 }
+
+/**
+ * Data structure holding the internal state of an oversampling filter.
+ */
+struct oversamplingBlock {
+    uint8_t oversampling;
+    uint8_t count;
+    uint32_t accumulator;
+};
+
+/**
+ * @brief Set the oversampling factor for a decimation filter.
+ *
+ * @param oversamplingBlock: pointer to the oversampling filter state.
+ * @param oversampling: oversampling factor, must be >= 1.
+ */
+void dsp_oversamplingSetOversampling(
+    struct oversamplingBlock *oversamplingBlock, uint8_t oversampling);
+
+/**
+ * @brief Run a single step of the oversampling decimation filter.
+ *
+ * Accumulates input samples and returns their truncated integer average
+ * once the oversampling factor number of samples have been collected.
+ *
+ * Input samples are treated as unsigned values for accumulation. Callers
+ * must ensure that the samples are non-negative (e.g. raw ADC readings)
+ * before the DC offset has been removed.
+ *
+ * @param oversamplingBlock: pointer to the oversampling filter state.
+ * @param sample: pointer to the input sample; overwritten with the
+ * decimated output when the function returns true.
+ * @return true when a decimated sample is ready, false otherwise.
+ */
+bool dsp_oversamplingDecimate(struct oversamplingBlock *oversamplingBlock,
+                              stream_sample_t *sample);
 
 #ifdef __cplusplus
 }
--- a/openrtx/src/core/audio_codec.c
+++ b/openrtx/src/core/audio_codec.c
@@ -22,6 +22,12 @@
 #include "hwconfig.h"
 
 #define BUF_SIZE 4
+#define CODEC2_FRAME_SAMPLES 160 /* Samples per codec2 3200-mode frame */
+#define DMA_BUF_SAMPLES (CODEC2_FRAME_SAMPLES * 2) /* Double-buffered DMA */
+
+#ifndef CONFIG_MIC_OVERSAMPLE
+#define CONFIG_MIC_OVERSAMPLE 1
+#endif
 
 static pathId audioPath;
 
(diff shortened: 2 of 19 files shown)


turnrye:
Hey @JetseVerschuren could you also provide a review of this? Thanks in advance!

silseva:
Can you separate the clang-format changes to a separate commit, please?

turnrye:
> Can you separate the clang-format changes to a separate commit, please?

Done!

openrtx/include/core/dsp.h
```diff
@@ -60,6 +61,42 @@ static inline void dsp_removeDcOffset(struct dcBlock *dcb, int16_t *buffer,
         buffer[i] = dsp_dcBlockFilter(dcb, buffer[i]);
 }
 
+/**
+ * Data structure holding the internal state of an oversampling filter.
+ */
+struct oversamplingBlock {
```
silseva: Just a note: the `dcBlock` data structure is named so because it stands for dcBlock**ing** (filter).
You can call the oversampling data structure "oversampling" and it will be fine.

openrtx/include/core/dsp.h
```diff
@@ -60,6 +61,42 @@ static inline void dsp_removeDcOffset(struct dcBlock *dcb, int16_t *buffer,
...
+    uint8_t oversampling;
```
silseva: This can be an input parameter of `dsp_oversamplingDecimate`

openrtx/include/core/dsp.h
```diff
@@ -60,6 +61,42 @@ static inline void dsp_removeDcOffset(struct dcBlock *dcb, int16_t *buffer,
...
+void dsp_oversamplingSetOversampling(
+    struct oversamplingBlock *oversamplingBlock, uint8_t oversampling);
+
+/**
+ * @brief Run a single step of the oversampling decimation filter.
+ *
+ * Accumulates input samples and returns their truncated integer average
+ * once the oversampling factor number of samples have been collected.
+ *
+ * Input samples are treated as unsigned values for accumulation. Callers
```
silseva: > Input samples are treated as unsigned values for accumulation. Callers must ensure that the samples are non-negative (e.g. raw ADC readings) before the DC offset has been removed.

This cannot be granted! The audio API makes use of `int16_t` values because a particular platform can return values without a DC offset. This is the case of linux, for example

openrtx/src/core/audio_codec.c
```diff
@@ -161,45 +163,83 @@ static void *encodeFunc(void *arg)
...
+    // overflowing.
+    // Starting a new thread is an option, but the new stack is already 2kB,
+    // and it also adds a big amount of extra complexity and shared buffers,
+    // so not actually beneficial.
+    // The best solution would be to implement the decimation at a DMA
+    // interrupt/callback, but that's too complex for now, and there's
+    // enough heap available still.
+    // Allocate on the heap, as the stack isn't big enough for oversampling
+    // above 2 or so.
+    stream_sample_t *adcBuf = malloc(DMA_BUF_SAMPLES * CONFIG_MIC_OVERSAMPLE
```
silseva, line 172: > The best solution would be to implement the decimation at a DMA interrupt/callback, but that's too complex for now, and there's enough heap available still.

This cannot be done, as it will break the real-time constraints. I think we can remove the whole comment as well. :)

openrtx/src/core/audio_codec.c
```diff
@@ -161,45 +163,83 @@ static void *encodeFunc(void *arg)
...
-
-        dataBlock_t audio = inputStream_getData(iStream);
-        if (audio.data == NULL)
-            break;
-
-#ifndef PLATFORM_LINUX
-        for (size_t i = 0; i < audio.len; i++) {
-            int16_t sample;
+        stream_sample_t audioBuf[CODEC2_FRAME_SAMPLES];
+        for (size_t i = 0; i < CODEC2_FRAME_SAMPLES;) {
```
silseva: Do we really need this `for` loop? `CODEC2_FRAME_SAMPLES` is divisible by two (and I never expect to have a codec requiring an odd number of samples) so, for any oversampling ratio, the number of samples returned by `inputStream_getData` will be an integer multiple of the oversampling ratio and we'll never fall "across" an unfinished block.
To be demonstrated, though.

platform/targets/CS7000-PLUS/hwconfig.h
```diff
@@ -69,6 +69,9 @@ extern const struct gpsDevice gps;
 /* Device has a vibration motor */
 #define CONFIG_VIBRATION
 
+/* Microphone audio input */
+#define CONFIG_MIC_GAIN 32
```
silseva, line 73: I'd call this `CONFIG_CODEC_ENCODE_GAIN`: yes, now it is being used mainly for the microphone, but codec2 can accept any input stream.
I'd change the oversampling macro to `CONFIG_CODEC_OVERSAMPLING` as well

openrtx/src/core/audio_codec.c
```diff
@@ -22,6 +22,8 @@
 #include "hwconfig.h"
 
 #define BUF_SIZE 4
+#define CODEC2_FRAME_SAMPLES 160 /* Samples per codec2 3200-mode frame */
+#define DMA_BUF_SAMPLES (CODEC2_FRAME_SAMPLES * 2) /* Double-buffered DMA */
```
silseva, line 27: Here I'd add
```
#ifndef CONFIG_MIC_OVERSAMPLE
#define CONFIG_MIC_OVERSAMPLE 1
#endif
```
in this way we can entirely omit the directive when there is no oversample

openrtx/include/core/dsp.h
```diff
@@ -11,6 +11,7 @@
 #include <stddef.h>
 #include <stdint.h>
 #include <string.h>
+#include "interfaces/audio.h"
```
silseva: I plan to remove the `stream_sample_t` typedef: you can just replace it with `int16_t` and drop this include directive

silseva (changes requested):

platform/targets/CS7000-PLUS/hwconfig.h
turnrye: > I'd call this CONFIG_CODEC_ENCODE_GAIN: yes, now it is being used mainly for the microphone, but codec2 can accept any input stream.

I think this isn't a desired feature though... the gain is specifically for scaling up mic ADC readings. If we were to use encodeFunc for something different from the mic, why would we want to have it scale using these hardware specific gain values?

Same logic on oversampling -- it's to do with compensating for hardware mic noise, it may not be the right thing to do for other sources.

silseva: Fine, let's keep them as they are now.

silseva:
Why did you remove the oversampling code from dsp.h/dsp.cpp? :cry: That code was good, and having an oversampling implementation can be useful also in the future.
My point was only about reducing the complexity of the `for` loop applying the oversample decimation. :)

silseva (approved):

silseva:
@turnrye  is this good to go?

turnrye:
> @turnrye is this good to go?

Sorry -- got pulled away from the computer! Yes I added back the oversampling functionality into dsp but with things simplified a bit. This is ready for review again, and hopefully I got it right this time :)

PR #443, opened 2026-04-15
drivers: Fix bug in STM32 GPIO OPEN_DRAIN_PU case

ranguli:
It looks like this might be have been a typo. OPEN_DRAIN and OPEN_DRAIN_PU are identical case statements, but the name and comment both imply that the pull-up should be set but it isn't.

This code path doesn't seem to be used based on a grep for `OPEN_DRAIN_PU`, but presumably it would break something in the future if it was used.


silseva:
Definitely a copy-paste error. Thank you!

silseva (approved):

PR #445, opened 2026-04-17
UI Snapshot Testing

turnrye:
This change adds a ui testing tool (called end to end "e2e", because in the future it can be extended to black box test other things besides just the UI).

Run the script with `./scripts/run_e2e.py --help`.

```
./scripts/run_e2e.py --help
usage: run_e2e.py [-h] [--binary BINARY] [--variant VARIANT] [--build-dir BUILD_DIR] [-j JOBS] [--tolerance TOLERANCE] [--update-golden]
                  [--wrapper WRAPPER]
                  [test_script]

OpenRTX E2E test runner

positional arguments:
  test_script           Path to test script (.txt); omit to run all discovered tests

options:
  -h, --help            show this help message and exit
  --binary BINARY       Path to openrtx binary
  --variant VARIANT     UI variant (default: default)
  --build-dir BUILD_DIR
                        Build directory containing emulator binaries (default: build_linux)
  -j JOBS, --jobs JOBS  Parallel test workers when running all tests (default: number of CPUs)
  --tolerance TOLERANCE
                        Max differing pixels before a screenshot fails (default: 0)
  --update-golden       Overwrite golden images with current output
  --wrapper WRAPPER     Command (with args) to wrap the emulator with, e.g. 'gdb --batch -ex run -ex bt --args' or 'valgrind'. Inserted
                        between faketime and the binary. Must be non-interactive: stdin is piped to the test script and stdout/stderr are
                        captured.
```

The test cases are `txt` files leveraging the existing DSL. Each snapshot image is compared against a golden image that is checked in. This runs on two targets: linux and linux_module17, which represent two substantially different UXes today.

This also included a number of expansions to the linux platform, the dsl, and a few bug fixes:
* Fixed a race where the async screenshot shell command could return before the BMP file was actually written
* Fixed an out-of-range stream ID bug that could trigger a global-buffer-overflow and intermittent emulator shutdown crashes
* Fixed shutdown race/crash by removing [exit()](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) from the worker-thread terminate path so cleanup and atexit handling happen safely on the main thread

All of these errors were previously being encountered using the UI testing tooling that was built, getting in the way of the test scripts behaving in a consistent way. With these bug fixes added (plus `wait_stable` and `screenshot_sync`) now the tests are predictable and consistent.


silseva:
The only concern I have about adding the assets for the e2e testing is that in the long term they will have a significant impact on the repository size. Should we create a separate repository only for this kind of testing?

turnrye:
We could shift approaches and instead compare hashes, it means we'd lose the ability to see the expected result, but it'd avoid entirely the image size issue. We maybe could even still have another repo then but have it be periodic snapshots rather than be the authoritative source for the actual results. Maybe there are other options, too.

turnrye:
Closing -- will roll this into my upcoming UI rework

PR #446, opened 2026-04-17
scripts: uv and ruff

turnrye:
This change adopts the Black standards and tooling for our python scripts. The goal of this is to make it so that as scripts are added or updated, there doesn't need to be time spent on formatting code -- the format is set, easy to adopt, and enforced.

For more about Black or its accolades, see https://github.com/psf/black.

This change:
- Adds a formatter call to CI
- Adds Black as a requirement
- Updates the CONTRIBUTING guide to reference Black for python scripts
- Reformats the python files to follow the standard


turnrye:
This results in a hilarious diff for `bin2sgl.py`, but any IDE makes it trivial to fold those block up so this doesn't pose a real nuisance.

silseva:
There are some cases where, in my opinion, the code readability becomes a bit worse.
For example
```
symbols = sorted([f for f in os.listdir(os.path.join(absolute_path, symbol_source_dir)) if "svg" in f])
```
becoming
```
symbols = sorted(
    [
        f
        for f in os.listdir(os.path.join(absolute_path, symbol_source_dir))
        if "svg" in f
    ]
)
```

Is there a possible solution for this or we just carry on?

ranguli:
I think `black` is very opinionated about how it styles Python. `ruff` is quite popular, it might allow more control of styling?

turnrye:
Yes black is intentionally rigid. But its opinions are well established, but ruff is a great alternative if openrtx wants to set its own python standards. Im not strongly opinionated though, so thats why i started with Black.

silseva:
I use ruff at work, we can try it and see how it behaves with our scripts. In general I prefer to have readable code even if not completely adherent to the standards than the other way round.

turnrye:
Great, ill change this over to ruff.

silseva:
@turnrye is there a reason why this PR is marked as draft?

silseva:
Looks good to me!
There seems to be an issue with the `intelhex` package, though:
```
/workspaces/OpenRTX/scripts/dfu-convert.py -b 0x08000000:openrtx_cs7000_bin openrtx_cs7000_dfu
Traceback (most recent call last):
  File "/workspaces/OpenRTX/scripts/dfu-convert.py", line 9, in <module>
    from intelhex import IntelHex
ModuleNotFoundError: No module named 'intelhex'
```

turnrye:
> Looks good to me! There seems to be an issue with the `intelhex` package, though:
> 
> ```
> /workspaces/OpenRTX/scripts/dfu-convert.py -b 0x08000000:openrtx_cs7000_bin openrtx_cs7000_dfu
> Traceback (most recent call last):
>   File "/workspaces/OpenRTX/scripts/dfu-convert.py", line 9, in <module>
>     from intelhex import IntelHex
> ModuleNotFoundError: No module named 'intelhex'
> ```

Oop, thanks for pointing this out. This was happening because while my changes moved where the python deps were managed, they didn't update how they got called via meson. So, I've had to make a few changes to make meson "aware" of our using UV:
- meson now checks to ensure uv is installed
- meson invokes python scripts via `uv run`, which then leverages the `.venv` 

This makes it where intelhex no longer has to be a system-installed package for our builds.

silseva:
Thank you!
Out of curiosity: what is the advantage of using `uv` instead of `pip`?

turnrye:
> Thank you! Out of curiosity: what is the advantage of using `uv` instead of `pip`?

`uv`it handles other concerns like env creation, python versioning, and dependency locking that pip doesn't. In my limited experience, I've found `uv` is a bit more foolproof. Also, is a good bit faster than `pip` thanks to aggressive caching and parallel downloading -- so a minor quality of life improvement there.

turnrye:
This as a single PR is kinda rough. And there are a couple of independent steps that are each valuable. So, I'm going to break this into stages:
1. https://github.com/OpenRTX/OpenRTX/pull/508 -- install using uv
2. adopt ruff
3. adopt uv's project workflow

Closing this PR. See you in the other ones!

PR #447, opened 2026-04-17
ui: Refactor UI code into several files

ranguli:
**Note:** This is currently a proof-of-concept/proposal.


ranguli:
I think this proof of concept can be closed, as we've chosen to go the C++ route so this doesn't really fit into the current UI refactoring plans.

PR #448, opened 2026-04-17
RTX packet interface API

silseva:
Common interface between application and operating modes for sending and receiving packet data. The API is asynchronous and meant to allow zero-copy transmission and reception of packets.

## Application side:

Application code sends packet transmission and reception requests by means of packet descriptors. A packet descriptor contains a status variable, a pointer to the packet data and a size field. To submit a request, the application code populates the descriptor, sets its status to `PKT_STATUS_IDLE` and calls `rtx_addPacketRx` or `rtx_addPacketTx`: the rtx layer accepts the request and sets the status to `PKT_STATUS_SUBMITTED`. Once the status is set to `PKT_STATUS_SUBMITTED`, the application code should not change the struct fields. Once submitted, the application code polls the `status` field until it becomes `PKT_STATUS_DONE` or `PKT_STATUS_ERROR`.

The `size` field assumes the following meanings:
| Operation | PKT_STATUS_IDLE | PKT_STATUS_DONE | PKT_STATUS_ERROR |
| --------------- | ------------------------- | ---------------------------- | ----------------------------- |
| TX               |  # bytes to send | (no change) | ERRNO code |
| RX               |  max buffer size | # bytes received | ERRNO code |

TX example:
```
uint8_t msg[42];
struct pktDesc desc {
    .status = PKT_STATUS_IDLE,
    .buffer = msg,
    .size = sizeof(msg);
};

if (rtx_addPacketTx(&desc))
    puts("msg submit error");

[...]
    
if (desc->status == PKT_STATUS_DONE)
    puts("msg send ok");

if (desc->status == PKT_STATUS_ERROR)
    puts("msg send error");
```
The descriptor must live also after the call to `rtx_addPacketRx`/`rtx_addPacketTx`.

## OpMode side

OpMode classes supporting packet data have to implement the `addPacketRx` and `addPacketTx` functions. In doing so, is necessary to take into account that these functions can be called by multiple threads at the same time. A useful primitive is the `NbRingBuffer` class, providing a non-blocking multi-threaded ring buffer implementation.

Code example:
```
NbRingBuffer < struct pktDesc *, 4 > rxPktQueue;
struct pktDesc *currPkt = NULL;

int addPacketRx(struct pktDesc *packet) override
{
    return rxPktQueue.tryPush(packet);
}

[...]

rxPktQueue.tryPop(currPkt);
if (currPkt) {
    // do actual RX
    [...]
    
    // Once done
    currPkt->status = PKT_STATUS_DONE;
    currPkt = NULL;
}
```

OpModes can also keep a queue of a single descriptor:
```
pthread_mutex_t mutex;
struct pktDesc *currPkt = NULL;

int addPacketRx(struct pktDesc *packet) override
{
    int ret = 0;
    
    pthread_mutex_lock(&mutex);
    if (currPkt)
        ret = -EBUSY;
    else
         currPkt = packet;
    pthread_mutex_unlock(&mutex);

    return ret;
}

[...]

if (currPkt) {
    // do actual RX
    [...]
    
    // Once done
    currPkt->status = PKT_STATUS_DONE;
    currPkt = NULL;
}
```

### DIFF
--- a/openrtx/include/core/ringbuf.hpp
+++ b/openrtx/include/core/ringbuf.hpp
@@ -1,6 +1,6 @@
 /*
  * SPDX-FileCopyrightText: Copyright 2020-2026 OpenRTX Contributors
- * 
+ *
  * SPDX-License-Identifier: GPL-3.0-or-later
  */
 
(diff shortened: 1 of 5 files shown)


silseva:
@rxt1077 can you review this as well, please?

openrtx/include/core/ringbuf.hpp
```diff
@@ -141,50 +104,147 @@ class RingBuffer
...
+     * If the buffer is empty the execution flow is blocked until at least one
+     * element is available.
      *
-     * Note: reset is "lazy", as it just sets read pointer, write pointer and
-     * number of elements to zero. No actual erasure of the stored elements is
-     * done until they are effectively overwritten by new push()es.
+     * @param elem: place where to store the popped element.
      */
-    void reset()
+    void pop(T &elem)
```
turnrye, line 210: The signature for `pop` and `push` is changing, but the usages in `Demodulator.cpp` were not updated. This doesn't break the build because its hidden behind `ENABLE_DEMOD_LOG` (which none of our builds enable), but I think we should still clean that up to avoid breaking that feature.

openrtx/include/core/ringbuf.hpp
```diff
@@ -13,111 +13,74 @@
...
-        // The call is blocking: wait until there is some free space
-        while(numElements >= N)
-        {
-            pthread_cond_wait(&not_full, &mutex);
-        }
-
-        // There is free space, push data into the queue
-        data[writePos] = elem;
-        writePos = (writePos + 1) % N;
+        if (full())
```
turnrye, line 51: I dont think it matters for the rtx usecases, but is it OK that this happens before locking? If we reuse this when there are multiple producers and consumers, I wonder if this could introduce race conditions.

openrtx/include/rtx/rtx.h
```diff
@@ -81,6 +82,25 @@ enum opstatus {
...
+    PKT_STATUS_ERROR,     /**< RX/TX error */
+};
+
+/**
+ * Data packet descriptor
+ */
+struct pktDesc {
+    enum pktStatus status;
+    void *buffer;
+    ssize_t size;
```
turnrye, line 101: I dont have much experience with this sort of thing, but the description you put on the PR was necessary for me to follow what this field was doing. I found that https://github.com/torvalds/linux/blob/master/include/uapi/linux/io_uring.h#L502 calls this "res" (result?) instead of size. Maybe that name would be more indicative that the field is overloaded?

turnrye:
I think the approach will serve us well! Thanks for making this!

rxt1077:
At some point we may need to think about multiple modes within an OpMode. Imagine multiple demodulators inside the update loop so you can receive AFSK 1200, 2400, 3600, etc. Do we want to start thinking about an abstract packet buffer that holds all the packets as opposed to individual Tx and Rx buffers for each mode?

rxt1077:
On a side note, this fixes a known issue with the APRS Rx code where it attempts to store an infinite amount of packets.

openrtx/include/core/ringbuf.hpp
silseva: Done!

openrtx/include/rtx/rtx.h
silseva: I added an extra `res` field to hold the result of the operation.

silseva:
> At some point we may need to think about multiple modes within an OpMode. Imagine multiple demodulators inside the update loop so you can receive AFSK 1200, 2400, 3600, etc. Do we want to start thinking about an abstract packet buffer that holds all the packets as opposed to individual Tx and Rx buffers for each mode?

I'd leave it for later, when there will be need for it. At the moment I'm not even really sure that some of the supported devices have enough computing power to run more than one demodulator at the same time still respecting the real-time requirements

openrtx/include/core/ringbuf.hpp
silseva: No, is a race condition. :sweat_smile: Fixed, anyway.

silseva:
@turnrye @rxt1077 reviews applied, let me know if you have other observations.

PR #450, opened 2026-04-19
(1/2) core: Add 'extern' blocks to core/ files used by the UI

ranguli:
This PR adds the following blocks at the top and bottom of C headers in `core/` used by the UI respectively:

```diff
+#ifdef __cplusplus
+extern "C" {
+#endif
```

```diff
+#ifdef __cplusplus
+} // extern "C"
+#endif
```

Additionally, modified files are added to the `clang-format` script and formatting is applied. 

This PR is the 1st of multiple proposed in a larger refactor, and is a requisite for refactoring the UI in C++.

PR #451, opened 2026-04-19
(2/4) ui: Conversion of default UI files from C to C++

ranguli:
This PR continues the work done in #450. It converts the UI `.c` files into `.cpp`, doing only the minimum necessary code changes required for compilation. The logic here is that the minimum-necessary work to convert the files to C++ is done is this PR (without any additional refactor or changes to the code) to make future UI refactoring work more easily distinguishable from the initial `.c` -> `.cpp` step.

With this PR in place, futurerefactoring work on a C++ based UI can be done. 

**Note: This PR only changes the default UI, not the Module 17 UI. If the Module 17 should also be converted in this PR, I can add additional commits to do so.**

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,9 +115,9 @@ target_sources(app
     openrtx/src/protocols/M17/FrameDecoder.cpp
     openrtx/src/protocols/M17/LinkSetupFrame.cpp
 
-    openrtx/src/ui/default/ui.c
-    openrtx/src/ui/default/ui_main.c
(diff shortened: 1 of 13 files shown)


silseva:
clang-format messed up data structured and array initializers due to missing commas in the last element: can you fix them, please?
In general, given that we plan to heavily refactor the UI code in any case: does it really make sense to run clang-format on them?

ranguli:
@silseva Totally agree. The PR should be updated so that it is based off the latest changes from #450, and also undo the clang-format.

PR #452, opened 2026-04-21
Enable reading of hardware info and calibration data for MD9600

marcoSchr:
This merge request enables reading the hardware info and calibration data from flash for the MD9600.

The hardware version is handled different from other MDx targets.
Here we read the major hardware version from flash.
We can then later use this in order to correctly initialize and use the RF part.

As far as I can tell the calibration data is the same as for the MD-UV3x0

### DIFF
--- a/platform/drivers/NVM/nvmem_MDx.c
+++ b/platform/drivers/NVM/nvmem_MDx.c
@@ -140,12 +140,18 @@ void nvm_readHwInfo(hwInfo_t *info)
     uint16_t freqMin = 0;
     uint16_t freqMax = 0;
     uint8_t  lcdInfo = 0;
+    #ifdef PLATFORM_MD9600
+    uint8_t  hwVersionMajor = 0;
+    #endif
 
     // Security register 3: base address 0x3000
     nvm_devRead(&secReg, 0x3000, info->name, 8);
     nvm_devRead(&secReg, 0x3014, &freqMin, 2);
     nvm_devRead(&secReg, 0x3016, &freqMax, 2);
     nvm_devRead(&secReg, 0x301D, &lcdInfo, 1);
+    #ifdef PLATFORM_MD9600
+    nvm_devRead(&secReg, 0x3035, &hwVersionMajor, 1);
+    #endif
 
     // Ensure correct null-termination of device name by removing the 0xff.
     for(uint8_t i = 0; i < sizeof(info->name); i++)
@@ -157,7 +163,11 @@ void nvm_readHwInfo(hwInfo_t *info)
     freqMin = ((uint16_t) bcdToBin(freqMin))/10;
     freqMax = ((uint16_t) bcdToBin(freqMax))/10;
 
+    #ifdef PLATFORM_MD9600
+    info->hw_version = hwVersionMajor;
+    #else
     info->hw_version = lcdInfo & 0x03;
+    #endif
 
     #ifdef PLATFORM_MD3x0
     // Single band device, either VHF or UHF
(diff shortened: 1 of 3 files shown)


platform/targets/MD-9600/platform.c
```diff
@@ -61,7 +51,11 @@ void platform_init()
     gpio_setMode(SPI2_SDI, ALTERNATE | ALTERNATE_FUNC(5));
     spiStm32_init(&spi2, 1300000, 0);
 
+    memset(&hwInfo, 0x00, sizeof(hwInfo));
+
     nvm_init();                      /* Initialise non volatile memory manager */
+    nvm_readCalibData(&calibration); /* Load calibration data                  */
```
silseva, line 57: I moved loading of calibration data to `radio_init` some time ago, on all targets.

platform/drivers/NVM/nvmem_MDx.c
```diff
@@ -140,12 +140,18 @@ void nvm_readHwInfo(hwInfo_t *info)
...
+    #endif
 
     // Security register 3: base address 0x3000
     nvm_devRead(&secReg, 0x3000, info->name, 8);
     nvm_devRead(&secReg, 0x3014, &freqMin, 2);
     nvm_devRead(&secReg, 0x3016, &freqMax, 2);
     nvm_devRead(&secReg, 0x301D, &lcdInfo, 1);
+    #ifdef PLATFORM_MD9600
+    nvm_devRead(&secReg, 0x3035, &hwVersionMajor, 1);
+    #endif
```
silseva, line 154: `lcdInfo` is not used on the MD9600, we can just do the following and get rid of two local variables:
```suggestion
    #ifdef PLATFORM_MD9600
     nvm_devRead(&secReg, 0x3035, &info->hw_version, 1);
    #else
    nvm_devRead(&secReg, 0x301D, &info->hw_version, 1);
    #endif
```

platform/drivers/NVM/nvmem_MDx.c
```diff
@@ -157,7 +163,11 @@ void nvm_readHwInfo(hwInfo_t *info)
     freqMin = ((uint16_t) bcdToBin(freqMin))/10;
     freqMax = ((uint16_t) bcdToBin(freqMax))/10;
 
+    #ifdef PLATFORM_MD9600
+    info->hw_version = hwVersionMajor;
+    #else
     info->hw_version = lcdInfo & 0x03;
```
silseva, line 169: We can move the bitmasking here, where it does make more sense to be: https://github.com/OpenRTX/OpenRTX/blob/4a83e8f6f4c095cc0580ed13ec11eba14fa190d8/platform/drivers/display/HX8353_MD3x.cpp#L206

silseva (changes requested):

platform/targets/MD-9600/platform.c
marcoSchr: I removed this again. We can add it back, when tackling the radio driver.

platform/drivers/NVM/nvmem_MDx.c
marcoSchr: I applied the changes in ba2bf8225445305e8f852fff8441f7f75c726132

platform/drivers/NVM/nvmem_MDx.c
marcoSchr: I applied the changes in ba2bf8225445305e8f852fff8441f7f75c726132

silseva:
Thank you!

PR #453, opened 2026-04-21
Fix display of input voltage for devices without a battery

marcoSchr:
For devices without a battery the voltage is displayed instead of a battery percentage. The formatting of the voltage expected a float, but we store the voltage as an integer. This has the effect, that no voltage was displayed.


silseva (approved):

silseva:
Thank you!

PR #455, opened 2026-04-28
Fix #444: Wraparound in offset 

### LINKED ISSUE #444: Wraparound when setting radio offset
Entering a large frequency as an offset results in an integer wraparound.

## Steps To Reproduce
1. Navigate to Settings -> Radio -> Offset
2. Enter the key sequence  "5" "0" "0" "0" "0" "0" "0" "0" "0" "0"
3. Observe the wraparound:

Image

**Note: Based on the  code, I believe there's a similar overflow bug in the Module 17 UI.**

##  Suggested fix

Replace the bounds checking here:

https://github.com/OpenRTX/OpenRTX/blob/088d906cb5f10370ac451a7aac95366ae95220a1/openrtx/src/ui/default/ui.c#L1245-L1248

with `__builtin_add_overflow()` and `__builtin_mul_overflow()` 

There appears to be a similar case in the Module 17 UI code as well.

I'll make a PR, although I can't test the case for the Module 17 on real hardware.

ranguli:
This PR fixes issue #444. The offset bug can occur in two separate cases, with the default UI as well as the UI variant when `CONFIG_UI_NO_KEYBOARD` is set.

To prevent a wraparound, the `__builtlin_mul_overflow` and `__builtin_add_overflow` macros from the compiler are used. These are supported in GCC and Clang so should not pose any portability issues for any of our targets.

To-Do: Validate if the fix in the `CONFIG_UI_NO_KEYBOARD` case works

### DIFF
--- a/openrtx/src/ui/default/ui.c
+++ b/openrtx/src/ui/default/ui.c
@@ -1238,14 +1238,18 @@ static void _ui_numberInputKeypad(uint32_t *num, kbd_msg_t msg)
     // Update reference values
     ui_state.input_number = *num % 10;
 #else
-    // Maximum frequency len is uint32_t max value number of decimal digits
-    if(ui_state.input_position >= 10)
-        return;
-
     // Get currently pressed number key
     uint8_t num_key = input_getPressedNumber(msg);
-    *num *= 10;
-    *num += num_key;
+
+    uint32_t tmp = 0;
+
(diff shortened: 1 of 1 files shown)


silseva:
If it works with normal keyboard, then I'd consider it good to go: the version for `CONFIG_UI_NO_KEYBOARD` is similar to the other one.

PR #456, opened 2026-04-28
NVM: Sliced VFO storage API

mdiepart:
This PR introduces a VFO storage API based on a structure slicer, also introduced in this PR. 

The structure slicer is a generic API that gets configured with a structure size and a slice size. The size contains a header (currently only a crc8). The crc8 helps in detecting slices of incorrect size, or incorrect structure sizes.

Based on the slicer, a VFO slicer is introduced. It keeps a copy of the VFO from NVM to compare it at write time and to write only the slices that contains changed bytes.

Currently, the VFO slicer API stores each slice in EEEPROM at the address corresponding to the slice index.

Currently a draft, the code compiles but is not tested yet. Reviews and comments are more than welcome.


mdiepart:
I tested the implementation and squashed a few bugs, I think the code is ready for review :)

openrtx/src/core/struct_slicer.c
```diff
@@ -0,0 +1,147 @@
...
+#include <stdint.h>
+#include <string.h>
+
+#define MIN(a, b) ((a) < (b) ? (a) : (b))
+
+typedef struct {
+    uint8_t crc8;
+} __attribute__((packed)) slice_header_t;
+
+uint8_t crc8(const uint8_t *data, const size_t len)
```
silseva: This can go here, so that in the future other modules can reuse it: https://github.com/OpenRTX/OpenRTX/blob/master/openrtx/include/core/crc.h

openrtx/src/core/struct_slicer.c
```diff
@@ -0,0 +1,147 @@
...
+            if (crc & 0x80)
+                crc = (crc << 1) ^ 0x31;
+            else
+                crc = (crc << 1);
+        }
+    }
+    return crc;
+}
+
+int struct_slicer_init(size_t struct_size, size_t slice_size,
```
silseva: I think we can omit this function, as one can simply write
```
static const struct_slicer_config_t config = {
    .struct_size = sizeof(struct something),
    .slice_size = 64,
};
```

openrtx/src/core/struct_slicer.c
```diff
@@ -0,0 +1,147 @@
...
+#include "core/utils.h"
+
+#include <stddef.h>
+#include <errno.h>
+#include <stdint.h>
+#include <string.h>
+
+#define MIN(a, b) ((a) < (b) ? (a) : (b))
+
+typedef struct {
```
silseva: If you change this to
```
struct  slice {
    uint8_t crc;
    uint8_t data[];
}__attribute__((packed)) ;
```
you can probably make your life a bit easier when implementing the functions because you can use the `data` field to access the payload without having to do pointer math.
See this: https://en.wikipedia.org/wiki/Flexible_array_member

openrtx/src/core/struct_slicer.c
mdiepart: Yeah, make sense, I'll do that

openrtx/src/core/struct_slicer.c
mdiepart: Thank you for the tip

mdiepart:
I addressed your reviews in separate fixup commits. I also removed the typedefs and fixed some minor compilation warnings. Let me know if you prefer to review the rebased version.

mdiepart:
Last commit is small but important fix: the vfo slicer expected nvm_read to return the number of bytes read but this is only the case in posix_file devices so the slicer worked on linux but not on handhelds. I'll see if I can update posix_file devices to be consistent with the rest of NVM devices (probably directly as commits to persistence branch).

openrtx/include/core/vfo.h
```diff
@@ -0,0 +1,58 @@
...
+    uint16_t version;
+    channel_t vfo;
+} __attribute__((packed));
+
+struct vfo_slicer {
+    int nvm_dev;
+    int nvm_part;
+    size_t entry_size;
+    uint16_t version;
+    struct struct_slicer_config slicer_conf;
```
silseva, line 23: This can probably be made constant and put inside `vfo.c`:
```
static const struct struct_slicer_config config = {
    .struct_size = sizeof(struct vfo),
    .slice_size = VFO_SLICE_SIZE,    // Eventually CONFIG_VFO_SLICE_SIZE
};
```
Or there is need to set the slice size dynamically?

openrtx/include/core/vfo.h
silseva, line 1: I'd call the functions `vfo_initStorage`, `vfo_load` and `vfo_save`, similarly to the settings.

openrtx/include/core/vfo.h
mdiepart: The idea was to avoid having defines outside the code path, I find it makes it more difficult to understand what is going on.

openrtx/include/core/vfo.h
mdiepart: The idea was to leave open the possibility of having different ways of saving the vfo for other platforms and thus not hog the generic name `vfo_save`, but I can make the change if you prefer it this way

openrtx/src/core/struct_slicer.c
```diff
@@ -0,0 +1,124 @@
...
+    const size_t copy_len = MIN(config->struct_size - struct_offset,
+                                DATA_LENGTH(config));
+
+    memcpy(s->data, structure + struct_offset, copy_len);
+
+    if (copy_len < DATA_LENGTH(config))
+        memset(s->data + copy_len, 0, DATA_LENGTH(config) - copy_len);
+
+    // Compute CRC over data content, not including zero padding or header
+    slice[0] = crc8(s->data, copy_len);
```
silseva:
```suggestion
    s->crc8 = crc8(s->data, copy_len);
```

openrtx/src/core/struct_slicer.c
```diff
@@ -0,0 +1,124 @@
...
+
+    struct slice *s = (struct slice *)slice;
+
+    // Offset of slice in struct
+    size_t struct_offset = index * DATA_LENGTH(config);
+    size_t length = MIN(config->struct_size - struct_offset,
+                        DATA_LENGTH(config));
+
+    // Check that the crc is correct
+    if (slice[0] != crc8(s->data, length))
```
silseva:
```suggestion
    if (s->crc8 != crc8(s->data, length))
```

openrtx/src/core/struct_slicer.c
mdiepart: Done, thanks!

openrtx/src/core/struct_slicer.c
mdiepart: Done, thanks!

openrtx/include/core/vfo.h
silseva: Well... I didn't take into account this possibility. Do you mean, for example, to read/write the entire VFO data structure from a file or EEPROM memory and use slicing in case of flash?

mdiepart: Yes, or if we want to go further with the linux implementation

mdiepart:
As a reminder, this PR only adds the API to the codebase and no device uses it yet. I'm waiting for the PR still open to be merged before opening a PR with the implementation on all the devices.

LGTM.

silseva:
Did you squash the fixup commits? If so, you may not have pushed your changes upstream

mdiepart:
Not yet, i was waiting for you to rebase the persistence branch. I'll do that quickly

silseva:
Ah, sorry. I did the rebase and force-push of the `persistence` branch, in the meantime

mdiepart:
There we go :)

PR #461, opened 2026-05-19
rtx: split txDisable into hard channel gate and PTT gate

turnrye:
This change breaks txDisable out into two distinct concepts: txDisable as a function of the rtx state, and pttDisable as a function of the UI state. pttDisable is the same as the previous txDisable definition, but txDisable is now a more relaxed definition so that features like packet modes can be sent from areas where you wouldnt want the PTT button to work.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1151,3 +1151,8 @@ test('minmea conversion Test', minmea_conversion_test)
 test('UI Check Standby Test', ui_check_standby_test)
 test('M17 Packet Frame Test', m17_packet_test)
 test('DSP Oversampling Test', dsp_oversampling_test)
+
+rtx_tx_disable_test = executable('rtx_tx_disable_test',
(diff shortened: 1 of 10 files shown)


silseva:
All good, but there is a small loophole: if `txDisable` is set but `pttDisable` is not, you can still transmit by pressing the PTT. It'd be good to have some code forcing `pttDisable` to true when `txDisable` is set, for example in this area https://github.com/OpenRTX/OpenRTX/blob/8895b268317e516674637812d570176f674a36b8/openrtx/src/rtx/rtx.cpp#L105

Now this is being done in the UI thread, but is somehow a weak enforcement because is being done outside the RTX thread.

turnrye:
Good catch, will fix that and the formatting errors!

turnrye:
Ok @silseva ready again!

PR #462, opened 2026-05-19
voicePrompts: Broken PR

BrotRooti:
<s> Expect a bigger header, so that the number of existing voicePrompts can be included. --

This is the start of the C Side for [Task 726](https://tasks.openrtx.org/project/openrtx/us/726).

The other PR is [here](https://github.com/OpenRTX/vpcgen/pull/5) </s>


BrotRooti:
I don't know what the hell happend to this PR, looks like it bugged out?

PR #463, opened 2026-05-19
NVM: switch to new APIs

mdiepart:
This PR implements the new NVM APIs (settings and VFO) to all the devices. However, I am missing some information regarding some platforms.

Once all the platforms will have been switched to the new APIs, I'll rework further the nvmem files by trying to get rid of the functions from the nvmem files to call the proper vfo load/save functions directly from the code calling the previous functions. Each device will be able to pick a storage method (A/B partition, sliced structures, flat storage, ...) through config macros.

Here is the plan (and what I need)

- [ ] GDx devices: I don't have a clear view on what the memory layout looks like on the device, as such I don't really know what partitions to create. The plan is to use flash for calib data (written at factory) and CPS, and EEPROM for settings and vfo (need help from @silseva)
- [x] Mod17: We need to carve out some space for the VFO and settings. Actually we need to decide if we want a VFO altogether.  The flash page size is not uniform.
- [ ] ttwrplus: I have no idea what is going on here and I'll need guidance from @n1zzo 

And then:
- [ ] Decide on the CONFIG macros to use and how as well as where the NVM APIs should be initialized
- [ ] Remove functions `nvm_readSettings`, `nvm_writeSettings`, `nvm_readVfoChannelData`, `nvm_writeSettingsAndVfo`


mdiepart:
I still don't really know the memory areas that we can use on the GDx devices, but I took a stab anyway.
Also, after a quick chat with @silseva, we realized that EEPROM can easily be used to store vfo without hassle if we take a slice size the size of the vfo (plus the overhead of the structure slicer). This way, flat storage is handled by the current slicer.

mdiepart:
@silseva could you have a look at the proposed implementation for GDx family?
Also, regarding Module17, if we save both VFO and settings, I see several possibilities:
- easy option: take 4 sectors of 128k at the end of the flash space. This still leaves 512K of flash and we currently use 200K
- super space saving option: use sectors 1 and 2 (16K each) for VFO, and sectors 3 (16K) and 4 (64K limited to 16K) for settings
- middle-ground: 2 last 16K sectors for vfo, 2 128K sectors for settings

Any other option you would like to suggest?

silseva:
Yes, I already had in mind to do so. I also brought my DM-1801 from Italy, so that I can test the persistence implementation on the real hardware.

For Module17 I have to think a bit: the second and third options are the ones I prefer, actually. Given that settings and VFO persistence are not going to be changed that often on these devices, is probably better to use the small sectors so that the majority of the memory is left to the program.

mdiepart:
I had a thought about the Module17 again. I think it is better not to mix sector sizes because the flash driver is not designed to handle this. I'll try implementing option 3 for now.

mdiepart:
We may encounter a problem: if we carve out some space in the middle of the flash area, we may have the following problems:
- The first sector will be almost completely empty. The only section that I am able to store in that sector is the isr vector (mandatory at address 0) and other bits and bobs for 420B. We can start picking parts of the code based on the files and so on but this will make for a pretty ugly linker file.
- Flashing a new firmware will most probably (at least in the current configuration) erase the carved-out memory area, meaning either the VFO or the settings will be erased at each firmware update

I suggest something else: 
We use the last two sectors for an EEEPROM and store the settings and the calibration data as whole entries. This means that module17 will need a custom implementation for settings (and calibration data, also because there is no common implementation for that) and that settings are less reliably stored on Module17 than on other platforms. In the end I do not think that this is a real problem. We are also doing with what we are given (no external flash or eeprom).

mdiepart:
Maybe this PR deserves a little update:

Apart from T-TWR Plus (@n1zzo if you have some time for this, it would help me a lot :) ), an NVM implementation have been proposed for all devices. 

A few updates were required here and there and are included in this PR:
- Because the VFO slicer is mainly designed to work over an EEEP device, and because of the way EEEP works, I needed to implement an erase function for EEEP. This way, we avoid having stale entries in the EEEP that will get dragged on forever
- In order to support having a VFO slicer on a regular EEPROM (GDx family of devices), the vfo slicer can be used with a slice the size of the VFO (so one single slice). This needs to take the vfo slicer and struct slicer overhead into account hence the related commits
- Mod17 is also a different target because it has no external memory and the only part of the flash that can be reasonable used is the end of the flash which has 128K pages. So we create an EEEPROM with two 128K pages and three partitions. We use the EEEPROM to store the vfo using the regular implementation and in addition to that, the settings and calib data are stored in their own EEEP partition along with a CRC and length for good measure. 

Implementations were quickly tested for: Module17, MD-UV380, CS7000 Plus.

silseva:
Can this PR be considered good to go? :smile: 
I'll smooth the remaining sharp corners, if needed.

mdiepart:
More than rounding sharp edges, I am interested in having a feedback for targets I do not own :)
Also, can the persistence branch be rebased on master so we make sure that there is indeed no compilation problem?

mdiepart:
This last commit allows the CS7000 radios to migrate old settings from v0.4.4. It was tested and validated on CS7000 Plus. I expect this should work on a cs7000 too.
@silseva Should we merge this pr?

PR #466, opened 2026-05-25
fix: unable to build new devcontainer

BrotRooti:
On March 31st the [miosix repo](https://github.com/fedetft/miosix-kernel.git) relocated their tools directory. This resulted in new devcontainers unable to build. (I don't know how some uncached ones still were able to build in this repo).

I updated the path in the Dockerfile and it now works again.


silseva:
Thank you!

PR #468, opened 2026-05-25
voicePrompts: Start work for Task 726

BrotRooti:
Expect a bigger header, so that the number of existing voicePrompts can be included.

This is the start of the C Side for [Task 726](https://tasks.openrtx.org/project/openrtx/us/726).

The other PR is [here](https://github.com/OpenRTX/vpcgen/pull/5)

### DIFF
--- a/openrtx/src/core/voicePrompts.c
+++ b/openrtx/src/core/voicePrompts.c
@@ -178,7 +178,7 @@ static inline bool checkVpHeader(const vpHeader_t *header)
 {
     return ((header->magic == VOICE_PROMPTS_DATA_MAGIC)
             && (header->version == VOICE_PROMPTS_DATA_VERSION)
-            && (header ->numPrompts == VOICE_PROMPTS_DATA_NUMBER))
+            && (header->numPrompts == VOICE_PROMPTS_DATA_NUMBER));
 }
 
 /**
--- a/openrtx/src/core/voicePrompts.c
+++ b/openrtx/src/core/voicePrompts.c
@@ -23,6 +23,8 @@
 
 static const uint32_t VOICE_PROMPTS_DATA_MAGIC = 0x5056;   //'VP'
 static const uint32_t VOICE_PROMPTS_DATA_VERSION = 0x1000; // v1000 OpenRTX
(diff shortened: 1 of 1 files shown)


BrotRooti:
I am sorry for opening another PR to the same topic, but it looks like the other one somehow broke and destroyed my local git history with it

silseva:
I didn't forget, I'll add this to the next release

openrtx/src/core/voicePrompts.c
```diff
@@ -174,7 +177,8 @@ static void fetchCodec2Data(uint8_t *data, const size_t offset)
 static inline bool checkVpHeader(const vpHeader_t *header)
 {
     return ((header->magic == VOICE_PROMPTS_DATA_MAGIC)
-            && (header->version == VOICE_PROMPTS_DATA_VERSION));
+            && (header->version == VOICE_PROMPTS_DATA_VERSION)
+            && (header->numPrompts == VOICE_PROMPTS_DATA_NUMBER));
```
silseva, line 181: Instead of checking for header consistency, which will force us to update `VOICE_PROMPTS_DATA_NUMBER` every time a new prompt entry is added, I'd move the check into [`vp_queuePrompt`]( https://github.com/OpenRTX/OpenRTX/blob/1b31ba1b08b5d732ead34fa5d22d49202e5710e6/openrtx/src/core/voicePrompts.c#L385).
The idea is:
```
void vp_queuePrompt(const uint16_t prompt)
{
    if (prompt >= header->numPrompts)
        return;

    [...]
}
```

In theory it should work, but needs to be tested. Since you already did most of the work, I prefer you add this change yourself and retain the authorship of the patch.

silseva (changes requested):

openrtx/src/core/voicePrompts.c
silseva: Correction: I re-read what I wrote in the Taiga task, and what it'd be nice to have is `fallback to spelled text in case a string is not present`: basically what is done by `vp_queueString` if `userDictLookup` returns zero. At this point I'm not sure whether adding the check in `vp_queuePrompt` is actually what we want.

PR #478, opened 2026-06-02
drivers: NVM: fix eraseSector return type

### LINKED ISSUE #475: [BUG] Flash erase returns "success" on error paths
- **Location:** `platform/drivers/NVM/flash_stm32f4.c` (`eraseSector`)
- **Issue:** Function returns `bool` but error branches return negative errno values; non-zero evaluates true.
- **Impact:** Failed erase can be interpreted as success.
- **Fix direction:** Return strict `false/true` for bool API or convert to
  `int` API consistently.

tl;dr this function has either the wrong type or the wrong return values: currently always returns true (the error codes are converted to 1)

https://github.com/OpenRTX/OpenRTX/blob/c658da2897b0746b5d146801414d718bdc48b3c1/platform/drivers/NVM/flash_stm32f4.c#L63-L87

bbehnood:
eraseSector() is declared as bool but returns negative errno values on failure. Since non-zero values convert to true, error conditions are reported as success. The only caller treats the return value as an errno-style integer (`ret < 0`), indicating that the intended API returns negative error codes on failure and 0 on success.

Change the return type to `int` and return 0 on success.

Fixes #475.

### DIFF
--- a/platform/drivers/NVM/flash_stm32f4.c
+++ b/platform/drivers/NVM/flash_stm32f4.c
@@ -58,9 +58,9 @@ static inline bool unlock()
  * Erase one sector of the MCU flash memory.
  *
  * @param secNum: sector number.
- * @return true for successful erase, false otherwise.
+ * @return 0 on success, negative errno on failure.
  */
-static bool eraseSector(const uint8_t secNum)
+static int eraseSector(const uint8_t secNum)
 {
     if (unlock() == false)
         return -EIO;
@@ -83,7 +83,7 @@ static bool eraseSector(const uint8_t secNum)
 
     FLASH->CR &= ~FLASH_CR_SER;
 
-    return true;
+    return 0;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


bbehnood:
The failing build job appears to fail during devcontainer setup while extracting MPFR (downloaded/mpfr-4.0.2.tar.xz missing in the miosix toolchain install step). The unit tests, Zephyr build, and formatting checks all pass. I don't think the failure is related to the changes in this PR, but please let me know if you'd like me to investigate further.

mdiepart:
Hi, thank you for your PR. It looks like this indeed slipped through the review when this driver was merged. 

For the sake of consistency, the `eraseSector` function in file `flash_stm32h7.c` should also be modified to return an integer and functions calling it should be modified to expect an integer instead of bool. Can you add this change?

bbehnood:
Thanks for the review!  I'll update that and adjust its callers as well.

bbehnood:
Updated as requested. The STM32H7 driver now uses errno-style return values throughout the erase path and the caller has been updated accordingly.

silseva (approved):

PR #481, opened 2026-06-07
Message registry and inbox

turnrye:
OpenRTX aspires to support multiple digital modes (M17, APRS, etc.) that can all send and receive messages. Rather than building a separate inbox per protocol, the message registry provides a single unified inbox that any protocol can plug into, and packet_io handles the thread-safe handoff between the UI and the radio hardware without coupling them directly.

# Message registry
This is the protocol-agnostic multiplexer; it owns no message storage, but instead holds borrowed pointers into each source's pool

* Sources register at compile time via a message_type_vtable_t (function pointer table) + opaque context
* Required vtable slots: count(), get(i)
* Optional slots: tick(), send(), supported_actions(), invoke_action(), start_compose()
* Each tick:
1. Calls every source's tick() (drain radio data first)
2. Sweeps all sources via count()/get() into a flat snapshot[]
3. Stable insertion-sorts by timestamp descending (newest first)
* messages_get(idx) returns borrowed pointers — valid only until the next tick
* Compose/reply routing: canCompose(mode) / startCompose(mode) match on vtable->mode_id
* Introduce the `CONFIG_MSG_SNAPSHOT_SIZE` concept, but with a placeholder for now; as platforms are registered and integrated, they should register what their needs are and simply add it in place of the current placeholder

# packet_io
This is the thread-safe handoff layer; it decouples the UI thread (specifically the message logic) from RTX thread, where protocols actually take place

* Two single-slot queues, shared mutex:
    * TX queue (pkt_tx_request_t): UI → RTX; carries outbound body, destination, mode, tag
    * RX queue (pkt_rx_event_t): RTX → UI; carries inbound messages and TX completion/failure notifications
* TX flow: send() vtable → packet_io_enqueue_tx() → RTX transmits → packet_io_enqueue_rx() (completion) → source tick() updates inbox entry status
* RX flow: RTX decodes packet → packet_io_enqueue_rx() → source tick() dequeues, creates inbox entry → visible in same-cycle snapshot

Some key goals driving the approach were:

* Be able to handle a diverse set of protocols with different messaging needs
* Each protocol source is responsible for its own storage, as they are expected to be diverse both in size and behavior (e.g. aprs packets are small and rather uniform in length, m17 sms can be large but vary in length)
* Respect OSI layering and separation of concerns from UI vs other threads
* Be mindful of impacts on RAM since messages can get quite large and add up quickly
* Don't add any specific platforms yet, but do add a demo that is accessible from the linux build
* Don't be responsible for what the send/reply message ui looks like, because that's too platform specific


openrtx/include/core/datetime.h
```diff
@@ -57,6 +57,15 @@ datetime_t localTimeToUtc(const datetime_t local_time, const int8_t timezone);
...
 void realignTimeInfo(datetime_t *time);
 
+/**
+ * @brief Compare two datetime_t timestamps.
+ *
+ * @param a: pointer to first datetime_t.
+ * @param b: pointer to second datetime_t.
+ * @return negative if a < b, zero if equal, positive if a > b.
+ */
+int datetime_cmp(const datetime_t *a, const datetime_t *b);
```
silseva, line 67: I'd call it `compareTimeInfo` for consistency with the other function

openrtx/include/core/MessageRegistry.hpp
```diff
@@ -0,0 +1,155 @@
...
+    /** Maximum snapshot entries across all sources.
+     *  Configured per target in hwconfig.h via CONFIG_MSG_SNAPSHOT_SIZE. */
+    static constexpr size_t MAX_MESSAGES_SNAPSHOT = CONFIG_MSG_SNAPSHOT_SIZE;
+
+    /**
+     * @brief Compare two datetime_t values chronologically.
+     *
+     * @return negative if a < b, zero if equal, positive if a > b.
+     */
+    static int datetimeCmp(datetime_t a, datetime_t b);
```
silseva, line 54: Is this still needed?

openrtx/include/core/MessageRegistry.hpp
```diff
@@ -0,0 +1,155 @@
...
+
+    /**
+     * @brief Initialise the registry with a fixed, compile-time source array.
+     *
+     * @param sources: pointer to an array of SourceEntry structs; must remain
+     *                 valid for the lifetime of the registry.
+     * @param count:   number of entries in the array.
+     */
+    void init(const SourceEntry *sources, size_t count);
```
silseva, line 63: I'd use C++ constructors, if possible

openrtx/include/core/MessageRegistry.hpp
```diff
@@ -0,0 +1,155 @@
...
+
+    /**
+     * @brief Clear the snapshot.  Source array pointer is released.
+     */
+    void terminate();
```
silseva, line 68: Same as above, this can be a destructor

silseva:
General note: please do not `typedef` `enum`s and `struct`s, they are types by themselves. The codebase still contains many of these, but I'm progressively removing them.

turnrye:
Thank you for the feedback @silseva! I'm going to focus on getting https://github.com/OpenRTX/OpenRTX/pull/427 landed first, then shift focus back over to here once that's landed. That way I can avoid rebase hell.

PR #491, opened 2026-07-04
CI hardening: branch names with slashes, devcontainer cache, Miosix pin

### LINKED ISSUE #469: Github build action needs to clone a specific version of miosix-kernel repo
The current github build action doesn't run properly anymore, the upstream miosix kernel repository moved the `_tools` folder to `tools` (https://github.com/fedetft/miosix-kernel/commit/82559563afc1364c1bb569d4f1577881276ac905) and it seems like the `gcc-9.2.0-mp3.2 ` subfolder also does not exist anymore.

In the build action, we need to either refer to a specific commit of the miosix kernel repository or install the toolchain by following the instructions on the miosix wiki (https://miosix.org/wiki/index.php?title=Linux_Quick_Start#Install_the_Miosix_Toolchain) ?

slintak:
Sorry for opening two PRs at once as my first contributions, but I wouldn't
sleep well leaving this unfinished :)

Both failures I hit on #490 were CI issues. First the build job failed because
my branch name contained a slash and the artifact renaming `mv` treated it as
a directory path. Then the re-run died downloading a toolchain tarball while
rebuilding the devcontainer image from scratch. The rebuild happens because
the ghcr cache image is only refreshed when devcontainer files change (last on
May 26), while the `ubuntu:24.04` base moved on Jul 2, so every run since gets
a cache miss and rebuilds the Miosix toolchain (about 40 minutes).

Commits:

**ci: replace slashes in branch name used for artifact file names**
This PR intentionally comes from a branch with a slash, as a live test.

**devcontainer: pin miosix-kernel to Miosix v3.01** (fixes #469)
Fetch a fixed SHA instead of upstream HEAD, bump deliberately.

**ci: rebuild devcontainer image on a weekly schedule**
Keeps the layer cache in sync with the floating base image.

**ci: cancel superseded PR runs and add job timeouts**
A push to a PR cancels the previous run; branches are never cancelled.

Happy to adjust or drop any of these. Side note: `openrtx_md9600_wrap` is
compiled by CI but never uploaded as an artifact. Is that intentional?

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -37,8 +37,15 @@ RUN apt-get update -y && apt-get install -y --no-install-recommends \
   && rm -rf /var/lib/apt/lists/*
 
 # compile miosix compiler
-RUN git clone --depth 1 https://github.com/fedetft/miosix-kernel.git \
-  && cd miosix-kernel/tools/compiler/gcc-9.2.0-mp3.2 \
+# Pinned to Miosix v3.01: upstream layout changes have broken this build
+# before (see issue #469), bump the SHA deliberately.
+ARG MIOSIX_KERNEL_SHA=361e8c35b09ad73c15ded5332dff21ce8c9a711d
+RUN git init miosix-kernel \
+  && cd miosix-kernel \
+  && git remote add origin https://github.com/fedetft/miosix-kernel.git \
+  && git fetch --depth 1 origin ${MIOSIX_KERNEL_SHA} \
+  && git checkout FETCH_HEAD \
+  && cd tools/compiler/gcc-9.2.0-mp3.2 \
   && sh download.sh \
   && bash install-script.sh -j`nproc`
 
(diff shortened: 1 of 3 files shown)


silseva:
> Sorry for opening two PRs at once as my first contributions, but I wouldn't sleep well leaving this unfinished :)

No problem at all! Actually I prefer to have many small PRs, each one bringing in a well-defined set of fixes or features, than a big, multi-topic one.

> Both failures I hit on #490 were CI issues. First the build job failed because my branch name contained a slash and the artifact renaming `mv` treated it as a directory path. Then the re-run died downloading a toolchain tarball while rebuilding the devcontainer image from scratch. The rebuild happens because the ghcr cache image is only refreshed when devcontainer files change (last on May 26), while the `ubuntu:24.04` base moved on Jul 2, so every run since gets a cache miss and rebuilds the Miosix toolchain (about 40 minutes).

Yep, both problems were lying around for a while...

> Side note: `openrtx_md9600_wrap` is compiled by CI but never uploaded as an artifact. Is that intentional?

Yes, is intentional. The MD-9600 is still missing the RTX driver (Marco DM4RCO is working on it), and publishing the binaries wouldn't be really meaningful.

Many thanks for the fixes!

turnrye:
Yes thank you so much for helping to clean this up!!

PR #492, opened 2026-07-05
Message registry, inbox, and M17 SMS mode

turnrye:
This PR does a few things that are all part of the same overarching user feature: achieving the ability to send and receive M17 SMS messages. In anticipation of upcoming APRS messaging and potentially other messaging modes, like AIS, this starts by adding messaging multiplexing in what's referred to as a registry. Then, it creaetes a unified messaging inbox. Finally, it introduces the ability to send and receive M17 SMS messages.

Messages in the long run should be persisted to nvmem to avoid issues of having too much in RAM -- this is especially painful with M17 where messages may be up to 821 bytes long. For now, messages are kept in memory and we're just limiting how many we can receive to what is a reasonable tradeoff per hardware platform. We can revisit this once memory persistence is settled and we can decide how much storage to offer per device for messaging and whether to do it in a uniform way or let each mode determine. Future problems though.


turnrye:
This PR supersedes https://github.com/OpenRTX/OpenRTX/pull/481. This is draft for the moment while I address the code review feedback received there.

silseva:
@turnrye I mainlined the gfx work that was in the `release-v0.5.0` branch and rebased this one on top of the current master. Can you synchronize your repository and rebase this PR on top of the current `release-v0.5.0`, please?

Thank you!

openrtx/include/core/messages.h
```diff
@@ -0,0 +1,474 @@
...
+};
+
+/**
+ * @brief Vtable linking a message source to the registry.
+ *
+ * `count` and `get` are required (must be non-NULL for any registered
+ * source).  All other function pointers are optional; set to NULL if the
+ * source does not support that capability.
+ */
+struct message_type_vtable {
```
silseva: The more I work on the kernel, the more I get linuxized: do you mind renaming this struct into struct `message_ops`?

openrtx/include/core/messages.h
```diff
@@ -0,0 +1,474 @@
...
+     *
+     * Holds one of the `enum opmode` values from rtx/rtx.h stored as a
+     * plain uint8_t to avoid a layering dependency on that header.
+     * Use OPMODE_NONE (0) for sources that have no associated mode (e.g.
+     * demo stubs); such sources are never offered for compose or reply.
+     */
+    uint8_t mode_id;
+};
+
+#ifdef CONFIG_MESSAGES
```
silseva: I think is better to keep these headers and source files free from `#ifdef CONFIG_MESSAGES` and add these macros where necessary, for example in the UI code calling the message management functions.
Consider that we're doing garbage collection of unused functions and variables at link time.

openrtx/include/core/messages.h
```diff
@@ -0,0 +1,474 @@
...
+ * @return 0 on success, -ENOENT if no matching source, or the source's
+ *         send() return value on failure.
+ */
+int messages_send(uint8_t mode, const char *body, size_t body_len,
+                  const char *recipient);
+
+#else  /* !CONFIG_MESSAGES */
+
+/* ---------------------------------------------------------------------- */
+/* No-op stubs so callers never need their own #ifdef CONFIG_MESSAGES.     */
```
silseva: > No-op stubs so callers never need their own #ifdef CONFIG_MESSAGES.

How messy does it actually become the caller code if we move `#ifdef CONFIG_MESSAGES`?

openrtx/src/core/voicePrompts.c
silseva, line 1: `enableSpkOutput` acquires a path with `PRIO_PROMPT`, which is higher than `PRIO_BEEP`: this somehow makes sense because you want to give audio feedback to the user rather than playing a ringtone. To be discussed further.

openrtx/include/core/messages.h
turnrye: Not bad at all. Good idea.

openrtx/src/core/threads.c
```diff
@@ -61,10 +69,26 @@ void *ui_threadFunc(void *arg)
...
 
         vp_tick();                           // continue playing voice prompts in progress if any.
 
+#ifdef CONFIG_MESSAGES
+        // Refresh the message inbox snapshot; if new unread entries
+        // arrived, trigger the configured notification tone.
+        size_t new_unread = messages_tick();
+        if (new_unread > 0)
+        {
+            notification_play_message_tone(notif_type, notif_tone);
```
silseva, line 87:
```suggestion
            notification_play_message_tone(state.settings.notification_type, state.settings.msg_notification_tone);
```

openrtx/include/core/m17_sms.h
```diff
@@ -0,0 +1,50 @@
...
+
+/* Default sizing tier, used by most targets. Targets with more RAM (e.g.
+ * CS7000/CS7000-PLUS/linux) define larger values in their own hwconfig.h,
+ * which take precedence over these fallbacks. */
+#ifndef M17_SMS_MAX_MESSAGES
+#define M17_SMS_MAX_MESSAGES 32
+#endif
+#ifndef M17_SMS_POOL_BYTES
+#define M17_SMS_POOL_BYTES 3200
+#endif
```
silseva, line 21: Let's adopt the usual `CONFIG_` syntax for these macros as well. I'd also move these two `#ifndef` blocks to `m17_sms.cpp` to minimize their scope.

openrtx/include/core/MessageRegistry.hpp
```diff
@@ -0,0 +1,442 @@
...
+/**
+ * @brief A single message source entry: ops + opaque context.
+ *
+ * Build an array of these at compile time and pass it to the
+ * `MessageRegistry` constructor.  `ops` must be non-NULL and its `count`/
+ * `get` members must be non-NULL for every registered source; this is a
+ * configuration contract checked defensively (not asserted) by every
+ * registry method that dereferences it.
+ */
+struct SourceEntry {
```
silseva, line 27: I think this is more a `struct MessageSource`: the fact that is going to be an entry in a list is somewhat secondary

openrtx/include/core/MessageRegistry.hpp
```diff
@@ -0,0 +1,442 @@
...
+     * `snapshotActive` selects which buffer is currently visible to callers. */
+    SnapshotEntry snapshots[2][MAX_MESSAGES_SNAPSHOT] = {};
+    size_t snapshotLens[2] = { 0, 0 };
+    uint8_t snapshotActive = 0;
+
+    bool dirty = false;
+};
+
+/* ---------------------------------------------------------------------- */
+/* Inline implementation.
```
silseva, line 228: I'd move the implementation to `MessageRegistry.cpp` to keep the code easier to read. Or is there a specific reason to have an inline implementation?

openrtx/include/core/MessageRegistry.hpp
```diff
@@ -0,0 +1,442 @@
...
+inline bool MessageRegistry::tick()
+{
+    /* Give each source a chance to drain the packet_io RX queue and
+     * advance TX completion status before we rebuild the snapshot. */
+    bool any_changed = false;
+    for (size_t s = 0; s < sourcesLen; s++) {
+        if (sources[s].ops == nullptr || sources[s].ops->tick == nullptr)
+            continue;
+        if (sources[s].ops->tick(sources[s].ctx))
+            any_changed = true;
```
silseva, line 255:
```suggestion
        any_changed |= sources[s].ops->tick(sources[s].ctx);
```

openrtx/include/core/MessageRegistry.hpp
```diff
@@ -0,0 +1,442 @@
...
+        uint8_t source_idx;
+    };
+
+    const SourceEntry *sources = nullptr;
+    size_t sourcesLen = 0;
+
+    /* Double-buffered snapshot: two buffers are alternated on each rebuild.
+     * `snapshotActive` selects which buffer is currently visible to callers. */
+    SnapshotEntry snapshots[2][MAX_MESSAGES_SNAPSHOT] = {};
+    size_t snapshotLens[2] = { 0, 0 };
```
silseva, line 221: For some reason my brain reads `snapshotLens` as "snapshot lens", and I was wondering what a snapshot lens would be. Perhaps we can rename it to `snapshotSize` to make it more clear.

openrtx/include/core/MessageRegistry.hpp
```diff
@@ -0,0 +1,442 @@
...
+    dirty = false;
+
+    /* Build into the inactive buffer to avoid stale pointers during rebuild. */
+    uint8_t next = snapshotActive ^ 1;
+    snapshotLens[next] = 0;
+
+    /* Collect all entries from all sources into the next buffer. */
+    for (size_t s = 0;
+         s < sourcesLen && snapshotLens[next] < MAX_MESSAGES_SNAPSHOT; s++) {
+        if (sources[s].ops == nullptr || sources[s].ops->count == nullptr)
```
silseva, line 271: I see that the construct `if (sources[s].ops == nullptr || sources[s].ops->count == nullptr)` is used a lot in the code: this makes it a good candidate for a macro.

```
#define IS_ACTION_DEFINED(source, action) (##source.ops && ##source.ops->##action)

if (!IS_ACTION_DEFINED(sources[s], count))
    continue
```

openrtx/include/core/messages.h
```diff
@@ -0,0 +1,382 @@
...
+#include <stdbool.h>
+#include <errno.h>
+#include "hwconfig.h"
+
+/* Default snapshot tier, used by most targets. Targets with more RAM (e.g.
+ * CS7000/CS7000-PLUS/linux) define a larger CONFIG_MSG_SNAPSHOT_SIZE in
+ * their own hwconfig.h, which takes precedence over this fallback. */
+#ifndef CONFIG_MSG_SNAPSHOT_SIZE
+#define CONFIG_MSG_SNAPSHOT_SIZE 32
+#endif
```
silseva, line 53: I'd move this to `messages.cpp`

openrtx/include/core/MessageRegistry.hpp
```diff
@@ -0,0 +1,442 @@
...
+    /**
+     * @brief Mark the snapshot as dirty, forcing a rebuild on next tick().
+     *
+     * Call this from a source when it adds or removes entries so the UI
+     * can skip an unnecessary rebuild on ticks where nothing changed.
+     */
+    void markDirty();
+
+private:
+    /** One resolved snapshot slot: the borrowed header pointer plus which
```
silseva, line 207: I think you can have `struct Snapshot` and define
```
struct SnapshotList {
    struct Snapshot entries[MAX_MESSAGES_SNAPSHOT];
    size_t numEntries
};
```

Then in the code you have
```
struct SnapshotList lists[2];
struct SnapshotList *curr = &list[0];
struct SnapshotList *next = &list[1];
```
you access the current and next snapshot list by pointer and you `std::swap` them when needed. This should make the code a easier to read. For example, instead of having `snapshots[next][snapshotLens[next]].hdr` you have `next->entries[next->numEntries].hdr`

turnrye:
@silseva and I spoke today for a code review call on this PR. In general he agreed with the featureset, and we thought that it would be best to land this change and then land the aprs branch utilizing this capability. He requested some architectural shifts, mostly around promoting better isolation of concerns and reusability (which I believe will net fewer LoC). 

Next steps are, I will rework this branch per the feedback, but rather than raising this as one massive PR again for review, we'll slice the PR into ~3 so that we can go through the final code review rounds in github more easily. So, closing this PR, and Ill keep the branch alive on my fork. See you in a reworked PR shortly :)

PR #494, opened 2026-07-11
Add audio sink on linux

turnrye:
This change implements an audio sink for the linux emulator. This is helpful for testing and evaluating the sound design as well as accessibility features. It leverages SDL rather than a lower level library like libalsa, which offers better compatibility for future explorations likee android's ndk and web emscripten.

I did not implement a source, because that implementation seems to require more work (my naive try resulted in jitters and hangs, and its not critical for my testing today).

### DIFF
--- a/platform/drivers/audio/audio_sdl.cpp
+++ b/platform/drivers/audio/audio_sdl.cpp
@@ -1,0 +1,308 @@
+/*
+ * SPDX-FileCopyrightText: Copyright 2020-2026 OpenRTX Contributors
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+#include <stdlib.h>
+#include <math.h>
+#include <errno.h>
+#include <pthread.h>
+#include "SDL2/SDL.h"
+#include "audio_sdl.h"
+
+/*
+ * The playback driver uses the SDL queue API rather than an audio callback: the
+ * callback model would fire on an SDL-owned thread (and, under Emscripten's
+ * PROXY_TO_PTHREAD, on the wrong side of the worker/main-thread split), whereas
+ * SDL_QueueAudio maps cleanly onto the blocking data()/sync() contract expected
+ * by the stream manager.
+ */
+
+// Number of extra queued chunks tolerated before a playback sync() blocks. Keeps
+// output latency bounded to a few audio periods.
+#define PLAY_QUEUE_SLACK 2
+
+// Upper bound on the playback drain loops, in ms. Comfortably above any real
+// drain time, but ensures a suspended device (e.g. an Emscripten audio context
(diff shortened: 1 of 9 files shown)


marcoSchr:
I had a look over your code and it looks good.
If someone with a bit more knowledge about the audio system can look at `platform/drivers/audio/audio_sdl.cpp` then it should be fine.

I also tested it on Linux and it is working.

What would be a nice feature would be to set correct hints for the audio stream names.
This would then show up in the system audio control panel.
At the moment it shows up as `openrtx_linux - openrtx_linux`.

platform/drivers/audio/audio_sdl.cpp
```diff
@@ -0,0 +1,308 @@
...
+ * run on different threads, so the init must be race-free. pthread_once gives us
+ * that; the subsystem is intentionally never de-initialised, matching the
+ * process-lifetime nature of the audio devices.
+ */
+static pthread_once_t sdlAudioOnce = PTHREAD_ONCE_INIT;
+static int sdlAudioReady = 0;
+
+static void sdlAudioInit(void)
+{
+    sdlAudioReady = (SDL_InitSubSystem(SDL_INIT_AUDIO) == 0) ? 1 : 0;
```
silseva: You can add `SDL_INIT_AUDIO` here and have it initialised with the rest: https://github.com/OpenRTX/OpenRTX/blob/ea89d8d7e723a92601bbdb0cce89c28eaa30a5fc/platform/targets/linux/emulator/sdl_engine.c#L266

platform/drivers/audio/audio_sdl.cpp
```diff
@@ -0,0 +1,308 @@
...
+const struct audioDriver sdl_play_audio_driver = {
+    .start = sdlPlay_start,
+    .data = sdlPlay_data,
+    .sync = sdlPlay_sync,
+    .stop = sdlPlay_stop,
+    .terminate = sdlPlay_halt,
+};
+#pragma GCC diagnostic pop
+
+/*
```
silseva: If possible, please move the tone generator driver to a dedicated .h/.c files, here: https://github.com/OpenRTX/OpenRTX/tree/master/platform/drivers/tones

At some point we'll merge the "beep" API inside the audio subsystem, but for the time being let's keep them separate for consistency.

turnrye:
This is ready again for a review!

silseva:
Comments for future PRs:
- when applying clang-format to already existing files, please do a dedicated commit
- please avoid doing things in one commit and then moving them around in another one (like here https://github.com/OpenRTX/OpenRTX/pull/494/changes/2081a74fcfec23e498b1020382d6d78669d10767)

PR #497, opened 2026-07-17
ci: check UI string tables stay in sync with English

### LINKED ISSUE #496: CI: fail if a new UI source string isn't registered in the Weblate translation template
Following up from the discussion in #435 (comment: https://github.com/OpenRTX/OpenRTX/issues/435#issuecomment-5001298007).

## Summary

New UI strings can currently be added in code without also being registered in the Weblate translation template, so translations can silently lag behind. This proposes a CI check that fails when that happens.

## Approach

- Identify how source strings are currently registered for Weblate (translation template file, format TBC after reading the current i18n setup)
- Add a CI step that compares strings referenced in code against the template and fails if any are missing
- Advisory-only to start, so it doesn't block unrelated PRs while the check is being proven out

Will pick this up and open a PR once I've had a look at the current i18n/Weblate setup in the repo.

chrisdebian:
## Summary

Closes #496.

Adds a CI check that fails if a `*Strings.h` translation file is missing a field present in `EnglishStrings.h`. A missing field isn't just an incomplete translation — these are C designated initializers, so an unset field is implicitly a NULL `const char*`, and every consumer (`voicePromptUtils.c`, `ui_menu.c`, etc.) dereferences it directly. It's a crash risk the first time that string is shown or spoken, not merely a blank display.

An extra/stale field (present in a translation, removed from English) is also flagged, since it usually means dead content lingering after a rename.

## One thing worth knowing

While writing this I found `SpanishStrings.h` currently has all the same fields as English but in a different textual order (`metaText` sits earlier). I traced the actual runtime indexing in `voicePrompts.c` to check whether this matters — it doesn't: designated initializers are order-independent in C, and the real indexing depends only on `stringsTable_t`'s field order in `ui_strings.h`, not on how any individual translation file writes its initializer list. So the check deliberately treats order as informational only, not a failure — didn't want to make this stricter than the actual risk warrants.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -20,3 +20,11 @@ jobs:
         run: ./scripts/clang_format.sh --check
       - name: REUSE Compliance Check (confirm copyright and licenses)
         uses: fsfe/reuse-action@v6
+
+  check-strings:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Check UI string tables are in sync with English
+        run: python3 ./scripts/check_strings.py
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -26,5 +26,5 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
-      - name: Check UI string tables are in sync with English
+      - name: Check UI string tables are in sync with stringsTable_t and English
         run: python3 ./scripts/check_strings.py
--- a/openrtx/include/ui/SpanishStrings.h
+++ b/openrtx/include/ui/SpanishStrings.h
@@ -69,7 +69,6 @@ const stringsTable_t spanishStrings =
     .gpsSettings       = "Ajustes de GPS",
     .m17settings       = "Ajustes de M17",
     .callsign          = "Licencia",
-    .metaText          = "Meta Txt",
     .resetToDefaults   = "Reseatear a fábrica",
     .toReset           = "Para resetear:",
     .pressEnterTwice   = "Apretar Enter 2 veces",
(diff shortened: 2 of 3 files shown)


turnrye:
Hi @chrisdebian, thank you for the PR! Ill provide a quick review of this tomorrow. At a glance this seems like a sound contribution. I did want to clarify for you the framing though -- today none of this repo is wired in to weblate or gettext tooling. That work is going to take place as part of an upcoming UI rearchitecture. This doesnt change the fact that this CI tooling will provide some helpful checks to give contributors meaningful feedback in the meantime while the old tooling is still in place. More from me soon!

chrisdebian:
Thanks for the context, that's helpful to know — makes sense that this sits as an interim safety net rather than trying to anticipate the Weblate/gettext rearchitecture. Happy to adjust scope once that work starts if this check needs to change shape. Look forward to the review whenever you get a chance.

turnrye (changes requested):
This is solid! But I do think there's another check we can add to make this one step more. Right now, this will check to make sure that the translated strings stay in sync with oneanother. But, what we really care about the most is that those stay in sync with `stringsTable_t`.  I think this would be a small addition using a similar approach but cover the last remaining failure mode.

Also, can you add a second commit that fixes the string ordering in Spanish so that the build logs stay clean?

chrisdebian:
Thanks for the review — both addressed:

1. Added a first pass that checks every *Strings.h file (English included) against stringsTable_t's own field list in ui_strings.h, before the existing English-vs-translations pass. Verified it actually catches the failure mode you described: temporarily dropped metaText from EnglishStrings.h locally and confirmed the new check flags English itself as missing it (the old check alone would've just shown Spanish as having a stale extra field, misleadingly pointing at the wrong file).

2. Second commit reorders SpanishStrings.h's fields to match stringsTable_t's declared order — metaText was the one out of place. Build logs are clean now, no order note.

Both commits build/lint clean locally (clang-format + the script's own self-test).

PR #498, opened 2026-07-20
HD2: minimal OpenRTX bring-up — display, backlight, keyboard, RTC, battery, GPS

fenfir:
First on-hardware bring-up of the Ailunce HD2 (HR_C7000 / CK803S) target on the external miosix kernel (board hrc7000_hd2). Boots the stock OpenRTX UI with the basics working, live-verified:

  - display   ST7735S 160x128 RGB565 over the HW i8080 controller @0x12000000
  - backlight PWM ch0 @0x140c0000 (duty clamped below period: duty==period
              wraps to LED-off on this comparator, so 100% mapped to dark)
  - keyboard  4x4 matrix on the shared LCD pins + side keys + rotary knob
  - RTC       DesignWare I2C2 @0x14080000 slave 0x70; the master is enabled
              ONCE and left enabled (toggling IC_ENABLE per transfer corrupted
              the FIFO -> garbage read-back)
  - battery   on-chip ADC ch2 -> platform_getVbat (Li-ion 2S)
  - GPS       polled UART2 (DesignWare 16550 @0x14050000, 9600 8N1) -> NMEA
              ring buffer -> CONFIG_GPS; driven by gps_task(), no ISR

Radio / DMR / audio / M17 are not yet brought up. The driver-layer gaps link through OpenRTX's shared stubs (`platform/drivers/stubs/{cps_io,nvmem}_stub.c`); the two core entry points that have no shared stub — `vp_*` (voice prompts) and `rtx_*` (radio task) — are no-ops in `hd2_core_stubs.c`, each replaced by its real implementation as that subsystem lands. Build tree: `platform/targets/HD2` links the external miosix (`MIOSIX_KPATH`), toolchain `csky-miosix-elf`.

## Review feedback addressed

Reworked per @silseva's review of the original single commit:

  - **Split into 5 logical commits**: HR_C7000 MCU support → HD2 platform + build → display + backlight → keyboard → GPS.
  - **`mcu/CSKY_V2` → `mcu/HR_C7000`** — C-SKY V2 is the CPU architecture; HR_C7000 is the MCU, so the SoC support layer is named for the SoC.
  - **`hd2_regs.h` → `platform/mcu/HR_C7000/registers.h`**, rewritten as CMSIS-style peripheral structs cast onto the MMIO bases (SOCSYS / GPIO / LCD i8080 / PWM / I2C2 / ADC) instead of a flat block of `#define` addresses. Register names and offsets are taken from the HR_C7000 user guide.
  - **New `gpio_hrc7000` driver** under `platform/drivers/GPIO/` (native `gpio_setMode`/`setPin`/`clearPin`/`readPin` over the DW_apb_gpio banks), used in place of ad-hoc GPIO register writes. Pin-function muxing stays with the SOCSYS pad-mux, as this SoC has no per-GPIO alternate-function register.
  - **`hd2_stubs.c` removed** in favour of the shared `platform/drivers/stubs/` drivers; only the `vp_*` / `rtx_*` core entry points (which have no shared stub) remain, in `hd2_core_stubs.c`.

### DIFF
--- a/platform/drivers/ADC/adcHrc7000.c
+++ b/platform/drivers/ADC/adcHrc7000.c
@@ -1,0 +1,106 @@
+/*
+ * SPDX-FileCopyrightText: Copyright 2026 OpenRTX Contributors
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ * Driver for the HR_C7000 on-chip ADC (base 0x140d0000, manual 4.13),
+ * implementing the standard peripherals/adc.h interface.  Single-shot 10-bit
+ * conversions, busy-polled (no RTOS/semaphore).  On the Ailunce HD2 the battery
+ * pack sits on channel 2 (ADC_HW->DATA_CD[9:0]).
+ *
+ * The reset-release pulse in adcHrc7000_init() is essential: without it the
+ * conversion FSM stays held and DATA reads 0 (live-verified 2026-06-01; with it
+ * ch2 reads a stable ~0x350).
+ */
+
+#include "adcHrc7000.h"
+#include "registers.h"
+#include <pthread.h>
+#include <stdbool.h>
+
+#define ADC_STATE_BUSY 0x01u    /* ADC_CTRL_BUSY              */
+#define ADC_STATE_FSM 0x3eu     /* SAMP_FSM_STATE (bits[5:1]) */
+#define ADC_SAMPLE_MASK 0x3ffu  /* 10-bit result             */
+#define ADC_GUARD_ITERS 200000u /* bounded busy-wait guard   */
+
+/* Poll CTRL_STATE until the conversion FSM reaches the requested phase:
+ *   waitBusy = false -> BUSY + SAMP_FSM clear (conversion idle / complete);
+ *   waitBusy = true  -> BUSY or SAMP_FSM set (conversion has started).
+ * Returns 1 when the phase is reached, 0 on the bounded-guard timeout. */
+static int adc_wait(bool waitBusy)
+{
+    for (uint32_t guard = 0; guard < ADC_GUARD_ITERS; ++guard) {
+        bool busy = (ADC_HW->CTRL_STATE & (ADC_STATE_BUSY | ADC_STATE_FSM))
+                 != 0u;
+        if (busy == waitBusy)
+            return 1;
+    }
+    return 0;
+}
+
+int adcHrc7000_init(const struct Adc *adc)
+{
+    ADC_HW->CTRL = 0u;
+    ADC_HW->CTRL = 8u; /* PD_FORCE soft-reset (the missing step) */
+    ADC_HW->CTRL_STOP = 2u;
+    (void)adc_wait(false);
+    ADC_HW->CTRL = 0u; /* release reset -> normal operation */
+    ADC_HW->SEOC_TIME = 0xa20u;
+    ADC_HW->P2S_EN = 0u;
+    ADC_HW->INTR = 0u;
+    ADC_HW->CH_VLD = 0u;
+
+    if (adc->mutex != NULL)
+        pthread_mutex_init((pthread_mutex_t *)adc->mutex, NULL);
+
+    return 0;
+}
+
+void adcHrc7000_terminate(const struct Adc *adc)
+{
+    if (adc->mutex != NULL)
+        pthread_mutex_destroy((pthread_mutex_t *)adc->mutex);
+}
+
+uint16_t adcHrc7000_sample(const struct Adc *adc, const uint32_t channel)
+{
+    (void)adc; /* single ADC block on the HR_C7000 */
+
+    if (channel > 7u)
+        return 0;
+    if (!adc_wait(false))
+        return 0;
+
+    ADC_HW->CH_VLD = (1u << channel);
+    ADC_HW->START = 1u;
+
+    /* Wait for the FSM to actually enter BUSY before polling for completion.
+     * Best-effort (bounded): if the conversion already finished we fall through
+     * to the idle-wait below, which latches the just-finished result.  Without
+     * this the idle-wait can observe the still-idle FSM and return the PREVIOUS
+     * conversion's DATA -- on the first sample that is the reset value 0, read
+     * by the battery monitor as a flat pack (spurious low-battery alarm). */
+    (void)adc_wait(true);
+    if (!adc_wait(false))
+        return 0;
+
+    uint32_t data;
+    switch (channel >> 1) {
+        case 0:
+            data = ADC_HW->DATA_AB;
+            break; /* ch 0/1 */
+        case 1:
+            data = ADC_HW->DATA_CD;
+            break; /* ch 2/3 */
+        case 2:
+            data = ADC_HW->DATA_EF;
+            break; /* ch 4/5 */
+        default:
+            data = ADC_HW->DATA_GH;
+            break; /* ch 6/7 */
+    }
+    if (channel & 1u)
+        data >>= 16;
+
+    return (uint16_t)(data & ADC_SAMPLE_MASK);
+}
--- a/platform/drivers/ADC/adcHrc7000.h
+++ b/platform/drivers/ADC/adcHrc7000.h
@@ -1,0 +1,58 @@
+/*
+ * SPDX-FileCopyrightText: Copyright 2026 OpenRTX Contributors
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ * Driver for the HR_C7000 on-chip ADC (used on the Ailunce HD2 and other
+ * HR_C7000 targets), implementing the standard peripherals/adc.h interface.
+ * Single-shot 10-bit conversions, busy-polled.
+ */
+
+#ifndef ADC_HRC7000_H
+#define ADC_HRC7000_H
+
+#include "peripherals/adc.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * Define an instance of the HR_C7000 ADC device driver.  The chip has a single
+ * ADC block, so no peripheral base is needed (priv is unused).
+ *
+ * @param name: instance name.
+ * @param mutx: pointer to mutex for concurrent access, can be NULL.
+ * @param res:  conversion factor from counts to uV (see ADC_COUNTS_TO_UV).
+ */
+#define ADC_HRC7000_DEVICE_DEFINE(name, mutx, res)             \
+    extern uint16_t adcHrc7000_sample(const struct Adc *adc,   \
+                                      const uint32_t channel); \
+    const struct Adc name = {                                  \
+        .sample = &adcHrc7000_sample,                          \
+        .priv = NULL,                                          \
+        .mutex = mutx,                                         \
+        .countsTouV = res,                                     \
+    };
+
+/**
+ * Initialise the HR_C7000 ADC (the vendor reset-release pulse the conversion
+ * FSM needs; without it DATA reads 0) and its mutex, if any.
+ *
+ * @param adc: pointer to ADC device handle.
+ * @return zero on success.
+ */
+int adcHrc7000_init(const struct Adc *adc);
+
+/**
+ * Shut down the HR_C7000 ADC.
+ *
+ * @param adc: pointer to ADC device handle.
+ */
+void adcHrc7000_terminate(const struct Adc *adc);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* ADC_HRC7000_H */
--- a/platform/drivers/GPIO/gpio_hrc7000.h
+++ b/platform/drivers/GPIO/gpio_hrc7000.h
@@ -1,0 +1,92 @@
+/*
+ * SPDX-FileCopyrightText: Copyright 2020-2026 OpenRTX Contributors
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+#ifndef GPIO_HRC7000_H
+#define GPIO_HRC7000_H
+
+#include "peripherals/gpio.h"
+#include "registers.h"
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/*
+ * Native GPIO driver for the HR_C7000 (Ailunce HD2) DesignWare DW_apb_gpio
+ * banks GPIOA/GPIOB/GPIOC (see GPIO_TypeDef in registers.h): DR is the output
+ * latch, DDR the direction register (1 = output), EXT_PORT the input read-back.
+ *
+ * Standard pin-function muxing is handled by the SOCSYS pad-mux, not the
+ * per-GPIO registers, so gpio_setMode() covers only direction and level (see
+ * it for how ALTERNATE modes are treated).  The one exception is the vendor
+ * "alternate-function B" selector (per-bank register at +0x78, outside the
+ * DW_apb_gpio map); use gpio_setAltFuncB() for it.
+ */
+
+/**
+ * Configure gpio pin functional mode.
+ *
+ * @param port: gpio port (GPIOA/GPIOB/GPIOC).
+ * @param pin: gpio pin number, between 0 and 31.
+ * @param mode: functional mode (bit 7:0); the alternate-function field (bit
+ * 15:8) is ignored -- this SoC muxes pin functions via the SOCSYS pad-mux.
+ */
+void gpio_setMode(const void *port, const uint8_t pin, const uint16_t mode);
+
+/**
+ * Set gpio pin to high logic level.
+ *
+ * @param port: gpio port.
+ * @param pin: gpio pin number.
+ */
+static inline void gpio_setPin(const void *port, const uint8_t pin)
+{
+    ((GPIO_TypeDef *)port)->DR |= (1u << pin);
+}
+
+/**
+ * Set gpio pin to low logic level.
+ *
+ * @param port: gpio port.
+ * @param pin: gpio pin number.
+ */
+static inline void gpio_clearPin(const void *port, const uint8_t pin)
+{
+    ((GPIO_TypeDef *)port)->DR &= ~(1u << pin);
+}
+
+/**
+ * Read gpio pin's logic level.
+ *
+ * @param port: gpio port.
+ * @param pin: gpio pin number.
+ * @return 1 if pin is at high logic level, 0 if pin is at low logic level.
+ */
+static inline uint8_t gpio_readPin(const void *port, const uint8_t pin)
+{
+    return (((GPIO_TypeDef *)port)->EXT_PORT >> pin) & 1u;
+}
+
+/**
+ * Select the HR_C7000 vendor "alternate-function B" for a pin (per-bank
+ * register at +0x78).  This is distinct from the SOCSYS pad-mux and is sticky
+ * across gpio_setMode() direction changes; the keypad-matrix rows use it to
+ * sense while the same pins otherwise carry LCD-bus signals.
+ *
+ * @param port: gpio port (GPIOA/GPIOB/GPIOC).
+ * @param pin: gpio pin number, between 0 and 31.
+ */
+static inline void gpio_setAltFuncB(const void *port, const uint8_t pin)
+{
+    ((GPIO_TypeDef *)port)->ALT_FUNC_B |= (1u << pin);
+}
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* GPIO_HRC7000_H */
--- a/platform/drivers/backlight/backlight_HD2.c
+++ b/platform/drivers/backlight/backlight_HD2.c
@@ -1,0 +1,69 @@
+/*
+ * SPDX-FileCopyrightText: Copyright 2026 OpenRTX Contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ * HD2 LCD-backlight driver: the LED is driven by PWM channel 0 of the
+ * HR_C7000 PWM block (0x140c0000).  The channel-start register order
+ * replicates the vendor V2.1.3 pwm_channel_start sequence exactly.
+ *
+ * Owns the OpenRTX display_setBacklightLevel() entry point (the backlight
+ * driver owns the PWM; the display driver only owns the LCD bus).
+ */
+
+#include "interfaces/display.h"
+#include "registers.h"
+#include <stdint.h>
+
+#define BACKLIGHT_PWM_HZ 10000u /* perception-flat dimming frequency */
+
+static void pwm_program(PWM_Channel_TypeDef *p, unsigned freq_hz,
+                        unsigned duty_per_10000)
+{
+    /* Vendor V2.1.3 channel-start order: clear enable/mode bits, select the
+     * gated source, set period + duty, then re-enable. */
+    p->CFG &= ~1u;
+    p->CFG &= ~4u;
+    p->CFG &= ~8u;
+    p->CFG &= ~0x30u;
+    p->CFG |= 0x20u;
+    p->CFG |= 0x100u;
+    p->CFG |= 0x200u;
+    p->RPT = 5;
+    p->PER = freq_hz ? (PWM_TIMER_HZ / freq_hz) : 1u;
+    p->FP = (p->PER * duty_per_10000) / 10000u;
+    /* Never let duty == period: on this comparator that wraps to a
+     * permanently-LOW output (LED off) instead of "always on".  Full
+     * brightness (level 100 -> duty_per_10000 10000) must map to period-1. */
+    if (p->FP >= p->PER)
+        p->FP = p->PER - 1u;
+    p->CFG |= 4u;
+    p->CFG |= 1u;
+}
+
+void backlight_init(void)
+{
+    /* Restore the PWM ch0 duty-gating control words (STATUS @0x10 and the three
+     * reserved words @0x14-0x1c) to their V2.1.3 boot value.  Left at 0 (as the
+     * kernel/IAP hand-off leaves them for us) the duty cycle is squashed and the
+     * LED never lights, regardless of the period/duty we program below. */
+    PWM_CH0->STATUS = PWM_CH0_GATE_ON;
+    PWM_CH0->RESERVED_HW[0] = PWM_CH0_GATE_ON;
+    PWM_CH0->RESERVED_HW[1] = PWM_CH0_GATE_ON;
+    PWM_CH0->RESERVED_HW[2] = PWM_CH0_GATE_ON;
+}
+
+void backlight_terminate(void)
+{
+    pwm_program(PWM_CH0, BACKLIGHT_PWM_HZ, 0); /* duty 0 = dark */
+}
+
+void display_setBacklightLevel(uint8_t level)
+{
+    if (level > 100)
+        level = 100;
+
+    /* Keep the PWM running and zero the duty when off: halting the counter
+     * would hold the output pin at its last (possibly high) level, leaving
+     * the LED lit. */
+    pwm_program(PWM_CH0, BACKLIGHT_PWM_HZ, (unsigned)level * 100u);
+}
--- a/platform/drivers/display/ST7735S_HD2.c
+++ b/platform/drivers/display/ST7735S_HD2.c
@@ -1,0 +1,197 @@
+/*
+ * SPDX-FileCopyrightText: Copyright 2026 OpenRTX Contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ * ST7735S display driver for the Ailunce HD2 (160x128 RGB565), driven
+ * through the HR_C7000 hardware i8080 controller @0x12000000 (manual 5.3):
+ * a write to INDEX (+0x00) emits one command cycle, a write to DATA (+0x04)
+ * one 8-bit data cycle; the controller generates CS/RS/WR timing per WCFG.
+ *
+ * The controller owns PTC3-6 (CS/RS/WR/RD) and PTC7-14 (DB0-7) while the
+ * PTC pad mux (DIPLEX2) is in i8080 mode; the panel RESET (PTC2) stays a
+ * plain GPIO.  This driver establishes that mux at init and leaves it in
+ * i8080 mode as the resting state (the keyboard scan borrows the pins and
+ * restores i8080 mode when done).
+ *
+ * OpenRTX's graphics layer owns the single 40 KiB RGB565 framebuffer and
+ * pushes it here via display_renderRows(); this driver keeps no copy.
+ */
+
+#include "interfaces/display.h"
+#include "interfaces/delays.h"
+#include "hwconfig.h"
+#include "registers.h"
+#include "pinmap.h"
+#include "peripherals/gpio.h"
+#include <stdint.h>
+#include <stddef.h>
+#include "drivers/backlight/backlight.h"
+
+/* i8080 write-strobe timing (moved out of the SoC registers.h -- it belongs to
+ * the ST7735S driver, not the HR_C7000 register map).  The panel RESET pin
+ * (PTC2, GPIO, active-low) is driven via LCD_RST from pinmap.h. */
+#define LCD_WCFG_DEFAULT 0x020202u /* 2/2/2 strobes ~143 ns */
+
+/* ST7735S command opcodes used by the init sequence. */
+enum {
+    ST7735_SLPOUT = 0x11,
+    ST7735_DISPON = 0x29,
+    ST7735_CASET = 0x2a,
+    ST7735_RASET = 0x2b,
+    ST7735_RAMWR = 0x2c,
+    ST7735_MADCTL = 0x36,
+    ST7735_COLMOD = 0x3a,
+};
+
+static inline void send_cmd(uint8_t cmd)
+{
+    LCD_HW->INDEX = cmd;
+}
+static inline void send_data(uint8_t d)
+{
+    LCD_HW->DATA = d;
+}
+
+static void set_window(uint8_t x0, uint8_t x1, uint8_t y0, uint8_t y1)
+{
+    send_cmd(ST7735_CASET);
+    send_data(0);
+    send_data(x0);
+    send_data(0);
+    send_data(x1);
+    send_cmd(ST7735_RASET);
+    send_data(0);
+    send_data(y0);
+    send_data(0);
+    send_data(y1);
+}
+
+void display_init(void)
+{
+    /* Claim the PTC pads for the i8080 controller and make the panel-reset
+     * pin (PTC2) a GPIO output.  DIPLEX2 upper bits are write-only -- always
+     * write the whole-register constant, never read-modify-write.  GPIOC DDR
+     * marks PTC2..14 (reset + LCD bus) as outputs: this is the resting state
+     * the keyboard scan saves and restores around borrowing the shared pins. */
+    GPIOC->DDR = 0x00007ffcu;
+    SOCSYS->IO_DIPLEX2 = HD2_DIPLEX2_LCD_I80;
+    gpio_setMode(LCD_RST, OUTPUT);
+    LCD_HW->WCFG = LCD_WCFG_DEFAULT; /* 2/2/2 write strobes (~143 ns) */
+
+    /* Reset pulse (active-low) to ST7735S datasheet timing. */
+    LCD_HW->INDEX; /* (barrier: ensure mux write posted) */
+    gpio_setPin(LCD_RST);
+    delayMs(10);
+    gpio_clearPin(LCD_RST);
+    delayMs(10);
+    gpio_setPin(LCD_RST);
+    delayMs(120);
+
+    send_cmd(ST7735_SLPOUT);
+    delayMs(120);
+
+    /* Frame rate (vendor V2.1.3 values). */
+    send_cmd(0xb1);
+    send_data(5);
+    send_data(0x3c);
+    send_data(0x3c);
+    send_cmd(0xb2);
+    send_data(5);
+    send_data(0x3c);
+    send_data(0x3c);
+    send_cmd(0xb3);
+    send_data(5);
+    send_data(0x3c);
+    send_data(0x3c);
+    send_data(5);
+    send_data(0x3c);
+    send_data(0x3c);
+    send_cmd(0xb4);
+    send_data(3);
+
+    /* Power control. */
+    send_cmd(0xc0);
+    send_data(0x28);
+    send_data(8);
+    send_data(4);
+    send_cmd(0xc1);
+    send_data(0xc0);
+    send_cmd(0xc2);
+    send_data(0x0d);
+    send_data(0);
+    send_cmd(0xc3);
+    send_data(0x8d);
+    send_data(0x2a);
+    send_cmd(0xc4);
+    send_data(0x8d);
+    send_data(0xee);
+    send_cmd(0xc5);
+    send_data(0x1a);
+
+    /* Gamma (positive / negative), 16 bytes each. */
+    static const uint8_t gp[16] = {
+        0x04, 0x22, 0x07, 0x0a, 0x2e, 0x30, 0x25, 0x2a,
+        0x28, 0x26, 0x2e, 0x3a, 0x00, 0x01, 0x03, 0x13,
+    };
+    send_cmd(0xe0);
+    for (unsigned i = 0; i < sizeof gp; i++)
+        send_data(gp[i]);
+
+    static const uint8_t gn[16] = {
+        0x04, 0x16, 0x06, 0x0d, 0x2d, 0x26, 0x23, 0x27,
+        0x27, 0x25, 0x2d, 0x3b, 0x00, 0x01, 0x04, 0x13,
+    };
+    send_cmd(0xe1);
+    for (unsigned i = 0; i < sizeof gn; i++)
+        send_data(gn[i]);
+
+    send_cmd(ST7735_MADCTL);
+    send_data(0xa0); /* landscape, RGB order */
+    send_cmd(ST7735_COLMOD);
+    send_data(0x05); /* RGB565 */
+
+    /* Blank GRAM to black (full-window RAMWR + zeros). */
+    set_window(0, CONFIG_SCREEN_WIDTH - 1, 0, CONFIG_SCREEN_HEIGHT - 1);
+    send_cmd(ST7735_RAMWR);
+    for (size_t i = 0; i < (size_t)CONFIG_SCREEN_WIDTH * CONFIG_SCREEN_HEIGHT;
+         ++i) {
+        LCD_HW->DATA = 0;
+        LCD_HW->DATA = 0;
+    }
+
+    send_cmd(ST7735_DISPON);
+
+    backlight_init(); /* explicit init; the core sets the level (openrtx.c) */
+}
+
+void display_terminate(void)
+{
+    /* Framebuffer is owned by the graphics layer; nothing to free. */
+}
+
+void display_renderRows(uint8_t startRow, uint8_t endRow, void *fb)
+{
+    if (endRow > CONFIG_SCREEN_HEIGHT)
+        endRow = CONFIG_SCREEN_HEIGHT;
+    if (startRow >= endRow)
+        return;
+
+    set_window(0, CONFIG_SCREEN_WIDTH - 1, startRow, endRow - 1);
+    send_cmd(ST7735_RAMWR);
+
+    const uint16_t *p = (const uint16_t *)fb
+                      + (size_t)startRow * CONFIG_SCREEN_WIDTH;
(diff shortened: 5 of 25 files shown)


silseva:
Thanks for the PR!
Do you mind splitting it in more commits, please? Each commit should be a distinct, complete change touching a specific part. I'd have the following commits:
- glue code for HR_C7000 hardware (glue code, RTC, ADC, GPIO driver)
- HD2 platform support (hwconfig, main, platform.c, CMakelists, ...)
- display and backlight
- keyboard
- GPS

Also, the subfolder under MCU should be HR_C7000: CSKY_V2 is the architecture. The same is for `hd2_regs.h`: it should go under `mcu/HR_C7000` and calling it `registers.h` should suffice. For that, instead of having a list of `#define` directives with macros, you can create a struct with the list of registers and then cast the base address of the peripheral. See for example this: https://github.com/OpenRTX/OpenRTX/blob/ea89d8d7e723a92601bbdb0cce89c28eaa30a5fc/platform/mcu/CMSIS/Device/ST/STM32F4xx/Include/stm32f405xx.h#L431 and https://github.com/OpenRTX/OpenRTX/blob/ea89d8d7e723a92601bbdb0cce89c28eaa30a5fc/platform/mcu/CMSIS/Device/ST/STM32F4xx/Include/stm32f405xx.h#L1001

platform/drivers/keyboard/keyboard_HD2.c
```diff
@@ -0,0 +1,161 @@
...
+ */
+
+#include "interfaces/keyboard.h"
+#include "hd2_regs.h"
+#include <stdint.h>
+
+#define GPIO_DR(base) (*(volatile uint32_t *)((base) + 0x00u))
+#define GPIO_DDR(base) (*(volatile uint32_t *)((base) + 0x04u))
+#define GPIO_EXT_PORT(base) (*(volatile uint32_t *)((base) + 0x50u))
+#define GPIO_ALT_B(base) (*(volatile uint32_t *)((base) + 0x78u))
```
silseva: Please create a gpio_hrc7000 driver and use it instead of the macros. You'll find some examples on how to do it here: https://github.com/OpenRTX/OpenRTX/tree/master/platform/drivers/GPIO

platform/targets/HD2/hd2_core_stubs.c
silseva, line 1: You don't need this file. Please add to the CMakelists the already existing stubs instead: https://github.com/OpenRTX/OpenRTX/tree/master/platform/drivers/stubs

fenfir: I'm working on bringing up rtx now. I'm thinking to keep the vp_ stubs for now. Should I move them somewhere more central? 

I am still doing work on getting codec to up to speed. I'm at 20.6ms, trying to get it below 20ms. I still have some stuttering on the voiceprompts

fenfir: I also don't want to introduce a new dependency on my custom codec2-mod branch, especially before it's done.

lemielek: Are you doing optimalisation of codec on OpenRTX version or on M17 version?
https://github.com/M17-Project/Codec2-mod

silseva: I was about to suggest to move to the M17 version of Codec2: Wojciech did some optimizations which reduced the computation time significantly

fenfir: the m17 version. the problem is that it's become it's own beast since I've had to do a lot of very low level optimization. The ck803s doesn't have floating point, I did a fixed point conversion on Wojciech's version, but then had do do a bunch of other stuff to get it to run fast enough. That's why I don't want to include it right now.

fenfir: https://github.com/fenfir/Codec2-mod/pull/1

fenfir: I also have another version of voiceprompts that don't use codec2, if you would rather I include that for now.

silseva: > the m17 version. the problem is that it's become it's own beast since I've had to do a lot of very low level optimization. The ck803s doesn't have floating point, I did a fixed point conversion on Wojciech's version, but then had do do a bunch of other stuff to get it to run fast enough. That's why I don't want to include it right now.

This is going to be useful as well for other targets which don't have an FPU. Actually it'd be interesting to merge your work with Wojciech's optimizations.

> I also have another version of voiceprompts that don't use codec2, if you would rather I include that for now.

This is interesting as well. Are you just writing the uncompressed speech into flash or are you doing something else?

silseva:
@fenfir do you mind waiting before pushing new changes until this first block gets merged, please? Otherwise reviewing becomes very difficult for me. Another possibility is to open a second PR with the rtx part.

fenfir:
I re-pushed the original 5, with a small change to GPS to bring it inline with the others, and skipping the rtx bring up for now.

platform/targets/HD2/hd2_core_stubs.c
fenfir: I'm converting the codec2 .vpc to adpcm, then an integer based adpcm decoder/player.

platform/drivers/backlight/backlight_HD2.c
```diff
@@ -0,0 +1,82 @@
...
+    PWM_CH0->RESERVED_HW[2] = PWM_CH0_GATE_ON;
+
+    bl_initialised = 1;
+    bl_last_level = 0xff;
+}
+
+void backlight_terminate(void)
+{
+    if (bl_initialised)
+        pwm_program(PWM_CH0, BACKLIGHT_PWM_HZ, 0); /* duty 0 = dark */
```
silseva: This should be enough:
```suggestion
pwm_program(PWM_CH0, BACKLIGHT_PWM_HZ, 0); /* duty 0 = dark */
```

platform/drivers/backlight/backlight_HD2.c
```diff
@@ -0,0 +1,82 @@
...
+}
+
+void display_setBacklightLevel(uint8_t level)
+{
+    if (!bl_initialised)
+        backlight_init();
```
silseva: `backlight_init` should be called explicitly by the display or platform driver: I prefer to not have an implicit initialization because it can hide bugs.

platform/drivers/backlight/backlight_HD2.c
```diff
@@ -0,0 +1,82 @@
...
+
+    if (level > 100)
+        level = 100;
+    if (level == bl_last_level) /* idempotent: the UI can call this often */
+        return;
```
silseva: Unless setting the same value many times can cause issues, I'd just allow it.

platform/drivers/display/ST7735S_HD2.c
```diff
@@ -0,0 +1,196 @@
...
+#include "interfaces/display.h"
+#include "interfaces/delays.h"
+#include "hwconfig.h"
+#include "registers.h"
+#include "peripherals/gpio.h"
+#include <stdint.h>
+#include <stddef.h>
+
+extern void delayMs(unsigned int ms);
+extern void backlight_init(void);
```
silseva: Please include the respective headers instead of forward declaring external functions.

platform/mcu/HR_C7000/registers.h
```diff
@@ -0,0 +1,248 @@
...
+ *  LCD -- HR_C7000 hardware i8080 controller (base 0x12000000, manual 5.3)
+ * ------------------------------------------------------------------------- */
+typedef struct {
+    volatile uint32_t INDEX; /* 0x00 command cycle */
+    volatile uint32_t DATA;  /* 0x04 8-bit data cycle */
+    uint32_t RESERVED0[2];   /* 0x08 - 0x0c */
+    volatile uint32_t WCFG;  /* 0x10 write strobe timing */
+} LCD_TypeDef;
+
+#define LCD_WCFG_DEFAULT 0x020202u /* 2/2/2 strobes ~143 ns */
```
silseva: Plase move this to `ST7735S_HD2.c`

platform/drivers/display/ST7735S_HD2.c
```diff
@@ -0,0 +1,196 @@
...
+    for (size_t i = 0; i < (size_t)CONFIG_SCREEN_WIDTH * CONFIG_SCREEN_HEIGHT;
+         ++i) {
+        LCD_HW->DATA = 0;
+        LCD_HW->DATA = 0;
+    }
+
+    send_cmd(ST7735_DISPON);
+
+    backlight_init();
+    display_setBacklightLevel(50);
```
silseva: This line can be omitted, backlight level is set later on: https://github.com/OpenRTX/OpenRTX/blob/ea89d8d7e723a92601bbdb0cce89c28eaa30a5fc/openrtx/src/core/openrtx.c#L60
I

platform/drivers/display/ST7735S_HD2.c
```diff
@@ -0,0 +1,196 @@
...
+
+void display_render(void *fb)
+{
+    display_renderRows(0, CONFIG_SCREEN_HEIGHT, fb);
+}
+
+void display_setContrast(uint8_t contrast)
+{
+    (void)contrast; /* ST7735S has no contrast register */
+}
```
silseva: Probably this can be omitted, given that `CONFIG_SCREEN_CONTRAST` is not defined for the HD2

platform/drivers/keyboard/keyboard_HD2.c
```diff
@@ -0,0 +1,169 @@
...
+        gpio_setMode(GPIOC, KBD_ROW_SHIFT + row, INPUT);
+    for (unsigned col = 0; col < KBD_COL_COUNT; ++col) {
+        gpio_setMode(GPIOC, KBD_COL_SHIFT + col, OUTPUT);
+        gpio_setPin(GPIOC, KBD_COL_SHIFT + col);
+    }
+
+    uint16_t out = 0;
+    for (unsigned col = 0; col < KBD_COL_COUNT; ++col) {
+        gpio_clearPin(GPIOC, KBD_COL_SHIFT + col); /* drive this col low */
+        scan_settle();
```
silseva: Please use a `delayUs`, if possible.

silseva:
> should I rebase the changes onto the 5 commits? or add a clean up commit?

The first one, you can use fixup commits: https://fle.github.io/git-tip-keep-your-branch-clean-with-fixup-and-autosquash.html

platform/mcu/HR_C7000/registers.h
```diff
@@ -0,0 +1,248 @@
...
+ * PTC pad-mux values (IO_DIPLEX2, manual 4.4.4.6 -- upper bits are write-only,
+ * so always write the whole-register constant, never read-modify-write):
+ *   PTC_GPIO: every PTC pad in GPIO mode (keypad matrix scan).
+ *   LCD_I80:  bits 3..18 = 0 -> the HW i8080 controller @0x12000000 owns
+ *             PTC3-6 (CS/RS/WR/RD) and PTC7-14 (DB0-7); the rest stays GPIO
+ *             (PTC2 = panel reset, ADC/DAC pads).
+ * The keypad matrix shares the LCD data pins (rows PTC7-10, cols PTC11-14),
+ * so a scan momentarily swaps the mux GPIO<->I80 (see keyboard_HD2.c).
+ */
+#define HD2_DIPLEX2_PTC_GPIO 0x3ffffffbu
```
silseva: These are HD2-specific parts: I'd move them to a `pinmap.h` file under `targets/HD2`.

platform/drivers/ADC/adcHrc7000.c
silseva, line 1: Please implement the ADC driver following the standard driver interface: https://github.com/OpenRTX/OpenRTX/blob/master/openrtx/include/peripherals/adc.h

You can take as an example what was done for the STM32F4:
- https://github.com/OpenRTX/OpenRTX/blob/master/platform/drivers/ADC/adc_stm32f4.c
- https://github.com/OpenRTX/OpenRTX/blob/ea89d8d7e723a92601bbdb0cce89c28eaa30a5fc/platform/targets/MD-3x0/platform.c#L82

There is no need for `adc_hd2_battery_raw_avg` because battery voltage is already filtered by the state update function.

The driver name should be `adcHrc7000` or something similar: the peripheral is actually part of the HR_C7000 chip and the same driver can be used in other targets as well.

platform/targets/HD2/hd2_glue.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+void sleepUntil(long long timestampMs)
+{
+    miosix::Thread::nanoSleepUntil(timestampMs * 1000000LL);
+}
+
+long long getTick(void)
+{
+    return miosix::getTime() / 1000000LL; // ns -> ms (TICK_FREQ=1000)
+}
```
silseva: These functions should actually go in a dedicated `delays.cpp` file in `mcu/HR_C7000/drivers`.
See https://github.com/OpenRTX/OpenRTX/blob/master/platform/mcu/STM32F4xx/drivers/delays.cpp

silseva:
A couple of general comments:
- like the other targets, it'd be good to have a `pinmap.h` file where each GPIO is assigned a unique label specifying its function. Instead of having `gpio_readPin(GPIOB, 11)`, it is better to have `gpio_readPin(PTT_IN)`. See for example https://github.com/OpenRTX/OpenRTX/blob/master/platform/targets/MD-3x0/pinmap.h
- please remove all the implicit initialization paths like
```
    if (!rtc_ready)
        rtc_init()
 ```
A peripheral should be initialized only explicitly by calling its init function: if this does not happen, is a bug and it should cause side effects.

File: platform/targets/HD2/hd2_core_stubs.c
Comment (fenfir) [reply platform/targets/HD2/hd2_core_stubs.c:1]: In relation to the other comment about not pushing changes that expand the scope, can I leave these stubs for now? I have further prs that clear them, the rtx bring up and the adpcm bring up.

Or should I stub them in a common area?

File: platform/drivers/keyboard/keyboard_HD2.c (hunk 1 above)
Comment (fenfir) [reply platform/drivers/keyboard/keyboard_HD2.c]: Done. Added `platform/drivers/GPIO/gpio_hrc7000.{c,h}` implementing the standard `peripherals/gpio.h` interface (`setMode`/`setPin`/`clearPin`/`readPin`); the keyboard now uses it. The one HR_C7000-specific piece — the vendor "alternate-function B" selector (per-bank register at `+0x78`, outside the DW_apb_gpio map and separate from the SOCSYS pad-mux) — is exposed as `gpio_setAltFuncB()` in the driver rather than a raw macro.

File: platform/drivers/backlight/backlight_HD2.c (hunk 2 above)
Comment (fenfir) [reply platform/drivers/backlight/backlight_HD2.c]: Done.

File: platform/drivers/backlight/backlight_HD2.c (hunk 3 above)
Comment (fenfir) [reply platform/drivers/backlight/backlight_HD2.c]: Done — `backlight_init()` is now called explicitly from `display_init` (matching `ST7735R_CS7000`), no init-on-first-use. Also removed a stray double-init: it had been called from both `platform_init` and `display_init`.

File: platform/drivers/backlight/backlight_HD2.c (hunk 4 above)
Comment (fenfir) [reply platform/drivers/backlight/backlight_HD2.c]: Done — dropped the idempotent guard; setting the same level is allowed.

File: platform/drivers/display/ST7735S_HD2.c (hunk 5 above)
Comment (fenfir) [reply platform/drivers/display/ST7735S_HD2.c]: Done — includes `drivers/backlight/backlight.h` and `pinmap.h`; the extern decls are removed.

File: platform/mcu/HR_C7000/registers.h (hunk 6 above)
Comment (fenfir) [reply platform/mcu/HR_C7000/registers.h]: Done — `LCD_WCFG_DEFAULT` / `LCD_RESET_BIT` moved into `ST7735S_HD2.c`.

File: platform/drivers/display/ST7735S_HD2.c (hunk 7 above)
Comment (fenfir) [reply platform/drivers/display/ST7735S_HD2.c]: Done — removed; the core sets the level in `openrtx.c`.

File: platform/drivers/display/ST7735S_HD2.c (hunk 8 above)
Comment (fenfir) [reply platform/drivers/display/ST7735S_HD2.c]: Done — removed (`CONFIG_SCREEN_CONTRAST` isn't defined for HD2).

File: platform/drivers/keyboard/keyboard_HD2.c (hunk 9 above)
Comment (fenfir) [reply platform/drivers/keyboard/keyboard_HD2.c]: Done — the scan settle now uses `delayUs(10)`.

File: platform/mcu/HR_C7000/registers.h (hunk 10 above)
Comment (fenfir) [reply platform/mcu/HR_C7000/registers.h]: Done — board pins/pad-mux moved to `platform/targets/HD2/pinmap.h`.

File: platform/drivers/ADC/adcHrc7000.c
Comment (fenfir) [reply platform/drivers/ADC/adcHrc7000.c:1]: Done — reimplemented as `adcHrc7000` following `peripherals/adc.h` (per the STM32F4 example), dropped the battery averaging (the state update filters it), and renamed it since the ADC is an HR_C7000 peripheral reusable by other targets. While there I also fixed a first-conversion START→BUSY race that made the *initial* battery read 0 (spurious low-battery panic at boot).

File: platform/targets/HD2/hd2_glue.cpp (hunk 11 above)
Comment (fenfir) [reply platform/targets/HD2/hd2_glue.cpp]: Done — moved to `platform/mcu/HR_C7000/drivers/delays.cpp`.

File: platform/drivers/ADC/adcHrc7000.c [hunk 12]
```diff
@@ -0,0 +1,109 @@
...
+#include "registers.h"
+#include <pthread.h>
+
+#define ADC_STATE_BUSY 0x01u    /* ADC_CTRL_BUSY              */
+#define ADC_STATE_FSM 0x3eu     /* SAMP_FSM_STATE (bits[5:1]) */
+#define ADC_SAMPLE_MASK 0x3ffu  /* 10-bit result             */
+#define ADC_GUARD_ITERS 200000u /* bounded busy-wait guard   */
+
+/* Wait for BUSY + SAMP_FSM clear.  Returns 1 on idle, 0 on timeout. */
+static int adc_wait_idle(void)
```
silseva: This function is almost equal to `adc_wait_busy`: probably it does make sense to have a single function which does both things based on a parameter.

platform/drivers/ADC/adcHrc7000.c
```diff
@@ -0,0 +1,109 @@
...
+
+void adcHrc7000_terminate(const struct Adc *adc)
+{
+    if (adc->mutex != NULL)
+        pthread_mutex_destroy((pthread_mutex_t *)adc->mutex);
+}
+
+uint16_t adcHrc7000_sample(const struct Adc *adc, const uint32_t channel)
+{
+    (void)adc; /* single ADC block on the HR_C7000 */
```
silseva, line 67: You need to acquire the mutex before starting a new conversion and release it afterwards.
```
    if(adc->mutex != NULL)
        pthread_mutex_lock((pthread_mutex_t *) adc->mutex);
```

platform/drivers/ADC/adcHrc7000.h
```diff
@@ -0,0 +1,56 @@
...
+ * @param mutx: pointer to mutex for concurrent access, can be NULL.
+ * @param res:  conversion factor from counts to uV (see ADC_COUNTS_TO_UV).
+ */
+#define ADC_HRC7000_DEVICE_DEFINE(name, mutx, res)             \
+    extern uint16_t adcHrc7000_sample(const struct Adc *adc,   \
+                                      const uint32_t channel); \
+    const struct Adc name = { .sample = &adcHrc7000_sample,    \
+                              .priv = NULL,                    \
+                              .mutex = mutx,                   \
+                              .countsTouV = res };
```
silseva:
```suggestion
                              .countsTouV = res, };
```
Adding a comma after `res`, otherwise clang-format misbehaves

platform/drivers/display/ST7735S_HD2.c
```diff
@@ -0,0 +1,195 @@
...
+#include "peripherals/gpio.h"
+#include <stdint.h>
+#include <stddef.h>
+#include "drivers/backlight/backlight.h"
+
+/* Display/panel constants (moved out of the SoC registers.h -- these belong to
+ * the ST7735S driver, not the HR_C7000 register map): the i8080 write-strobe
+ * timing and the panel RESET pin (PTC2, GPIO, active-low). */
+#define LCD_WCFG_DEFAULT 0x020202u /* 2/2/2 strobes ~143 ns */
+#define LCD_RESET_BIT (1u << 2)    /* panel reset on PTC2 (GPIO, active-low) */
```
silseva: This is not used anywhere, it can be removed.

platform/drivers/GPIO/gpio_hrc7000.h
```diff
@@ -0,0 +1,92 @@
...
+ * register at +0x78).  This is distinct from the SOCSYS pad-mux and is sticky
+ * across gpio_setMode() direction changes; the keypad-matrix rows use it to
+ * sense while the same pins otherwise carry LCD-bus signals.
+ *
+ * @param port: gpio port (GPIOA/GPIOB/GPIOC).
+ * @param pin: gpio pin number, between 0 and 31.
+ */
+static inline void gpio_setAltFuncB(const void *port, const uint8_t pin)
+{
+    ((GPIO_TypeDef *)port)->ALT_FUNC_B |= (1u << pin);
```
silseva, line 85: Perhaps this can become the setting for `ALTERNATE` mode in `gpio_setMode`? Just an idea

platform/mcu/HR_C7000/drivers/rtc_hd2.c
```diff
@@ -0,0 +1,260 @@
...
+#define RTC_VAL_H 0x02u
+#define RTC_VAL_D_L 0x03u
+#define RTC_VAL_D_H 0x04u
+#define RTC_LOAD_S 0x0cu
+#define RTC_LOAD_M 0x0du
+#define RTC_LOAD_H 0x0eu
+#define RTC_LOAD_D_L 0x0fu
+#define RTC_LOAD_D_H 0x10u
+#define RTC_CCR 0x11u
+#define RTC_LOAD_VSTAT 0x12u
```
silseva: It'd be nice to have an `enum`, like in the display driver for the command set.

platform/mcu/HR_C7000/drivers/rtc_hd2.c
```diff
@@ -0,0 +1,260 @@
...
+    } while (0)
+
+/*
+ * Enable the controller ONCE and leave it enabled (target set while disabled,
+ * then wait for IC_ENABLE_STATUS to settle), mirroring the proven i2c_csky
+ * (I2C1) sequence on the same DesignWare core.  Toggling IC_ENABLE around every
+ * transfer -- and not waiting for it to settle -- corrupts the FIFO/command
+ * alignment and returns garbage (the original RTC round-trip failure).
+ */
+static void i2c2_setup(void)
```
silseva: The HR_C7000 has more than one I2C peripheral: if they are all equal is better to have a unique I2C driver, create an instance for the I2C1 and use this instance in the RTC driver.
The I2C driver follows the usual convention, with the API being listed here: https://github.com/OpenRTX/OpenRTX/blob/master/openrtx/include/peripherals/i2c.h It will be the first I2C driver, if you think there is something which needs to be changed in the API let me know: having an external view is always good.

platform/mcu/HR_C7000/drivers/rtc_hd2.c
```diff
@@ -0,0 +1,260 @@
...
+
+    I2C2_WAIT_IDLE();
+    I2C2->IC_START = 0u;
+    i2c2_recover();
+    return val;
+}
+
+/* ---- civil-date <-> days-since-1970 (Howard Hinnant, branch-free) --- */
+
+static int32_t civil_to_days(int32_t y, uint32_t m, uint32_t d)
```
silseva: This function can go to `datetime.h/.c`: in this way it becomes a library function and it can be reused in the future.
https://github.com/OpenRTX/OpenRTX/blob/master/openrtx/include/core/datetime.h

platform/mcu/HR_C7000/drivers/rtc_hd2.c
```diff
@@ -0,0 +1,260 @@
...
+{
+    y -= (m <= 2);
+    int32_t era = (y >= 0 ? y : y - 399) / 400;
+    uint32_t yoe = (uint32_t)(y - era * 400);
+    uint32_t doy = (153u * (m + (m > 2 ? -3u : 9u)) + 2u) / 5u + d - 1u;
+    uint32_t doe = yoe * 365u + yoe / 4u - yoe / 100u + doy;
+    return era * 146097 + (int32_t)doe - 719468;
+}
+
+static void days_to_civil(int32_t z, int32_t *y, uint32_t *m, uint32_t *d)
```
silseva: Same as `civil_to_days`

platform/targets/HD2/CMakeLists.txt
```diff
@@ -0,0 +1,115 @@
...
+# radio / DMR / audio / M17 / NVM -- those are stubbed (hd2_stubs.c) so the
+# core + UI link.
+cmake_minimum_required(VERSION 3.21)
+
+# Kernel dependency.  Normally miosix-kernel is a sibling of OpenRTX under
+# vendor/ (../../../../miosix-kernel/miosix).  This worktree lives outside that
+# tree, so MIOSIX_KPATH defaults to the known checkout and is overridable:
+#   cmake -DMIOSIX_KPATH=/path/to/miosix-kernel/miosix ...
+set(MIOSIX_KPATH "/Users/zach/Documents/fenfir/hd1/hd2-clean/vendor/miosix-kernel/miosix"
+    CACHE PATH "Path to the miosix kernel tree")
```
silseva: Probably is better to require the user to always invoke cmake with `-DMIOSIX_KPATH=/path/to/miosix-kernel/miosix` instead of having an implicit default that will fail everywhere except on one computer.

platform/targets/HD2/hd2_core_stubs.c
```diff
@@ -0,0 +1,98 @@
...
+{
+    (void)string;
+    (void)flags;
+}
+void vp_queueStringTableEntry(const char *const *e)
+{
+    (void)e;
+}
+
+/* ---- rtx (no radio task in the minimal bring-up) -------------------------- */
```
silseva, line 71: By default the rtx layer is initialized to no operating mode and just keeps the thread yielded. Instead of stubbing it, you can just include the stub radio driver and force the opmode to FM or NONE

platform/targets/HD2/main.cpp
silseva, line 1: There is already a `main()` which is exactly like this: https://github.com/OpenRTX/OpenRTX/blob/master/openrtx/src/main.c

Is there a specific reason to have this one?

platform/targets/HD2/platform.c
```diff
@@ -0,0 +1,163 @@
...
+#include "drivers/ADC/adcHrc7000.h"
+#include "drivers/GPS/gps_HD2.h"
+#include "drivers/backlight/backlight.h"
+#include "hwconfig.h"
+#include "pinmap.h"
+
+/* On-chip ADC.  Battery pack on channel 2, read through a 1:3 divider; the
+ * cell reference is 3.3 V and the converter is 10-bit.  No mutex -- the state
+ * task is the only sampler. */
+#define ADC_VBAT_CH 2u
```
silseva: Conventionally these definitions go to `hwconfig.h`, using `enum`s.

platform/targets/HD2/platform.c
```diff
@@ -0,0 +1,163 @@
...
+ADC_HRC7000_DEVICE_DEFINE(adc1, NULL, ADC_COUNTS_TO_UV(3300000, 10))
```
silseva: Do you mind moving this to `hwconfig.c`? I know that in other targets these declarations are scattered around, but the plan is to move everything to `hwconfig.h/.c`.
You can take the CS7000 as a reference: https://github.com/OpenRTX/OpenRTX/tree/master/platform/targets/CS7000

platform/targets/HD2/platform.c
```diff
@@ -0,0 +1,163 @@
...
+    /* Establish the V2.1.3 SoC pin-mux + GPIO baseline the display and
+     * backlight depend on.  The Miosix kernel already brought up the PLL,
+     * timers and UART0; here we only own the pad mux / GPIO directions that
+     * the vendor firmware sets and the LCD/PWM hardware needs.  (Clocks are
+     * NOT re-touched under Miosix -- re-running the PLL init hangs the kernel
+     * timer.)  These are whole-register vendor snapshot values; the upper
+     * DIPLEX bits are write-only, so never read-modify-write them. */
+    SOCSYS->IO_DIPLEX0 = 0x00000060u; /* PTA: vendor pad-mux baseline */
+    SOCSYS->IO_DIPLEX1 = 0x00000007u; /* PTA: SPI0 pads (vendor baseline) */
+    SOCSYS->IO_DIPLEX2 = HD2_DIPLEX2_LCD_I80; /* PTC: LCD i8080 owns the bus */
```
silseva: This should go to the display driver, so that everything is in one place

platform/targets/HD2/platform.c
```diff
@@ -0,0 +1,163 @@
...
+
+    GPIOA->DDR = 0x003401e0u;
+    GPIOA->DR |= 0x001401e0u;
+
+    GPIOB->DDR = 0x3d7ae51fu; /* incl. knob/PTT sense as input */
+    GPIOB->DR |= 0x00506414u; /* preserves the PTB13 power latch */
+
+    GPIOC->DDR = 0x00007ffcu; /* PTC2..14 = LCD-bus outputs */
```
silseva: Same as above :)

platform/targets/HD2/platform.c
```diff
@@ -0,0 +1,163 @@
...
+     * the vendor firmware sets and the LCD/PWM hardware needs.  (Clocks are
+     * NOT re-touched under Miosix -- re-running the PLL init hangs the kernel
+     * timer.)  These are whole-register vendor snapshot values; the upper
+     * DIPLEX bits are write-only, so never read-modify-write them. */
+    SOCSYS->IO_DIPLEX0 = 0x00000060u; /* PTA: vendor pad-mux baseline */
+    SOCSYS->IO_DIPLEX1 = 0x00000007u; /* PTA: SPI0 pads (vendor baseline) */
+    SOCSYS->IO_DIPLEX2 = HD2_DIPLEX2_LCD_I80; /* PTC: LCD i8080 owns the bus */
+
+    GPIOA->DDR = 0x003401e0u;
+    GPIOA->DR |= 0x001401e0u;
```
silseva, line 42: Please add a comment to document these two lines.

platform/targets/HD2/hd2_core_stubs.c
silseva: If you can fix the rtx one as I suggested, it'd be nice. If it takes more than five minutes, ignore it and we keep these stubs as they are now and remove them in later PRs.

silseva:
Note for future PRs: if possible, make each commit bisectable so that we can go back to them in case something does not work.

fenfir:
what do you mean by bisectable? I've generally been working forward a step at a time after this set of changes, and avoiding making too many changes at one time. This particular case has been a bit hard since I pushed the FW way beyond here, but have scaled it back to make all if this more reviewable by re-deriving what I had known working to create a reasonable reviewable set.

I'm testing on device for this set every time before moving forward. My thought after this is to write an on device test suite that can be interactively worked through to test all of the features of the FW on device.

platform/drivers/ADC/adcHrc7000.c
fenfir: Done — merged `adc_wait_idle`/`adc_wait_busy` into a single `static int adc_wait(bool waitBusy)` and updated both call sites in `adcHrc7000_sample`.

platform/drivers/ADC/adcHrc7000.c
fenfir: The generic `adc_getRawSample()` wrapper in `peripherals/adc.h` already locks `adc->mutex` around the driver's `sample()`, so taking the lock inside `sample()` as well would double-lock it (the STM32F4 driver's `sample()` likewise doesn't lock for that reason).

The real bug this points at is that `adc1` was defined with a `NULL` mutex, so nothing was actually being serialized. Fixed: `adc1` now gets a real `pthread_mutex_t` (`adc1Mutex` in `hwconfig.c`, matching the CS7000 pattern), so the wrapper serializes concurrent conversions. Happy to move the lock into the driver instead if you'd prefer it there.

platform/drivers/ADC/adcHrc7000.h
fenfir: Done.

platform/drivers/display/ST7735S_HD2.c
fenfir: Done — removed.

platform/drivers/GPIO/gpio_hrc7000.h
fenfir: Kept as a separate hook for now. On this SoC the standard alternate-function/pad-mux selection goes through the SOCSYS `IO_DIPLEX*` registers, whereas `gpio_setMode(..., ALTERNATE)` maps to the DW_apb_gpio block's own mode bits. `gpio_setAltFuncB` pokes a bespoke per-bank vendor register at `+0x78` that's distinct from both, so folding it into `setMode(ALTERNATE)` would conflate two different mechanisms. Happy to revisit unifying them if you'd prefer.

platform/targets/HD2/hd2_core_stubs.c
fenfir: I looked into wiring up the real rtx path (dropping the `rtx_*` stubs and adding `rtx.cpp` + `radio_stub.c`), but `rtx.cpp` instantiates `OpMode_FM` unconditionally and `OpMode_FM.cpp` calls `AT1846S::instance()` directly, so linking it pulls in the full AT1846S baseband + I2C1 HAL — exactly the scope this PR is trying to keep out.

So I've kept the `rtx_*` (and `vp_*`) stubs in `hd2_core_stubs.c` for this PR. The FM branch that follows brings up the AT1846S HAL + OpMode_FM properly and clears the `rtx_*` stubs; the ADPCM/codec2 voice-prompt branches clear the `vp_*` ones.

One upstream thought while I was in there: `fmMode` is instantiated unconditionally in `rtx.cpp` (only `m17Mode` is config-gated). If `fmMode` were gated the same way, a radio could link the rtx stack without a baseband driver. Happy to open that as a separate change if it'd be useful.

sp5wwp:
@fenfir I dropped a comment there as well (a possible improvement in the clz function).

platform/targets/HD2/hd2_core_stubs.c
silseva: > I looked into wiring up the real rtx path (dropping the `rtx_*` stubs and adding `rtx.cpp` + `radio_stub.c`), but `rtx.cpp` instantiates `OpMode_FM` unconditionally and `OpMode_FM.cpp` calls `AT1846S::instance()` directly, so linking it pulls in the full AT1846S baseband + I2C1 HAL — exactly the scope this PR is trying to keep out.

Ouch! I'll need to remove this dependency, then.

> So I've kept the `rtx_*` (and `vp_*`) stubs in `hd2_core_stubs.c` for this PR. The FM branch that follows brings up the AT1846S HAL + OpMode_FM properly and clears the `rtx_*` stubs; the ADPCM/codec2 voice-prompt branches clear the `vp_*` ones.

Looks good to me!

> One upstream thought while I was in there: `fmMode` is instantiated unconditionally in `rtx.cpp` (only `m17Mode` is config-gated). If `fmMode` were gated the same way, a radio could link the rtx stack without a baseband driver. Happy to open that as a separate change if it'd be useful.

Indeed, FM mode should be gated as well. There are targets like Module17 which do not support FM and it does not make sense to have it enabled there.

platform/drivers/ADC/adcHrc7000.c
silseva: Thank you!

platform/drivers/ADC/adcHrc7000.c
silseva: > The generic `adc_getRawSample()` wrapper in `peripherals/adc.h` already locks `adc->mutex` around the driver's `sample()`, so taking the lock inside `sample()` as well would double-lock it (the STM32F4 driver's `sample()` likewise doesn't lock for that reason).

You're right, I completely forgot about that.

> The real bug this points at is that `adc1` was defined with a `NULL` mutex, so nothing was actually being serialized. Fixed: `adc1` now gets a real `pthread_mutex_t` (`adc1Mutex` in `hwconfig.c`, matching the CS7000 pattern), so the wrapper serializes concurrent conversions. Happy to move the lock into the driver instead if you'd prefer it there.

Mutex locking should stay in `adc_getRawSample`, so that is implemented once for all the drivers. I wrote that part and then I completely forgot about it.

platform/drivers/GPIO/gpio_hrc7000.h
silseva: Fine, let's leave it as it is for now. Thank you for the clarification! :)

silseva:
> what do you mean by bisectable? I've generally been working forward a step at a time after this set of changes, and avoiding making too many changes at one time. This particular case has been a bit hard since I pushed the FW way beyond here, but have scaled it back to make all if this more reviewable by re-deriving what I had known working to create a reasonable reviewable set.
> 
> I'm testing on device for this set every time before moving forward. My thought after this is to write an on device test suite that can be interactively worked through to test all of the features of the FW on device.

Is all good, my comment was referred to this line in your PR description:

> Note: the commits are grouped for reviewability, not bisectability — CMakeLists.txt lives in the platform commit, so only the branch tip builds standalone.

I know that having each commit building successfully is really difficult, if not impossible, especially on the early bringup of a device. Again, all good and thanks for your work!

platform/targets/HD2/platform.c
fenfir: Done — `ADC_VBAT_CH`, the `adc1` device instance (+ its mutex), and the `hwInfo` struct all moved to `hwconfig.{h,c}`.

The one residual is the `extern const hwInfo_t hwInfo;` you see here: `hwconfig.h` can't declare it, because `hwInfo_t` comes from `interfaces/platform.h`, and `platform.h` already includes `hwconfig.h` — so pulling `platform.h` into `hwconfig.h` creates an include cycle (which bit me: the `CONFIG_*`-gated decls in `platform.h` get skipped). CS7000 sidesteps the same issue by keeping its `hwInfo` definition in `platform.c`; I kept the definition in `hwconfig.c` per your "everything in hwconfig" plan and left only this forward `extern` in `platform.c`. Happy to move it back to a `static` in `platform.c` (CS7000-style) if you'd prefer no extern at all.

platform/targets/HD2/platform.c
fenfir: Done — added a comment naming the PTA pins: these are the V2.1.3 vendor boot-snapshot DDR/DR values for PTA (rotary encoder on PTA0/1 as inputs; the rest kept verbatim as the known-good pad state).

platform/targets/HD2/platform.c
fenfir: Done — moved into `display_init` (ST7735S_HD2.c), alongside the `GPIOC->DDR` LCD-bus directions, so the LCD i8080 pad-mux lives entirely in the display driver.

platform/targets/HD2/platform.c
fenfir: Done — the `GPIOC->DDR` LCD-bus directions moved into `display_init` too.

platform/targets/HD2/hd2_core_stubs.c
fenfir: Great — I'll open a separate PR gating `fmMode` in `rtx.cpp` behind a config the same way `m17Mode` is, so Module17-class targets (and this minimal HD2 build) can link the rtx stack without a baseband/FM driver.

platform/mcu/HR_C7000/drivers/i2c_hrc7000.c
```diff
@@ -0,0 +1,164 @@
...
+}
+
+/* Base controller config (master, 7-bit, fast-mode, restart-enabled).  IC_CON /
+ * timing can only change while disabled, so leave the controller OFF here; the
+ * first transfer sets IC_TAR and enables it (ensure_target). */
+static void configure(I2C_TypeDef *hw)
+{
+    hw->IC_ENABLE = 0u;
+    for (int spin = I2C_SPIN_LIMIT; spin && (hw->IC_ENABLE_STATUS & 1u);
+         --spin) {
```
silseva: Can this be changed to
```suggestion
    wait_status(hw, IC_ENABLE_STATUS, true)
```
?

platform/mcu/HR_C7000/drivers/i2c_hrc7000.c
```diff
@@ -0,0 +1,164 @@
...
+ * A single-slave bus (e.g. the RTC) thus configures the target exactly once. */
+static void ensure_target(I2C_TypeDef *hw, uint8_t addr)
+{
+    if ((hw->IC_ENABLE_STATUS & 1u) && (hw->IC_TAR == addr))
+        return;
+
+    hw->IC_ENABLE = 0u;
+    for (int spin = I2C_SPIN_LIMIT; spin && (hw->IC_ENABLE_STATUS & 1u);
+         --spin) {
+    }
```
silseva: Also here

platform/mcu/HR_C7000/drivers/i2c_hrc7000.c
```diff
@@ -0,0 +1,164 @@
...
+    hw->IC_TAR = addr;
+    hw->IC_ENABLE = 1u;
+    for (int spin = I2C_SPIN_LIMIT; spin && !(hw->IC_ENABLE_STATUS & 1u);
+         --spin) {
+    }
```
silseva: And here

platform/mcu/HR_C7000/drivers/i2c_hrc7000.c
```diff
@@ -0,0 +1,164 @@
...
+
+    return 0;
+}
+
+const struct i2cApi i2cHrc7000_driver = {
+    .read = i2c_hrc7000_read,
+    .write = i2c_hrc7000_write,
+};
+
+int i2c_init(const struct i2cDevice *dev, const uint8_t speed)
```
silseva, line 146: I'd add `phread_mutex_init(dev->mutex)` here, as in the ADC driver.

platform/mcu/HR_C7000/drivers/i2c_hrc7000.c
fenfir: Done — factored all three `IC_ENABLE_STATUS` spin-loops into the bounded-wait helper. Since `IC_ENABLE_STATUS` is a separate register from `IC_STATUS`, I generalized the helper to take the register pointer: `wait_bits(volatile uint32_t *reg, uint32_t mask, bool set)`. So this is now `wait_bits(&hw->IC_ENABLE_STATUS, 1u, false)`, and the `IC_STATUS` FIFO/bus waits use the same helper (`wait_bits(&hw->IC_STATUS, ...)`).

platform/mcu/HR_C7000/drivers/i2c_hrc7000.c
fenfir: Done — same helper here (`wait_bits(&hw->IC_ENABLE_STATUS, 1u, false)` for the disable wait, `wait_bits(&hw->IC_ENABLE_STATUS, 1u, true)` for the enable wait).

platform/mcu/HR_C7000/drivers/i2c_hrc7000.c
fenfir: Done — same helper here (`wait_bits(&hw->IC_ENABLE_STATUS, 1u, false)` for the disable wait, `wait_bits(&hw->IC_ENABLE_STATUS, 1u, true)` for the enable wait).

platform/mcu/HR_C7000/drivers/i2c_hrc7000.c
fenfir: Done — `i2c_init` now calls `pthread_mutex_init((pthread_mutex_t *)dev->mutex, NULL)` when `dev->mutex != NULL`, matching `adcHrc7000_init`.

fenfir:
The code format check failed, but it looks like a transient docker issue. I can't re-trigger the job without higher privileges I think, but it's clean locally.

Also the bisectablity in the description was a claude artifact from this PR since I was consolidating commits from re-implimenting a bunch of previous work into the 5 core, reviewable, commits here.

silseva (approved):

fenfir:
Thanks for the review and the approve. Once this is merged, I'll prepare the rtx FM bringup PR

silseva:
Thanks to you for the work!
I'm merging it in a dedicated `Ailunce-HD2` branch so that we can mainline all the work in a single block, as I did for the CS7000 radio. We also need to update all the other targets to miosix 3.0, and I'll probably couple that work with this porting and update the CMake consequently.

PR #500, opened 2026-07-27
HD2: analog-FM radio bring-up (AT1846S + rtx/OpMode_FM)

fenfir:
Builds on the merged minimal bring-up (`Ailunce-HD2`) to add the **analog-FM radio** for the Ailunce HD2 (HR_C7000 / CK803S): the AT1846S transceiver, FM RX/TX, and the `rtx`/`OpMode_FM` wiring, plus a clean audio/modem HAL split.

**On-air verified** — FM RX (squelch + audio) and FM voice TX confirmed on hardware.

### Commits (layered, each a complete step)
1. **mcu: HR_C7000 FM radio/audio/modem registers + atomic GPIO** — SOCSYS register additions (offsets guarded with `static_assert(offsetof)`), and nesting-safe atomic `gpio_setPin/clearPin` (GPIOB is now shared between the UI and radio threads).
2. **HD2: AT1846S transceiver driver** — rides the shared `i2c_hrc7000` driver via an `i2c1` device instance in `hwconfig.c` (no bespoke bus driver; conforms to `peripherals/i2c.h`, mutex in the wrapper).
3. **HD2: analog-FM radio HAL + audio routing** — `radio_HD2` owns the modem datapath; `audio_HD2` owns the codec, brought up idiomatically from `audio_init()` at `platform_init` (as in GDx/MDx/CS7000) and driving the RX AF mute through `radio_enableAfOutput/disableAfOutput`.
4. **HD2: wire rtx task + OpMode_FM, drop rtx stubs** — links `rtx.cpp` + `OpMode_FM.cpp`; removes the `rtx_*` stubs.
5. **rtx: OpMode_FM hardware RF-squelch hook** — see discussion below.

### Notes / design
- **Codec vs modem separation:** the HR_C7000 codec (`0x1600_09xx` block + `LINEOUT`/`AUDIO_CONTROL` gate) is owned by `audio_HD2`; the baseband RF datapath (`BB_DAC_CTRL`/`BB_ADC_CTRL` — these are the *baseband modulation/demod* converters per datasheet §8, **not** the codec — plus `WORK_MODE`, AT1846S) stays in `radio_HD2`. The shared `SYS_SOFT_RSTN` is handed off via an RMW in the modem boot that preserves the codec/audio-if reset bits `audio_init` owns. Register names were corrected to match the datasheet.
- **FM-only:** DMR/Layer-2 register writes were dropped (the Layer-2 struct members are kept in `registers.h` for map completeness, marked unused).

### For discussion
- **`radio_checkRxRfSquelch` hook (commit 5):** OpMode_FM normally thresholds RSSI for RF squelch; this adds an optional hook so a target with a hardware carrier comparator (the AT1846S's, steadier than RSSI) can report squelch directly. I implemented it as a **weak default** in `OpMode_FM.cpp` to avoid touching every target, but I know there's no weak-symbol precedent in the radio HAL — happy to convert it to the full contract (`return false` in each `radio_*.cpp` + `radio_stub.c`) if you'd prefer. Your call on the interface convention.
- The codec-DAC voice-prompt path (PTB17 audio-path select) lands with the voice-prompt follow-up, not this FM-only PR.

### DIFF
--- a/openrtx/src/rtx/OpMode_FM.cpp
+++ b/openrtx/src/rtx/OpMode_FM.cpp
@@ -14,6 +14,18 @@
 #include "drivers/baseband/AT1846S.h"
 #endif
 
+/*
+ * Weak default for the hardware RF-squelch hook (radio.h).  Targets whose
+ * transceiver exposes an on-chip carrier/squelch comparator override this with
+ * a strong definition to report that decision; the default reports "no hardware
+ * squelch", so update() falls back to thresholding radio_getRssi().
+ */
+extern "C" __attribute__((weak)) bool radio_checkRxRfSquelch(bool *open)
+{
+    (void) open;
+    return false;
+}
+
 /**
  * \internal
  * On MD-UV3x0 radios the volume knob does not regulate the amplitude of the
@@ -78,15 +90,25 @@ void OpMode_FM::update(rtxStatus_t *const status, const bool newCfg)
     // RX logic
     if(status->opStatus == RX)
     {
-        // RF squelch mechanism
-        // This turns squelch (0 to 15) into RSSI (-127.0dbm to -61dbm)
-        rssi_t squelch = -127 + (status->sqlLevel * 66) / 15;
-        rssi_t rssi    = rtx_getRssi();
+        // RF squelch mechanism.  Prefer the target's hardware carrier/squelch
+        // comparator if it has one (steadier than thresholding the RSSI);
+        // otherwise turn the squelch level (0 to 15) into an RSSI threshold.
+        bool hwSqlOpen;
+        if(radio_checkRxRfSquelch(&hwSqlOpen))
+        {
+            rfSqlOpen = hwSqlOpen;
+        }
+        else
+        {
+            // This turns squelch (0 to 15) into RSSI (-127.0dbm to -61dbm)
+            rssi_t squelch = -127 + (status->sqlLevel * 66) / 15;
+            rssi_t rssi    = rtx_getRssi();
 
-        // Provide a bit of hysteresis, only change state if the RSSI has
-        // moved more than 1dBm on either side of the current squelch setting.
-        if((rfSqlOpen == false) && (rssi > (squelch + 1))) rfSqlOpen = true;
-        if((rfSqlOpen == true)  && (rssi < (squelch - 1))) rfSqlOpen = false;
+            // Provide a bit of hysteresis, only change state if the RSSI has
+            // moved more than 1dBm on either side of the current squelch.
+            if((rfSqlOpen == false) && (rssi > (squelch + 1))) rfSqlOpen = true;
+            if((rfSqlOpen == true)  && (rssi < (squelch - 1))) rfSqlOpen = false;
+        }
 
         // Local flags for current RF and tone squelch status
         bool rfSql   = ((status->rxToneEn == 0) && (rfSqlOpen == true));
--- a/platform/drivers/GPIO/gpio_hrc7000.h
+++ b/platform/drivers/GPIO/gpio_hrc7000.h
@@ -37,6 +37,28 @@ extern "C" {
  */
 void gpio_setMode(const void *port, const uint8_t pin, const uint16_t mode);
 
+/*
+ * The DW_apb_gpio has no atomic bit-set/clear register, so setPin/clearPin are
+ * read-modify-write on the shared port DR.  In the threaded build a port (esp.
+ * GPIOB: LEDs + PTT + power-hold + the radio/audio pins) is poked from several
+ * threads, and a plain |=/&= can lose an update if a context switch lands
+ * between the load and the store -- worst case the lost bit is PTB13 power-hold.
+ * Single core, so masking interrupts around the RMW makes it atomic.  Uses the
+ * CK803S PSR (cr<0,0>) save + `psrclr ie` idiom (same as targets/HD2/main.c);
+ * restoring the saved PSR is nesting-safe.
+ */
+static inline uint32_t gpio_irqSave(void)
+{
+    uint32_t psr;
+    __asm__ volatile("mfcr %0, cr<0, 0>\n\tpsrclr ie" : "=r"(psr) : : "memory");
+    return psr;
+}
+
+static inline void gpio_irqRestore(uint32_t psr)
+{
+    __asm__ volatile("mtcr %0, cr<0, 0>" : : "r"(psr) : "memory");
+}
+
 /**
  * Set gpio pin to high logic level.
  *
@@ -45,7 +67,9 @@ void gpio_setMode(const void *port, const uint8_t pin, const uint16_t mode);
  */
 static inline void gpio_setPin(const void *port, const uint8_t pin)
 {
+    uint32_t psr = gpio_irqSave();
     ((GPIO_TypeDef *)port)->DR |= (1u << pin);
+    gpio_irqRestore(psr);
 }
 
 /**
@@ -56,7 +80,9 @@ static inline void gpio_setPin(const void *port, const uint8_t pin)
  */
 static inline void gpio_clearPin(const void *port, const uint8_t pin)
 {
+    uint32_t psr = gpio_irqSave();
     ((GPIO_TypeDef *)port)->DR &= ~(1u << pin);
+    gpio_irqRestore(psr);
 }
 
 /**
--- a/platform/drivers/audio/audio_HD2.c
+++ b/platform/drivers/audio/audio_HD2.c
@@ -1,0 +1,286 @@
+/*
+ * SPDX-FileCopyrightText: Copyright 2026 OpenRTX Contributors
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ *
+ * OpenRTX audio.h routing driver for the Ailunce HD2 (HR_C7000).
+ *
+ * Implements the standard OpenRTX audio path matrix (sources {MIC,RTX,MCU} ->
+ * sinks {SPK,RTX,MCU}), modeled on platform/drivers/audio/audio_GDx.c.  It is
+ * the single owner of the HD2's *board-level* audio routing: the GPIOB
+ * speaker-amp (PTB4) + gain (PTB17) + RX-audio route (PTB10) lines that
+ * radio_HD2.cpp deliberately left "to the audio agent", plus the codec /
+ * socsys audio-gate bring-up.
+ *
+ * Division of labour (one source of truth for the HW-verified register
+ * values):
+ *   - This file OWNS the HR_C7000 codec bring-up (codec_bringup(), the
+ *     byte-register CB() sequence captured LIVE from a vendor unit playing FM).
+ *     It runs at platform_init time from audio_init(), like every other OpenRTX
+ *     target, and is idempotent (a static latch makes repeat calls free).
+ *   - hd2_audio_out_warm() sets the codec audio-OUTPUT stage (LINEOUT_CTRL/
+ *     AUDIO_CONTROL/AUDIO_BUFFER_CLR) for MCU-PCM playback (beep path).
+ *   - The AT1846S RX-audio chip-side mute (reg 0x30 bit7) is released via the
+ *     radio.h hook radio_enableAfOutput()/disableAfOutput().
+ *   - This file also owns the board-level GPIO twiddles.
+ *
+ * NOTE (FM bring-up build): there is NO MCU->codec-DAC PCM playback path in
+ * this build (outputStream_HD2.cpp is not linked), so the SINK_SPK output
+ * device carries a NULL driver -- the endpoint enum is kept, but no PCM stream
+ * driver is wired.  The analog FM-RX path (SOURCE_RTX -> SINK_SPK) is pure GPIO
+ * and is fully preserved.
+ *
+ * TX (MIC->RTX) is intentionally not keyed here.
+ */
+
+#include "interfaces/audio.h"
+#include "interfaces/radio.h"
+#include "interfaces/delays.h"
+#include "drivers/GPIO/gpio_hrc7000.h"
+#include "pinmap.h" /* -> registers.h (SOCSYS, CODEC_BYTE) */
+
+/*
+ * HR_C7000 PCM/audio-codec block (byte-addressed MMIO @ 0x16000900, verified
+ * alive + readable on hardware).  Port of boot_init_modem_audio_path
+ * (0x0305e62c) + the analog-FM tail of boot_init_audio_route (0x0305e7d8).
+ * MUST use byte writes -- the block ignores partial word writes.
+ *
+ * CODEC_BYTE(off) comes from registers.h (byte-addressed PCM block); keep the
+ * short local CB() name this driver uses.
+ */
+#define CB(off) CODEC_BYTE(off)
+
+/* Codec DAC output gain (GCR_DACL @ codec 0xdf): vendor sources the low 6 bits
+ * (godl) from codeplug calibration; our FM build has no codeplug, so use the
+ * vendor's known-good value (= g_tune_data 0x34 | 0x80, from a vendor unit
+ * playing FM) so the codec DAC isn't mis-gained. */
+#define CODEC_DACL_GAIN 0xb4u
+
+static bool g_codec_inited = false;
+
+/*
+ * HR_C7000 codec bring-up (verbatim vendor sequence).  Pulses the PCM block
+ * reset then clears the codec soft-reset (SOCSYS->SYS_SOFT_RSTN bit4 once; the
+ * bit is auto-releasing, see radio_HD2.cpp's SYS_SOFT_RSTN note), waits on the
+ * LINEOUT_CTRL[31] standby handshake, then sets the DAC gain + lineout.  Runs
+ * once at platform_init time from audio_init(); idempotent.
+ *
+ * Ownership contract: audio_init() (this file) owns the codec/audio-if reset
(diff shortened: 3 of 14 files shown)


openrtx/src/rtx/OpMode_FM.cpp
```diff
@@ -14,6 +14,18 @@
 #include "drivers/baseband/AT1846S.h"
 #endif
 
+/*
```
silseva, line 17: Instead of providing a weak default implementation, I'd make it selectable via `CONFIG_RTX_RF_SQUELCH`: the targets with this feature it will have the config and provide an implementation. Is the same approach we had for the RTC

openrtx/src/rtx/OpMode_FM.cpp
```diff
@@ -78,15 +90,25 @@ void OpMode_FM::update(rtxStatus_t *const status, const bool newCfg)
...
+        else
+        {
+            // This turns squelch (0 to 15) into RSSI (-127.0dbm to -61dbm)
+            rssi_t squelch = -127 + (status->sqlLevel * 66) / 15;
+            rssi_t rssi    = rtx_getRssi();
+
+            // Provide a bit of hysteresis, only change state if the RSSI has
+            // moved more than 1dBm on either side of the current squelch.
+            if((rfSqlOpen == false) && (rssi > (squelch + 1))) rfSqlOpen = true;
+            if((rfSqlOpen == true)  && (rssi < (squelch - 1))) rfSqlOpen = false;
```
silseva, line 110: This can become a private member function. This, coupled with `CONFIG_RTX_RF_SQUELCH` can reduce the code to:
```
#ifdef CONFIG_RTX_RF_SQUELCH
bool sq = radio_checkRxRfSquelch();
#else
bool sq = chekRxSquelch();
#endif
```

platform/drivers/audio/audio_HD2.c
```diff
@@ -0,0 +1,286 @@
...
+ */
+#define CB(off) CODEC_BYTE(off)
+
+/* Codec DAC output gain (GCR_DACL @ codec 0xdf): vendor sources the low 6 bits
+ * (godl) from codeplug calibration; our FM build has no codeplug, so use the
+ * vendor's known-good value (= g_tune_data 0x34 | 0x80, from a vendor unit
+ * playing FM) so the codec DAC isn't mis-gained. */
+#define CODEC_DACL_GAIN 0xb4u
+
+static bool g_codec_inited = false;
```
silseva, line 59: The upper layers guarantee that `audio_init()` is called during the startup phase: this, in turn, calls `codec_bringup()`, so there is no real need for this variable. As with ADC, a double init is actually an error and is better if it provides a (possibly) noticeable effect.

platform/drivers/audio/audio_HD2.c
```diff
@@ -0,0 +1,286 @@
...
+             * here we only route the analog AF (PTB10) and unmute the speaker
+             * amp (PTB4). */
+            rx_route_on();
+            spkr_amp_unmute();
+            break;
+
+        case PATH(SOURCE_MCU, SINK_SPK):
+            /* Beep / voice-prompt playback from the MCU (mixes through the
+             * codec DAC -> lineout -> speaker, so the codec must be warm). */
+            hd2_audio_out_warm();
```
silseva, line 236: The code path inside this function is such that it will execute only once: what about calling it only once from `audio_init`?

platform/targets/HD2/pinmap.h
```diff
@@ -52,4 +52,19 @@
...
+#define SPKR_GAIN_PIN \
+    17 /* PTB17 speaker-amp analog PATH SELECT (NOT a plain gain):
+                            * HIGH = AT1846S analog demod (drive HIGH for analog-FM RX
+                            * -- the root-cause of the faint-FM-RX saga, LOW = ~inaudible);
+                            * LOW  = codec-DAC lineout (MCU beep / voice-prompt playback,
+                            * where HIGH would leave the codec DAC silent). */
+
+/* GPIOB analog-FM TX board pins (driven by pin number via gpio_hrc7000). */
+#define BAND_SEL_PIN 19 /* PTB19 VHF/UHF band select: HIGH >= 300 MHz */
+#define MIC_PATH_PIN 3  /* PTB3  TX mic-path gate: HIGH while keyed */
```
silseva, line 68: Please define these as `PORT,PIN` like the others, so that the macro contains the full information about the gpio

platform/targets/HD2/hwconfig.c
```diff
@@ -19,6 +19,12 @@
 static pthread_mutex_t adc1Mutex;
 ADC_HRC7000_DEVICE_DEFINE(adc1, &adc1Mutex, ADC_COUNTS_TO_UV(3300000, 10))
 
+/* Radio-bus I2C1 DesignWare master (base 0x14070000) -- the AT1846S transceiver.
+ * Pads PTA7/PTA8 are muxed to I2C1 by platform_init's IO_DIPLEX0 baseline; the
+ * mutex is taken via i2c_acquire()/i2c_release() by the AT1846S driver. */
+static pthread_mutex_t i2c1Mutex = PTHREAD_MUTEX_INITIALIZER;
```
silseva, line 25:
```suggestion
static pthread_mutex_t i2c1Mutex;
```
The I2C driver is responsible for the initialization of the mutex

platform/drivers/GPIO/gpio_hrc7000.h
```diff
@@ -37,6 +37,28 @@ extern "C" {
  */
 void gpio_setMode(const void *port, const uint8_t pin, const uint16_t mode);
 
+/*
```
silseva, line 40: I didn't see the miosix BSP yet, but these IRQ save/restore routines should already be in the operating system part. If so, here it should be enough to call the miosix API for interrupt locking/unlocking

platform/drivers/baseband/AT1846S_HD2.cpp
```diff
@@ -0,0 +1,332 @@
...
+     *
+     * The chan_struct flag selecting between them lives in the radio
+     * layer (out of scope here).  Default to the 0x4006/0x4046 path,
+     * which is what the analog FM bring-up needs.
+     *
+     * HD2 has at1846s_set_freq_alt @ 0x03058d68 (uses 0x4026/0x7026 and
+     * also reloads the audio bank); TODO when the channel layer lands.
+     */
+    i2c_writeReg16(0x30, 0x4006);
+    i2c_writeReg16(0x30, 0x4046);
```
silseva, line 283: This is what [this function](https://github.com/OpenRTX/OpenRTX/blob/33e5a4684e91d20f3ae3a0ef3cc27eb01ed35f19/platform/drivers/baseband/AT1846S.h#L381) does

platform/drivers/baseband/AT1846S_HD2.cpp
```diff
@@ -0,0 +1,332 @@
...
+    i2c_writeReg16(0x79, 0xD932); // HD2-specific VCO/PLL fine-tune
+    i2c_writeReg16(0x68, 0x05E5); // HD2-specific VCO/PLL fine-tune
+    i2c_writeReg16(0x6B, 0x02FE); // HD2-specific VCO/PLL fine-tune
+    i2c_writeReg16(0x71, 0x0F1C); // HD2-specific VCO/PLL fine-tune
+    i2c_writeReg16(0x73, 0x0A3E); // HD2-specific VCO/PLL fine-tune
+    i2c_writeReg16(0x74, 0x090E); // HD2-specific VCO/PLL fine-tune
+    i2c_writeReg16(0x75, 0x0833); // HD2-specific VCO/PLL fine-tune
+    i2c_writeReg16(0x76, 0x0806); // HD2-specific VCO/PLL fine-tune
+
+    /* --- GDx supplement (pre-cal) ----------------------------------------
```
silseva, line 153: If the HD2 original firmware does not have this "GDx supplement" and the radio is working perfectly also without it, I'd remove it. The AT1846S documentation is very small, so we do not really know whether we really need this part or not.

platform/drivers/baseband/radio_HD2.cpp
```diff
@@ -0,0 +1,558 @@
...
+#include "drivers/baseband/AT1846S.h"
+#include "hwconfig.h" /* i2c1 device instance (radio bus) */
+#include "pinmap.h"   /* BAND_SEL_PIN / MIC_PATH_PIN board pin labels */
+#include "drivers/GPIO/gpio_hrc7000.h"
+#include "radioUtils.h"
+
+// Exact-value AT1846S writes for the TX key/dekey sequence (this file, below);
+// TX was brought up and verified with these exact words -- keep them
+// bit-identical rather than routing through the AT1846S class setters.
+extern "C" void hd2_at1846s_write(uint8_t reg, uint16_t val);
```
silseva, line 43: The `extern "C"` keyword is needed only if the functions are accessed from C source files. Given that these functions are local to this translation unit, they can just be `static`

platform/drivers/baseband/radio_HD2.cpp
```diff
@@ -0,0 +1,558 @@
...
+static const rtxStatus_t *config;  // Radio configuration pointer
+static enum opstatus radioStatus;  // Current operating status
+static Band currRxBand = BND_NONE; // Current RX band
+
+/*
+ * APC TX-power level (CPU DAC channel B, 12-bit; vendor ramps this on
+ * every TX entry from a per-band power cal table we don't parse yet).
+ * Conservative default: quarter scale.
+ */
+static volatile uint16_t apcLevel = 0x400;
```
silseva, line 64: Why `volatile`? Is touched only by the rtx driver.

platform/drivers/baseband/radio_HD2.cpp
```diff
@@ -0,0 +1,558 @@
...
+
+/*
+ * Map the current codeplug TX power (mW) to the two hardware drive controls:
+ * APC DAC level (ch B) and AT1846S reg 0x0a padrv field.  Four discrete levels
+ * matching the UI (Extra Low / Low / Medium / High).  Values are conservative --
+ * even "High" sits well below the DAC ceiling (the PA runs hot near full drive).
+ */
+extern "C" void hd2_txpower_levels(uint16_t *apc, uint16_t *padrv)
+{
+    const uint32_t p = (config != nullptr) ? config->txPower : 1000u;
```
silseva, line 74: Don't access `config` implicitly from inside the function, make the input power a parameter:
```
static void hd2_txpower_levels(const uint32_t power, uint16_t *apc, uint16_t *padrv);
```

platform/drivers/baseband/radio_HD2.cpp
```diff
@@ -0,0 +1,558 @@
...
+        *apc = 0x200u;
+        *padrv = 0x0Fu;
+    } /* Medium    */
+    else {
+        *apc = 0x400u;
+        *padrv = 0x0Fu;
+    } /* High      */
+}
+
+static AT1846S &at1846s = AT1846S::instance(); // AT1846S driver (singleton)
```
silseva, line 93: Please move it near the other variables, so that there is no mix of functions and variables: this makes the code easier to read.

platform/drivers/baseband/radio_HD2.cpp
```diff
@@ -0,0 +1,558 @@
...
+    // band -> set, low band -> clear).  Steers the RF path / PA chain.
+    if (txFreq >= 300000000u)
+        gpio_setPin(GPIOB, BAND_SEL_PIN);
+    else
+        gpio_clearPin(GPIOB, BAND_SEL_PIN);
+
+    // TX power level -> APC DAC drive (ch B) + AT1846S reg 0x0a padrv.
+    uint16_t apc, padrv;
+    hd2_txpower_levels(&apc, &padrv); // map current txPower -> APC + padrv
+    apcLevel = apc; // stash into the file-scope volatile before driving the DAC
```
silseva, line 229: TX power is recomputed and applied each time a transmission starts: I think there is no need to save the APC level in a global variable.

platform/drivers/baseband/radio_HD2.cpp
```diff
@@ -0,0 +1,558 @@
...
+
+    DAC_HW->PD_MODE_EN &= ~0x2u; // ch B low-power mode off
+    DAC_HW->PD_CTRL &= ~0x2u;    // ch B power up
+    DAC_HW->DATA_B = apcLevel;
+
+    // Vendor TX parity: mic-path gate PTB3 high (vendor mic_path_set_by_freq).
+    gpio_setPin(GPIOB, MIC_PATH_PIN);
+
+    hd2_at1846s_write(0x0a, (uint16_t)((padrv << 11) | 0x0420u));
```
silseva, line 238: Perhaps you can use `at1816s.setPaDrive` here?

platform/drivers/baseband/radio_HD2.cpp
```diff
@@ -0,0 +1,558 @@
...
+
+    // TX FM deviation (reg 0x59 -- shared with the RX mixer gain, restored in
+    // hd2_fm_tx_unkey).  Pick by bandwidth and whether a sub-audio tone is
+    // active (the low 6 bits set the CTCSS/DCS deviation).
+    {
```
silseva, line 243: Perhaps you can use `at1816s.setTxDeviation` here?

platform/drivers/baseband/radio_HD2.cpp
```diff
@@ -0,0 +1,558 @@
...
+     * the mic feed into the engine is pure hardware -- no CPU sample
+     * pumping.  Do NOT touch SIG_CENTER/RF_MOD_BIAS: the boot/reset cal
+     * modulates cleanest.
+     */
+    if (config->txDisable == 1)
+        return;
+
+    // FM voice TX only (DMR / M17 out of scope for this build).
+    if (config->opMode != OPMODE_FM)
+        return;
```
silseva, line 303: This is already guarded by the fact that `CONFIG_OPMODE_M17` is not defined, thus there is no possibility to call `radio_enableTx` when M17 mode is enabled

platform/drivers/baseband/radio_HD2.cpp
```diff
@@ -0,0 +1,558 @@
...
+        radio_enableRx();
+}
+
+rssi_t radio_getRssi()
+{
+    /*
+     * AT1846S reg 0x1B upper byte holds the RSSI level; the generic driver maps
+     * it to dBm as (-137 + (reg >> 8)).
+     *
+     * Peak-hold (instant rise, slow fall ~2 dB/call @ ~33 Hz): with NO carrier
```
silseva, line 383: There is a lowpass filter in `rtx_task` whose purpose is to smooth the flickering of the RSSI: I'd improve that, if needed.

platform/drivers/baseband/radio_HD2.cpp
```diff
@@ -0,0 +1,558 @@
...
+ * setBandwidth filter-bank load, far too heavy to run on every squelch
+ * crossing.
+ */
+extern "C" void hd2_at1846s_rx_audio_config(void)
+{
+    at1846s.i2c_writeReg16(0x40,
+                           0x0030); /* RX AF-DSP LISTEN (NOT 0x11 scan-mute) */
+
+    at1846s.i2c_writeReg16(0x41, 0x471e);
+    at1846s.i2c_writeReg16(0x44, 0x0900u | (uint16_t)(g_fm_volume & 0xff));
```
silseva, line 429:
```suggestion
    at1846s.i2c_writeReg16(0x44, 0x0900u | 0x3fu); // RX audio volume
```

platform/drivers/baseband/radio_HD2.cpp
```diff
@@ -0,0 +1,558 @@
...
+    SOCSYS->SYS_INTERP_MASK =
+        SYS_INTERP_MASK_FM_VENDOR; /* modem/L2 interrupt mask (vendor live) */
+    /* No LAYER2_CONTROL/LAYER2_TXRX_CTRL: those are DMR TDMA slot-sync (datasheet
+     * §9/§10) with no analog-FM role -- verbatim vendor-snapshot cruft, and the
+     * Layer-2 engine is held in reset by SYS_SOFT_RSTN in an FM-only build. */
+}
+
+/* Live AT1846S register access, used by the TX key/dekey sequence above
+ * (hd2_fm_tx_key writes reg 0x0a/0x30/0x40/0x59 with exact verified words). */
+extern "C" uint16_t hd2_at1846s_read(uint8_t reg)
```
silseva, line 528: This is a C++ file, you can use `at1846s.i2c_readReg16` everywhere instead of creating a wrapper.

platform/drivers/baseband/radio_HD2.cpp
```diff
@@ -0,0 +1,558 @@
...
+{
+    return at1846s.i2c_readReg16(reg);
+}
+extern "C" void hd2_at1846s_write(uint8_t reg, uint16_t val)
```
silseva, line 532: Same as above

platform/drivers/baseband/radio_HD2.cpp
```diff
@@ -0,0 +1,558 @@
...
+
+/*
+ * Strong override of the OpMode_FM hardware RF-squelch hook (radio.h): the HD2
+ * has a real on-chip squelch comparator, so report sq_cmp instead of letting
+ * OpMode_FM threshold the jittery raw RSSI.
+ */
+extern "C" bool radio_checkRxRfSquelch(bool *open)
+{
+    *open = hd2_rx_carrier_detected();
+    return true;
```
silseva, line 557:
```suggestion
    return at1846s.i2c_readReg16(0x1C) & 0x0001u) != 0u;
```
Or, better, add a function to the AT1846S class (like `rxCtcssDetecte`) and call that from here.

PR #501, opened 2026-07-28
HD2: IMA-ADPCM voice prompts for the analog-FM build

fenfir:
Adds IMA-ADPCM voice prompts to the Ailunce HD2 analog-FM build.

> **Stacks on #500** (`hd2-fm`, the FM radio stack). Until #500 merges into `Ailunce-HD2`, this PR's diff shows both sets of commits; it narrows to the 5 voice-prompt commits once #500 lands. **Please merge after #500.**

## What & why

The HD2's HR_C7000 / CK803S core has no FPU, so codec2 voice-prompt *decode* is not realtime. Instead the stock `voiceprompts.vpc` is transcoded host-side to IMA-ADPCM (integer decode, trivially realtime) and played through the HR_C7000 codec DAC via the SAHB PCM mailbox.

## Commits
- **core:** IMA-ADPCM voice-prompt player + decode primitive (`lib/codec`) + host transcode tool (`scripts/vpc_to_adpcm.py`)
- **HD2:** codec-DAC PCM output-stream driver — an `audioDriver` impl over the SoC PCM-bridge frame IRQ
- **HD2:** build the ADPCM voice prompts; drop the placeholder `vp_*` stubs
- **HD2:** route MCU playback to the codec DAC (PTB17 path-select) + PCM-tone keytone
- **HD2:** store voice prompts at 4 kHz to fit the full vocabulary

## 4 kHz storage

ADPCM is ~10x the size of the codec2 source it is transcoded from; the full 229-clip vocabulary at 8 kHz (~656 KB) overflows the app-flash slot, so an earlier subset carried only digits/letters. Storing clips at 4 kHz — anti-alias FIR + 2:1 decimate host-side, 2x linear-interpolation upsample back to the codec DAC's fixed 8 kHz on-device — halves that to ~329 KB, which fits. The full menu/settings vocabulary now narrates.

## Testing

HW-verified on device: clean boot, full voice-prompt narration at levels 1–3, FM RX/TX unaffected, speaker amp powers down cleanly between prompts. Reviewed against the project's driver-layering and interface-conformance standards.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


fenfir:
Regarding the ADPCM voice prompts, I have a couple of of knobs to turn for a space trade off:
* High/Low Quality
* Subset the set
* Push to NVM

Also, for the voice prompts. What is 228 supposed to say? From context it seems like it's supposed to be "voice" but I don't hear that.

fenfir:
We can also skip this PR if we want to use the codec2 prompts, but I'm still trying to get the worst case time down to something reasonable.

The current state on my Codec2-mod work is 
min: 3.6 ms 
mean: 20.1 ms
max: 45.1 ms

PR #502, opened 2026-08-06
DTMF Support

turnrye:
This PR:

- Moves the 1750hz feature to being activated  via PTT+F1, in order to make way for...
- Most hardware platforms support "hot keyboard" style DTMF

The behavior is that if the user is holding PTT, and they are on a UV3x0 or CS7000 and they press the keypad 0-9*#, the corresponding DTMF tone is played over the speaker (observing the volume setting) and sent over RF. the 1750hz access tone also gets similar speaker feedback now.

There are two implementations for how this works, which was required due to one of my UX requirements of giving the user feedback of what is being transmitted by playing the sound also over the speaker. On the UV3x0/DM1701 platform, tones for RF are generated via the HR_C6000's built-in dtmf generator and for AF the come via the DAC so that the can use the volum knob's volume. For the CS7000 and plus, since the DAC playback can't be routed this way during transmission (I tried for a bit and failed!), the tone is generated via MCU and put to the rtx sink and then echoed to the speaker.

I tested both mechanisms with an rtl_sdr and found that error was small, though as you'd imagine the dedicated dtmf function performed slightly better. Both also checked out as working properly with my local allstar node for DTMF based repeater commands. I also verified the 1750hz tone feature as working now on both platforms.

As part of making these changes, I also had to update HR_C6000::sendTone() so that the table went to the proper register, which then allowed me to shift the uv3x0 to use this pattern rather than the AT1846S (because it failed to work in my testing).
Also, startAnalogTx() on both platforms now keeps the lineout2 enabled during TX so that the tone could be played ovr the speaker to inform the user of what was going on.

I investigated originally keeping the 1750hz tone on the # key, but on discord got feedback that the requisite setting was not optimal to maintain that + introduce dtmf. So, mapping the key was the suggestion and it seems to work nicely in my testing tonight.

Other platforms like the  MD-380/390, GD-77 and DM-1801 are not setup to us this feature now, as I don't have those on hand and the feature seems quite platform specific (though hopefully by introducing a couple of approaches here in this PR, those future additions would require far less work).

### DIFF
--- a/openrtx/include/core/state.h
+++ b/openrtx/include/core/state.h
@@ -41,6 +41,7 @@ typedef struct {
     uint16_t bank;
     uint8_t rtxStatus;
     bool tone_enabled;
+    uint8_t dtmf_code; /* FM DTMF digit being pressed, or DTMF_CODE_NONE */
 
     bool emergency;
     settings_t settings;
--- a/openrtx/include/rtx/OpMode_FM.hpp
+++ b/openrtx/include/rtx/OpMode_FM.hpp
@@ -8,6 +8,7 @@
 #define OPMODE_FM_H
 
 #include "core/audio_path.h"
+#include "hwconfig.h"
 #include "OpMode.hpp"
 
 /**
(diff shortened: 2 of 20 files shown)


turnrye:
@silseva and I synced to discuss this earlier today -- this PR is going to go into a closed state. I'll be cutting a small epic on the Taiga task tracker to explain a minor platform refactor that we want to do first, which ultimately will simplify the implementation that wee take here and make the device support easier (both today and for future targets). This work will be a little further down my personal to-do list for the next month or so. Closing this PR and will link it to the future issue once it's ready for reference.

PR #508, opened 2026-09-12
In CI, use UV to install Py deps for scripts

turnrye:
### Summary

Install the Python dependencies with `uv` instead of `pip`.  This let's us clean up a few CI things and in the near future make installs of local env a bit easier and faster thanks to some of `uv`'s improvements compared with pip and normal venv approaches.

### Related issue

This is a broken out step 1 from the original PR https://github.com/OpenRTX/OpenRTX/pull/446.

### Changelog entry

Suggesting that we omit this

### Checklist

- [x] This PR is one logical change; unrelated fixes are in separate PRs
- [x] Commits are tidy; each one builds, no "fix typo" commits left in the history
- [x] I have read [CONTRIBUTING.md](../blob/master/CONTRIBUTING.md)

This is a drop-in swap of the installer only: requirements.txt stays the source of truth, the scripts keep running against the same interpreter, and no build wiring changes.

In the devcontainer this also lets uv supply the interpreter, which removes the deadsnakes PPA and the extra apt round-trip it needed. The environment is still a plain virtualenv at /opt/venv on PATH, and it still holds the same package set, so nothing downstream of it changes. python3-pip is dropped from the image, as nothing installs with pip any more.

Adopting uv's project workflow -- pyproject.toml, a committed uv.lock and `uv run` -- is deliberately left out here so that this change stays small and easy to revert. That migration is worthwhile but it is a much larger diff, most of it the lockfile, and it makes uv a hard requirement of the build; it should be reviewed on its own rather than riding along with an installer swap.


silseva:
After this PR gets merged, will `uv` become an hard dependency also when the python scripts are run outside a venv? If we can keep the instructions inside the meson.build transparent, it'd be gold. I say this because some people, like me, have the required packages installed in the system environment (with a package manager, not even `pip`) and I'm a bit reluctant on forcing an additional dependency on `uv`.

turnrye:
> After this PR gets merged, will `uv` become an hard dependency also when the python scripts are run outside a venv? If we can keep the instructions inside the meson.build transparent, it'd be gold. I say this because some people, like me, have the required packages installed in the system environment (with a package manager, not even `pip`) and I'm a bit reluctant on forcing an additional dependency on `uv`.

No, there are no dependencies here on `uv` introduced to users (unless they use the devcontainer). This focuses instead on how CI operates. I _did_ make a change also in the AGENTS.md file that guides them to use `uv` rather than `pip`, but that's not required and is instead more of a suggestion.

PS: I think the reason you're asking this is actually because of one of the changes I had in my previous MR -- I will note your feedback and change the approach once I'm to that step (~2 PRs from now).

silseva:
> PS: I think the reason you're asking this is actually because of one of the changes I had in my previous MR -- I will note your feedback and change the approach once I'm to that step (~2 PRs from now).

Yes, that's the thing wanted to comment in the previous PR. Thank you!

PR #509, opened 2026-09-13
core: add message registry and dispatcher

turnrye:
# Summary

Add the protocol-independent core of the message inbox: a registry that owns all message storage, and a dispatcher that moves data packets between the rtx stage and protocol sources. First of three chained PRs replacing #492; the M17 SMS source and the UI follow, and nothing here is reachable until they do (everything is gated by CONFIG_MESSAGES, which no target defines yet).

# Related issue

Supersedes #492 (this PR + two follow-ups replace it).

Implements https://tasks.openrtx.org/project/openrtx/task/836

# Design notes and alternatives

I've made the following changes since #492 was reviewed, based on my understanding of feedback:

- Registry owns all storage. A fixed entry table plus a shared, variable-length body pool, evicting oldest-first; sources file received messages through messages_store() and keep nothing of their own. Eviction, unread tracking, lookup by sequence and, later, persistence live in one place regardless of protocol. Capacity is per target (CONFIG_MESSAGES_MAX_ENTRIES / CONFIG_MESSAGES_POOL_BYTES, default 32 / 3200 B).
- No packet_io, no RTX-thread task. struct pktDesc + rtx_addPacketRx()/rtx_addPacketTx() are already the thread-safe boundary, so the dispatcher just owns a small pool of descriptors and buffers on the UI thread: keeps a receive descriptor submitted while a source is registered for the current opmode, hands each completed packet to that source, and on transmit lets the source format into its buffer, submits it and reports the outcome to the registry. It never interprets packet content — the source gets the completed struct pktDesc as opaque bytes.
- Sources are stateless translators: struct message_ops { process_rx, format_tx, name, mode }, registered at run time. supported_actions, invoke_action, start_compose, count, get and tick are gone; the UI will call messages_mark_read() / messages_delete() directly.
- One memcpy per direction (RX slot → pool, pool → TX slot). Fully zero-copy was considered and rejected: it would tie inbox capacity to fixed 845-byte packet slots.

Alternative not taken: running the dispatcher on main_thread like the APRS packet engine. It sits on the UI thread next to the registry so no state crosses threads; if I misunderstood this please let me know @silseva 

# Testing performed

- Added new unit tests
- On hardware this PR is inert (no target enables CONFIG_MESSAGES), so only the builds were checked there.

## Tested on:

- [x] Linux emulator
- [ ] MD-3x0 (TYT MD-380 / MD-390)
- [ ] MD-UV3x0 (TYT MD-UV380 / UV390)
- [ ] MD-9600
- [ ] GDx (Radioddity GD-77 / Baofeng DM-1801)
- [ ] DM-1701
- [ ] CS7000 / CS7000-PLUS
- [ ] RT-4D
- [ ] Module17
- [ ] ttwrplus
- [ ] Not tested on hardware — help wanted

# Changelog entry

none (nothing user-visible until the M17 SMS source and UI land)

# Checklist

- [x] This PR is one logical change; unrelated fixes are in separate PRs
- [x] Commits are tidy; each one builds, no "fix typo" commits left in the history
- [x] I have read CONTRIBUTING.md


silseva (changes requested):
General comments:

- the message dispatcher ended up being pretty simple: I think we can move the logic directly inside the "messages" module and make the "process_rx" and "format_tx" two message callbacks alongside the others (which I saw in the previous PR). Doing so will also squeeze the code size a bit more.
- the general naming convention for functions in OpenRTX is moduleName_functionName: this should be used also here.
- for large buffers I'd use dynamic memory allocation whenever possible: given that we allocate/release large blocks of memory, the fragmentation should not be a real issue. Dynamic memory allocation allows us to use the RAM more efficiently: if all the buffers are static, we pay for them at compile time and we'll run out of memory faster than expected. The radios based on STM32F4x allocate .bss and .data in the 64kB CCM RAM, while the heap goes in the large 128kB RAM. More in general, given that messages is something which may be used, a "pay for what you effectively use" approach works better. This is just a general indication, we then need to find the best way to implement this principle.

silseva:
A clarification about dynamic memory allocation: the idea is to allocate memory when switching to an opmode which supports messages and release it when changing to something else. It may happen to have a back-to-back release/allocate pattern, but is fine.

openrtx/src/core/messages.c
```diff
@@ -0,0 +1,192 @@
...
+/**
+ * \internal
+ * Reserve room in the pool for a body of @p len characters plus NUL,
+ * evicting every entry whose body overlaps the reserved range. Those entries
+ * are always older than any other one still stored, since the pool is
+ * written in arrival order.
+ *
+ * @return pool offset of the reserved range.
+ */
+static uint16_t poolAlloc(uint16_t len)
```
silseva: For the pool of message bodies you can use `size_t` everywhere, to avoid casting.

openrtx/src/core/messages.c
```diff
@@ -0,0 +1,192 @@
...
+               "CONFIG_MESSAGES_POOL_BYTES must fit in a uint16_t offset");
+
+/*
+ * Entries are kept oldest-first, so entries[0] is the eviction candidate and
+ * entries[numEntries - 1] is the newest. Bodies live in the pool as
+ * NUL-terminated strings allocated in arrival order: poolHead is the next
+ * byte to write and wraps to zero when a body would not fit at the end.
+ */
+static struct message entries[CONFIG_MESSAGES_MAX_ENTRIES];
+static size_t numEntries;
```
silseva: IMO the message pool can be managed as a ring buffer: the newest element is stored at the head and you have automatic eviction of old entries without the need for a `memmove` (which tends to be expensive).

openrtx/src/core/messages.c
```diff
@@ -0,0 +1,192 @@
...
+    entry->body = &pool[offset];
+    entry->sequence = nextSequence++;
+    entry->sender[MSG_ADDR_MAX_LEN - 1] = '\0';
+    entry->recipient[MSG_ADDR_MAX_LEN - 1] = '\0';
+
+    if ((entry->direction == MSG_DIR_RX) && (entry->unread != 0))
+        newArrivals++;
+
+    if (sequence != NULL)
+        *sequence = entry->sequence;
```
silseva: The sequence number is always positive and greater than zero: perhaps we can just return it instead of using a parameter passed by pointer.

openrtx/src/core/messages.c
```diff
@@ -0,0 +1,192 @@
...
+    if (msg->body_len >= MSG_BODY_MAX_LEN)
+        return -EMSGSIZE;
+
+    if (numEntries == ARRAY_SIZE(entries))
+        removeEntry(0);
+
+    uint16_t offset = poolAlloc(msg->body_len);
+    if (msg->body_len > 0)
+        memcpy(&pool[offset], msg->body, msg->body_len);
+    pool[offset + msg->body_len] = '\0';
```
silseva: What about having a `struct message *message_alloc()` (or `request`, or whatever else) doing only the allocation of message body content in the pool? The idea is to have the caller code requesting for a message, then filing it accessing directly the body, and then submitting it to the queue: this should spare an expensive memcpy.

I don't know if it can make sense or not: you know the general picture better than me.

openrtx/include/core/messages.h
```diff
@@ -0,0 +1,183 @@
...
+
+/* Maximum number of entries held in the inbox. */
+#ifndef CONFIG_MESSAGES_MAX_ENTRIES
+#define CONFIG_MESSAGES_MAX_ENTRIES 32
+#endif
+
+/* Size of the shared body pool, in bytes. */
+#ifndef CONFIG_MESSAGES_POOL_BYTES
+#define CONFIG_MESSAGES_POOL_BYTES 3200
+#endif
```
silseva: These can go to `messages.c`

silseva:
I think something went wrong with pushing the commits: what I have now in the history is still the original code you submitted, before my first review comments.

turnrye:
Argh! You were right, I'd made a mistake with where I pushed things. I've addressed that and so now it's _actually_ ready for a re-review. I addressed your points plus a few things that logically followed them. One though that I'd challenge:

About `message_alloc()`: I worked it through and it ends up either not saving the copy or costing more memory. The received text sits in the packet buffer, which is resubmitted to the rtx stage for the next packet as soon as processRx() returns. So with an alloc-after-parse API, the memcpy into the pool still happens, it just moves from messages_store() into every protocol source. The only version that really deletes it is receiving straight into the pool, which would be worse.

So, I suggest we keep the copy to keep the coupling low.

turnrye:
I synced `release-v0.5.0` with master and then rebased this onto there, no changes though. This is ready for review again.

PR #510, opened 2026-09-14
Retevis C62 platform support

OE3ANC:
### Summary

This pull request brings Retevis C62 platform support to OpenRTX:

- Build system / board support
- Device tree / platform HAL
- Display driver
- Keyboard driver
- (generic) BK4819 driver
- (generic) BK1080 driver

This PR is based on the [pull request](https://github.com/OpenRTX/OpenRTX/pull/415) by @Tunas1337


west.yml
```diff
@@ -10,9 +10,14 @@ manifest:
...
     remote: zephyr
 
+  group-filter:
+    - -c62
+
   remotes:
     - name: zephyr
       url-base: https://github.com/zephyrproject-rtos
+    - name: c62-dependencies
+      url-base: https://github.com/OE3ANC
```
silseva:
```suggestion
      url-base: https://github.com/OpenRTX
```

openrtx/include/core/threads.h
```diff
@@ -12,10 +12,17 @@
...
  * Threads' stack sizes
  */
+#ifndef __ZEPHYR__
 #define UI_THREAD_STKSIZE     2048
 #define RTX_THREAD_STKSIZE    512
 #define CODEC2_THREAD_STKSIZE 16384
 #define AUDIO_THREAD_STKSIZE  512
+#else
+#define UI_THREAD_STKSIZE     2048
+#define RTX_THREAD_STKSIZE    1024
```
silseva, line 17: To be defined if increasing the stack size to 1kB will still be necessary in the future. In the meantime, instead of copying all the lines to change just one, I'd do the following:
```
#define UI_THREAD_STKSIZE     2048
#ifdef PLATFORM_C62
#define RTX_THREAD_STKSIZE    1024
#else
#define RTX_THREAD_STKSIZE    512
#endif
#define CODEC2_THREAD_STKSIZE 16384
#define AUDIO_THREAD_STKSIZE  512
```

platform/targets/c62/c62_defconfig
silseva, line 1: Some of the `CONFIG` directives in this file are duplicated with respect to `platform/mcu/CSK6011B/zephyr.conf`: I'd keep here only the things which are strictly related to C62 and move all the configurations for the CSK6011B chip in the other config file (for example `CONFIG_SOC_CSK6`).
We don't want to get crazy in the future just because some config is changed here, but the other file is still keeping it to the old value.

platform/targets/c62/c62_defconfig
```diff
@@ -0,0 +1,61 @@
...
+CONFIG_I2C_SHELL=n
+CONFIG_PWM_SHELL=n
+
+CONFIG_ADC=y
+CONFIG_ADC_SHELL=n #this requires a device tree entry before it can work.
+CONFIG_ADC_CSK6=y
+
+CONFIG_BK4819=y
+CONFIG_BK4819_LOG_LEVEL_DBG=n
+
```
silseva: These can be removed

platform/targets/c62/c62_defconfig
```diff
@@ -0,0 +1,61 @@
...
+CONFIG_BK1080=y
+CONFIG_BK1080_LOG_LEVEL_DBG=n
```
silseva: Same as above

platform/targets/c62/c62.dts
```diff
@@ -0,0 +1,395 @@
...
+		};
+	};
+	
+    bk4819_radio: bk4819 {
+        compatible = "beken,bk4819";
+        status = "okay";
+        sclk-gpios = <&gpioa 13 GPIO_ACTIVE_HIGH>;
+        sdata-gpios = <&gpioa 7 GPIO_ACTIVE_HIGH>;
+        scn-gpios = <&gpioa 8 GPIO_ACTIVE_LOW>;
+    };
```
silseva: BK4819 driver comes with OpenRTX, outsize Zephyr: these lines can be removed

platform/targets/c62/c62.dts
```diff
@@ -0,0 +1,395 @@
...
+
+	bk1080_radio: bk1080 {
+        compatible = "beken,bk1080";
+        status = "okay";
+        i2c-clk-gpios = <&gpioa 5 GPIO_ACTIVE_HIGH>;
+        i2c-data-gpios = <&gpioa 6 GPIO_ACTIVE_HIGH>;
+        power-gpios = <&gpiob 2 GPIO_ACTIVE_LOW>;
+    };
```
silseva: The same for these

platform/targets/c62/c62_pinctrl.dtsi
silseva, line 1: The indentation of this file is a bit crazy, it can be made uniform

platform/targets/c62/platform.c
```diff
@@ -0,0 +1,271 @@
...
+    GPIO_DT_SPEC_GET(DT_NODELABEL(ledkeyboard), gpios);
+
+static const struct pwm_dt_spec pwm_lcd_backlight =
+    PWM_DT_SPEC_GET(DT_NODELABEL(pwm_lcd_backlight));
+static const struct pwm_dt_spec pwm_rf_apc =
+    PWM_DT_SPEC_GET(DT_NODELABEL(pwm_rf_apc));
+
+const struct device *adc_dev = DEVICE_DT_GET(DT_NODELABEL(adc0));
+
+static hwInfo_t hwInfo = {
```
silseva:
```suggestion
static const hwInfo_t hwInfo = {
```

platform/targets/c62/platform.c
```diff
@@ -0,0 +1,283 @@
...
+    GPIO_DT_SPEC_GET(DT_ALIAS(ledgreen), gpios);
+static const struct gpio_dt_spec led_keyboard =
+    GPIO_DT_SPEC_GET(DT_NODELABEL(ledkeyboard), gpios);
+
+static const struct pwm_dt_spec pwm_lcd_backlight =
+    PWM_DT_SPEC_GET(DT_NODELABEL(pwm_lcd_backlight));
+static const struct pwm_dt_spec pwm_rf_apc =
+    PWM_DT_SPEC_GET(DT_NODELABEL(pwm_rf_apc));
+
+const struct device *radio = DEVICE_DT_GET(DT_NODELABEL(bk4819_radio));
```
silseva: To be removed

platform/targets/c62/platform.c
```diff
@@ -0,0 +1,283 @@
...
+const struct device *adc_dev = DEVICE_DT_GET(DT_NODELABEL(adc0));
```
silseva: Starting from the CS7000 BSP I introduced a new convention: all the devices are listed as `extern` in `hwconfig.h` and declared in `hwconfig.c`. I'd do the same for the C62.
For reference see:
- https://github.com/OpenRTX/OpenRTX/blob/master/platform/targets/CS7000/hwconfig.h#L32
- https://github.com/OpenRTX/OpenRTX/blob/master/platform/targets/CS7000/hwconfig.c

platform/targets/c62/platform.c
```diff
@@ -0,0 +1,283 @@
...
+
+static int battery_init(void);
+
+/**
+ * Set display brightness using PWM on pin A02
+ * PWM Period: 1000us (1kHz), adjustable duty cycle for brightness
+ * 
+ * @param brightness_percent: 0-100 (0 = off, 100 = full brightness)
+ */
+void platform_set_display_brightness(uint8_t brightness_percent)
```
silseva: This code should go straight into `display_setBacklightLevel` inside the display driver.

platform/targets/c62/platform.c
```diff
@@ -0,0 +1,283 @@
...
+    }
+}
+
+/**
+ * Set TX power using PWM on pin A03
+ * PWM Period: 1000us (1kHz), adjustable duty cycle for power level
+ * 
+ * @param power_percent: 0-100 (0 = min power, 100 = max power)
+ */
+void platform_set_tx_power(uint8_t power_percent)
```
silseva: This function should go directly in the radio driver

platform/targets/c62/platform.c
```diff
@@ -0,0 +1,283 @@
...
+    if (ret < 0) {
+        printk("Failed to set TX power PWM: %d\n", ret);
+    } else {
+        printk("TX power set to %d%%\n", power_percent);
+    }
+}
+
+void platform_init_csk6()
+{
+    printk("0x49 de OE3ANC from OPENRTX on the C62\n");
```
silseva: Nice, but I'd remove it, if you don't mind. :)

platform/targets/c62/platform.c
```diff
@@ -0,0 +1,283 @@
...
+    gpio_pin_configure_dt(&led_green, GPIO_OUTPUT_INACTIVE);
+    gpio_pin_configure_dt(&led_keyboard, GPIO_OUTPUT_INACTIVE);
+
+    // Configure speaker enable (don't set it as output if you want to use SWD debugging!)
+    gpio_pin_configure_dt(&speaker_enable, GPIO_OUTPUT);
+
+    gpio_pin_configure_dt(&dtmf_enable, GPIO_OUTPUT);
+
+    // DT_EN (Audio routing control for BK4819): 0 = DSP left channel output to AMP, 1 = DTMF from BK4819 to AMP
+    gpio_pin_set_dt(&dtmf_enable, 0); // DSP left channel output to amplifier
```
silseva: These three initializations should go inside `audio_init`, since they are part of the audio module

platform/targets/c62/platform.c
```diff
@@ -0,0 +1,283 @@
...
+    /* Init ADC for Battery reading */
+    battery_init();
+}
+
+void platform_terminate()
+{
+    ;
+}
+
+static int battery_init(void)
```
silseva, line 73: I'd move this directly to line 68 and get rid of the forward declaration.

platform/targets/c62/platform.c
```diff
@@ -0,0 +1,283 @@
...
+    return 0;
+}
+
+/**
+ * Read battery voltage using ADC2 on pin B08
+ * @return: Battery voltage in millivolts (mV), or 0 on error
+ */
+uint16_t platform_getVbat()
+{
+    return 7200; // Return fixed value until voltage reading is working reliably
```
silseva, line 113: We should implement battery voltage reading, otherwise the radio is unusable (it will die all of a sudden).

platform/targets/c62/platform.c
```diff
@@ -0,0 +1,283 @@
...
+static int16_t adc_raw_buffer[ADC_NUM_CHANNELS];
+struct adc_sequence sequence = {
+    /* individual channels will be added below */
+    .channels = BIT(
+        2), // BIT(2) Channel 2 (BAT_DET), BIT(1) Channel 1 (Charge detection)
+    .buffer = adc_raw_buffer,
+    /* buffer size in bytes, not number of samples */
+    .buffer_size = sizeof(adc_raw_buffer),
+    .resolution = 11, // 11-bit resolution
+};
```
silseva, line 41: If these variables are used only for battery voltage readout, I'd move them inside `platform_getVbat` and use a single `int16_t` instead of an array (at least until/unless we have charge detection)

platform/targets/c62/platform.c
```diff
@@ -0,0 +1,283 @@
...
+
+bool platform_getPttStatus()
+{
+    //return gpio_pin_get_dt(&button_ptt); // This may brick your radio! Verify what i did in radio_C62.cpp before running this!
+    return false;
+}
+
+bool platform_pwrButtonStatus()
+{
+    return true;
```
silseva, line 139: Hm. I think this should be enabled, eventually by checking the battery voltage (like we do in the MD-380). Otherwise the firmware is not able to detect when the radio is turning off.
This also rises a question: is there a pin to keep the radio alive irrespectively of the status of the main power switch?

platform/drivers/display/display_c62.c
```diff
@@ -0,0 +1,112 @@
...
+
+    // Process each 16-bit pixel
+    for (size_t i = 0; i < FB_SIZE; i += 2) {
+        // Extract the 16-bit RGB565 value
+        uint16_t pixel = (frameBuffer[i] << 8) | frameBuffer[i + 1];
+
+        // Store in shadow buffer with byte order swapped
+        // This fixes endianness issues which can cause color problems
+        shadowBuffer[i + 1] = (pixel >> 8) & 0xFF;
+        shadowBuffer[i] = pixel & 0xFF;
```
silseva, line 98: Instead of swapping bytes here, I'd add a `CONFIG_PIX_FMT_RGB565_BE` directive and adapt the graphics.c driver accordingly. All the pixels are set via the `gfx_setPixel` function, which easily allows to do such things. There is also the `__builtin_bswap16` function from GCC which maps to direct asm instructions if available.

platform/drivers/display/display_c62.c
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    display_write(displayDev, 0, 0, &displayBufDesc, shadowBuffer);
+}
+
+void display_setContrast(uint8_t contrast)
```
silseva: The `CONFIG_DISPLAY_CONTRAST` is not defined, so this function can be removed.

platform/drivers/keyboard/keyboard_c62.c
```diff
@@ -0,0 +1,116 @@
...
+
+void kbd_terminate()
+{
+}
+
+keyboard_t kbd_getKeys()
+{
+    uint32_t keys = 0;
+    const struct gpio_dt_spec *cols[] = { &row1, &row2, &row3, &row4 };
+    const struct gpio_dt_spec *rows[] = { &col1, &col2, &col3, &col4, &col5 };
```
silseva, line 53: These can go outside the function, with the other declarations.

platform/drivers/keyboard/keyboard_c62.c
```diff
@@ -0,0 +1,116 @@
...
+
+    // Map raw keys to KEY_ enum values
+    keyboard_t result = 0;
+
+    static const uint32_t keyMap[] = {
+        KEY_ENTER, KEY_UP,   KEY_DOWN, KEY_ESC, KEY_3, KEY_4,
+        KEY_5,     KEY_6,    KEY_7,    KEY_8,   KEY_9, KEY_STAR,
+        KEY_0,     KEY_HASH, 0,        0, // 14 and 15 are unused
+        KEY_MONI,  KEY_F2,   KEY_1,    KEY_2
+    };
```
silseva, line 95: Same for this

platform/drivers/baseband/radio_C62.cpp
```diff
@@ -0,0 +1,265 @@
...
+    GPIO_UHF_RX_LNA,
+    GPIO_VHF_TX_PA,
+    GPIO_UHF_TX_PA,
+    GPIO_ALC_TX_LED
+};
+
+/**
+ * Calculate DCS parity and compose
+ */
+static uint32_t cdcss_compose(uint16_t cdcss_code)
```
silseva, line 64: This function can go in `core/utils.c`, probably. Is it specific for the radio chip, or it is the generic implementation of CDCSS code composition?

platform/drivers/baseband/radio_C62.cpp
```diff
@@ -0,0 +1,265 @@
...
+    bk4819_set_modulation(true);
+    return;
+}
+
+void radio_disableAfOutput()
+{
+    BK4819_SetAF(0);
+}
+
+void radio_checkVOX()
```
silseva, line 174: I'd remove this function, for now, and reintroduce it later when we'll have a more comprehensive VOX framework. Moreover, VOX should be checked by OpModes like the PTT: forcing the radio status here is not correct.

platform/drivers/baseband/radio_C62.cpp
```diff
@@ -0,0 +1,265 @@
...
+    if (bk4819_get_vox()) {
+        if (radioStatus != TX)
+            radio_enableTx();
+    } else {
+        if (radioStatus != RX)
+            radio_enableRx();
+    }
+}
+
+void radio_setRxFilters(uint32_t freq)
```
silseva, line 187: I'd move this code inside `radio_updateConfiguration`

platform/drivers/baseband/radio_C62.cpp
```diff
@@ -0,0 +1,285 @@
...
+    }
+}
+
+void radio_enableRx()
+{
+    bk4819_gpio_pin_set(&c62_bk4819, 0, false); // VHF RX LNA
+    bk4819_gpio_pin_set(&c62_bk4819, 1, false); // UHF RX LNA
+    bk4819_gpio_pin_set(&c62_bk4819, 2, false); // UHF TX PA
+    bk4819_gpio_pin_set(&c62_bk4819, 3, false); // UHF TX PA
+    bk4819_gpio_pin_set(&c62_bk4819, 4, false); // ALC / TX LED
```
Marko13-dev, line 204: We should use the already available defines (GPIO_ALC_TX_LED, ...) here instead of the magic numbers 0, 1, 2, 3, 4

platform/drivers/baseband/radio_C62.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+    bk4819_rtx_off();
+    radioStatus = OFF;
+}
+
+void radio_updateConfiguration()
+{
+    // Set squelch
+    int squelch = -127 + (config->sqlLevel * 66) / 15;
+    bk4819_set_Squelch(((squelch + 160) * 2), ((squelch - 3 + 160) * 2), 0x5f,
```
silseva: Squelch is managed by OpMode_FM, I'd remove this for now. Like VOX, I plan to set up something allowing to exploit hardware capabilities when present.

platform/drivers/baseband/radio_C62.cpp
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    bk4819_gpio_pin_set(&c62_bk4819, GPIO_ALC_TX_LED,
+                        true); // ALC / TX LED
+
+    // depending on power level set PWM duty cycle for APC voltage control
+    // Maybe need table for this instead of crude linear mapping, and also consider frequency dependence of PA efficiency
+    platform_set_tx_power(std::min(
+        config->txPower * 100 / 5000,
+        100U)); // crude linear mapping of power to duty cycle, max at 5W
```
Marko13-dev, line 257: Following code is closer to reality (check with nano VNA) until we have a real table:
   platform_set_tx_power(std::min(
        config->txPower * config->txPower * 100 / (5000 * 5000),
        100U)); // crude quadratic mapping of power to duty cycle, max at 5W

platform/targets/c62/audio_c62.c
silseva, line 1: As discussed on discord, please remove the all the audio input/output (buffers, threads, ...): these should go to an audio stream implementation.
This interface is meant only to initialize the audio subsystem and for device muxing.

platform/drivers/baseband/radio_C62.cpp
```diff
@@ -0,0 +1,285 @@
...
+                        false); // VHF RX LNA
+    bk4819_gpio_pin_set(&c62_bk4819, GPIO_UHF_RX_LNA,
+                        false); // UHF RX LNA
+    bk4819_gpio_pin_set(&c62_bk4819, GPIO_VHF_TX_PA,
+                        false); // VHF TX PA
+    bk4819_gpio_pin_set(&c62_bk4819, GPIO_UHF_TX_PA,
+                        false); // UHF TX PA
+    bk4819_gpio_pin_set(&c62_bk4819, GPIO_ALC_TX_LED,
+                        false); // ALC / TX LED
+
```
Marko13-dev, line 232: Do we need a short delay here to let the hardware settle before activating the PA?

platform/drivers/baseband/radio_C62.cpp
```diff
@@ -0,0 +1,285 @@
...
+void radio_enableRx()
+{
+    bk4819_gpio_pin_set(&c62_bk4819, 0, false); // VHF RX LNA
+    bk4819_gpio_pin_set(&c62_bk4819, 1, false); // UHF RX LNA
+    bk4819_gpio_pin_set(&c62_bk4819, 2, false); // UHF TX PA
+    bk4819_gpio_pin_set(&c62_bk4819, 3, false); // UHF TX PA
+    bk4819_gpio_pin_set(&c62_bk4819, 4, false); // ALC / TX LED
+
+    radio_setRxFilters(config->rxFrequency);
```
Marko13-dev, line 206: Do we need a short delay here to let the hardware settle before activating the RX path?

silseva:
Please squash commit `d843f064c7b772f7ba126fb54dce785f32fc8fd4` into `157eef262aa8f3e059751068d7073cc2e4dc8aeb`.
The same for commit `bbe4bfa6870a0640202db0ccb431dc0db24e6619`: it needs to be split in two parts (one for each chip) and each part needs to be squashed inside the commit it refers to.

I didn't review the BK drivers yet, I'll wait for the commits to be squashed before looking at them.

Also: do we really need all the secondary services like wasm, vad-service, thinker-service? If possible, I'd remove what we do not need.

silseva (changes requested):

platform/drivers/keyboard/keyboard_c62.c
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    // Map raw keys to KEY_ enum values
+    keyboard_t result = 0;
+
+    static const uint32_t keyMap[] = {
+        KEY_ENTER, KEY_UP,   KEY_DOWN, KEY_ESC, KEY_3, KEY_4,
+        KEY_5,     KEY_6,    KEY_7,    KEY_8,   KEY_9, KEY_STAR,
+        KEY_0,     KEY_HASH, 0,        0, // 14 and 15 are unused
+        KEY_MONI,  KEY_F2,   KEY_1,    KEY_2
```
Marko13-dev, line 94: KEY_MONI is wrong. It must be KEY_F1. (KEY_MONI is checked below). If KEY_F1 should work as Monitor or Macro menu, then the mapping should be in another file.

platform/targets/c62/platform.c
```diff
@@ -0,0 +1,271 @@
...
+    return 0;
+}
+
+/**
+ * Read battery voltage using ADC2 on pin B08
+ * @return: Battery voltage in millivolts (mV), or 0 on error
+ */
+uint16_t platform_getVbat()
+{
+    return 7200; // Return fixed value until voltage reading is working reliably
```
Marko13-dev, line 113: FYI: I have working code to read battery voltage. I will provide in a future PR.

platform/targets/c62/platform.c
```diff
@@ -0,0 +1,271 @@
...
+
+int8_t platform_getChSelector()
+{
+    return 0;
+}
+
+bool platform_getPttStatus()
+{
+    //return gpio_pin_get_dt(&button_ptt); // This may brick your radio! Verify what i did in radio_C62.cpp before running this!
+    return false;
```
Marko13-dev, line 134: Should it still be disabled?

platform/targets/c62/platform.c
Marko13-dev: The main power switch cuts the supply. When the C62 is in the charging cradle the power is activated (even with main power switched off). This status can be theoretically detected by a pin.

platform/targets/c62/platform.c
Marko13-dev: I have a working implementation. I planned to provide it in a future PR, but I could also provide it to Andy for integration in this PR.

platform/drivers/baseband/radio_C62.cpp
silseva: Indeed, I didn't notice. Thank you!

platform/drivers/baseband/radio_C62.cpp
silseva: Let's use that mapping, then

platform/targets/c62/platform.c
silseva: If it can be added to this PR, squashing into the commit adding `platform.c`, it would be great. Otherwise, you can open a second PR with fixup commits to be added later (I'll merge all the PRs in a dedicated branch and then mainline everything once the port is stable, to allow for this kind of works).

platform/targets/c62/platform.c
silseva: Aha! So we need to check for the battery charge pin and make `platform_pwrButtonStatus()` return false when the radio is being charged.
We can exploit this mechanism: https://github.com/OpenRTX/OpenRTX/blob/1b31ba1b08b5d732ead34fa5d22d49202e5710e6/openrtx/src/main.c#L20

west.yml
OE3ANC: done, fixed in 07c0a7bdc9db9cc4addd4a57781fde7c2a937e89

openrtx/include/core/threads.h
OE3ANC: done, fixed in a86cdd63d9c9cd4bbabac6ae22fb511e54113ce1

platform/targets/c62/c62_defconfig
OE3ANC: Removed redundant configs, moved non-c62 configs from c62_defconfig to CSK6011Bs zephyr.conf in c960f5a34d9e3a67bcca1c012ae4202e1e7a686d

platform/targets/c62/c62_defconfig
OE3ANC: removed in bbe4bfa6870a0640202db0ccb431dc0db24e6619

platform/targets/c62/c62_defconfig
OE3ANC: removed in bbe4bfa6870a0640202db0ccb431dc0db24e6619

platform/targets/c62/c62.dts
OE3ANC: removed in bbe4bfa6870a0640202db0ccb431dc0db24e6619

platform/targets/c62/c62.dts
OE3ANC: removed in bbe4bfa6870a0640202db0ccb431dc0db24e6619

platform/targets/c62/c62_pinctrl.dtsi
OE3ANC: fixed in 69bf02af12662af843ab5fcdc65572f28f534587

platform/targets/c62/platform.c
OE3ANC: fixed in 19d8e7e4b6f5a419f098fc9bec1b480c88000b81

platform/targets/c62/platform.c
OE3ANC: removed in bbe4bfa6870a0640202db0ccb431dc0db24e6619

platform/targets/c62/platform.c
OE3ANC: updated in 02e5a214ce352d1db63c506717cf9ccbade50ef4

platform/targets/c62/platform.c
OE3ANC: moved it to the displaydriver in 07d879bd9a36c51cdb0fda750e76ed7034c8f378

platform/targets/c62/platform.c
OE3ANC: moved it to the radio driver in f265f6d85795fdece7fd33380ba7a5dc58965e17

platform/targets/c62/platform.c
OE3ANC: removed in d843f064c7b772f7ba126fb54dce785f32fc8fd4

platform/targets/c62/platform.c
OE3ANC: moved them into audio_init in 971254cfc5fe0d02ead20b91d86d1ad87f04c1e9

platform/drivers/display/display_c62.c
OE3ANC: removed in a4a19914e20af9d730f6b87f81c82a10d8220de1

platform/targets/c62/platform.c
OE3ANC: I'd keep it as it is so @Marko13-dev can provide an PR for it

silseva: Ok for me

platform/targets/c62/platform.c
OE3ANC: I'd keep it disabled until we have a way to get proper calibration table. Max power on my C62 with the current mappings we have exceeds 5W on the highest power settings

silseva: I'd make this function return the actual status of the PTT pin and disable TX on the radio driver. This will allow to add the calibration table and enable TX in a single commit, to not mention giving the possibility to do tests on the TX part up to the RF stage (excluded).

platform/drivers/baseband/radio_C62.cpp
OE3ANC: Could not hurt i think. How long should the delay be?

